PR #1, opened 2023-09-22
Fix path to build dir in run-debug.sh

bashor:


nikitabobko:
I prefix all the scripts with `./`. 

So I applied an additional patch and merged:
```
diff --git a/run-debug.sh b/run-debug.sh
index 4525835..71e30a1 100755
--- a/run-debug.sh
+++ b/run-debug.sh
@@ -8,4 +8,4 @@ cd "$(dirname "$0")"

 xcodegen # https://github.com/yonaskolb/XcodeGen
 ./build-debug.sh
-.build/AeroSpace-Debug.app/Contents/MacOS/AeroSpace-Debug
+./.build/AeroSpace-Debug.app/Contents/MacOS/AeroSpace-Debug
```

bashor:
> I prefix all the scripts with ./.

I'm wondering why.

nikitabobko:
Because otherwise, it might look like you're trying to run a program from the `$PATH`. When you see `./something` you immediately understand that it's an executable file that is located relative to PWD.

The second point is that if the script is located in the PWD. Then you can't run it by just doing `script`, right? You need to do `./script`. So it's a unification. All non-`$PATH` scripts are run with `./` prefix

I won't be surprised if the rule I'm describing is checked by some of the shell linters

PR #38, opened 2023-12-04
Support keypad

NilsBarlaug:
Hi!

Really appreciate the project you have created - I love it so far!

In i3 I really enjoy using the keypad (which I don't use for anything else) for window management shortcuts. I wonder if you would be willing to support keypad in the config?

I don't really know the codebase that well, but I've made a first attempt at implementing it. Haven't tested it yet, but I'd be willing to put in any additional work if necessary.


src/config/keysMap.swift
```diff
@@ -39,6 +39,25 @@ let keysMap: [String: Key] = [
...
+    "keypad5": .keypad5,
+    "keypad6": .keypad6,
+    "keypad7": .keypad7,
+    "keypad8": .keypad8,
+    "keypad9": .keypad9,
+    "keypadClear": .keypadClear,
+    "keypadDecimal": .keypadDecimal,
+    "keypadDivide": .keypadDivide,
+    "keypadEnter": .keypadEnter,
+    "keypadEquals": .keypadEquals,
```
nikitabobko, line 56: Please rename "keypadEquals" to "keypadEqual"

It's for consistency. The config already uses "equal" instead of "equals"

Please don't forget to rename it in the `default-config.toml`

src/config/keysMap.swift
```diff
@@ -39,6 +39,25 @@ let keysMap: [String: Key] = [
...
+    "keypad2": .keypad2,
+    "keypad3": .keypad3,
+    "keypad4": .keypad4,
+    "keypad5": .keypad5,
+    "keypad6": .keypad6,
+    "keypad7": .keypad7,
+    "keypad8": .keypad8,
+    "keypad9": .keypad9,
+    "keypadClear": .keypadClear,
+    "keypadDecimal": .keypadDecimal,
```
nikitabobko, line 53: How does the "keypadDecimal" key look on the keyboard? I'm looking at the Apple keyboard image right now and can't understand what key that might be https://www.apple.com/shop/product/MQ052LL/A/magic-keyboard-with-numeric-keypad-us-english

nikitabobko: I have figured it out. It's the period key on the keypad. Let's rename it to "keypadPeriod" then?

nikitabobko (changes requested):
I have tested your fix it works. Thanks!

src/config/keysMap.swift
NilsBarlaug: Good point 936e591

src/config/keysMap.swift
NilsBarlaug: I think it's actually more accurate to call it `keypadDecimal` (or similar) because it represent the decimal separator - and not necessarily period. It depends on the region/keyboard layout. For example, the Norwegian Magic keyboard has both period and comma 😅 :
image
And some (non-Apple) keyboards only have comma.

nikitabobko: Fair point, looking at https://en.wikipedia.org/wiki/Decimal_separator I have come up with the following names:
```
keypadDecimalMark
keypadDecimalSign
keypadDecimalPoint
keypadDecimalSeparator
keypadDecMark
keypadDecSign
keypadDecPoint
keypadDecSeparator
```

The "separator" looks too long for me. "point" refers more to the dot/period rather than the separator concept

Right now, I lean towards `keypadDecimalMark`/`keypadDecimalSign` but can't choose yet. Let me know what you think

nikitabobko: I don't like `keypadDecimal` because when I read it, I can't immediately understand what it means. Decimal what? It looks like an adjective without a noun

NilsBarlaug: Yeah, I agree - it's not immediately obvious.

I would speculate `keypadDecimalSeparator` is the most widely recognized (as the title of the Wikipedia article maybe reflects), but I see the point about it being long. I think both `keypadDecimalMark` and `keypadDecimalSign` is good options. `keypadDecimalMark` sounds (completely subjective) more natural to me 🤷 

Up to you 🙂

nikitabobko:
Will be released with the next release closer to the end of the week

PR #80, opened 2024-01-01
add NSDistributedNotificationCenter for better sketchybar support

### LINKED ISSUE #76: Support for Sketchybar Events
https://felixkratz.github.io/SketchyBar/config/events#creating-custom-events

Sketchybar uses notifications through `NSDistributedNotificationCenter` to support live updating of items. Would be cool if AeroSpace workspace changes would send such notifications.

MatthiasGrandl:
This PR adds a simple `NSDistributedNotificationCenter` notification to inform other applications of workspace focus changes.

Closes #76

### DIFF
--- a/src/command/WorkspaceCommand.swift
+++ b/src/command/WorkspaceCommand.swift
@@ -1,4 +1,5 @@
 import Common
+import Foundation
 
 struct WorkspaceCommand : Command {
     let info: CmdStaticInfo = WorkspaceCmdArgs.info
@@ -27,6 +28,12 @@ struct WorkspaceCommand : Command {
         }
         check(workspace.monitor.setActiveWorkspace(workspace))
         focusedWorkspaceName = workspace.name
+
+        // Firing Notification for e.g Sketchybar Integration        
+        let name = NSNotification.Name("\(Bundle.appId).WorkspaceFocus")
+        let userInfo = ["workspace": focusedWorkspaceName]
+        DistributedNotificationCenter.default().postNotificationName(name, object: nil, userInfo: userInfo)
+        
         return true
     }
 


MatthiasGrandl:
I tested it with Hammerspoon and Sketchybar and it works as expected.

This allows Sketchybar to subscribe to workspace changes.
```
sketchybar --add event aerospace_change bobko.aerospace.WorkspaceFocus
```

wojciech-kulik:
I'm testing this approach and I spotted one problem. If I click on the app from the Dock or if I use `cmd+tab` the notification is not triggered

src/command/WorkspaceCommand.swift
```diff
@@ -27,6 +28,12 @@ struct WorkspaceCommand : Command {
         }
         check(workspace.monitor.setActiveWorkspace(workspace))
         focusedWorkspaceName = workspace.name
+
+        // Firing Notification for e.g Sketchybar Integration        
+        let name = NSNotification.Name("\(Bundle.appId).WorkspaceFocus")
+        let userInfo = ["workspace": focusedWorkspaceName]
+        DistributedNotificationCenter.default().postNotificationName(name, object: nil, userInfo: userInfo)
```
nikitabobko, line 35: From the `userInfo` parameter docs:

> Sandboxed apps can send notifications only if they do not contain a dictionary. If the sending application is in an App Sandbox, userInfo must be nil. 

AeroSpace is not a Sandbox app. But if Apple makes it possible to use accessibility API in Sandboxed apps then I want to be able to make the app sandboxed.

Considering that users can obtain the currently focused workspace via CLI, I don't want to provide `userInfo` parameter

nikitabobko, line 35: The logic of posting notifications should be moved to `focusedWorkspaceName =` assignment. The current implementation doesn't post notifications when the workspace is changed via `cmd+tab`

MatthiasGrandl:
@wojciech-kulik Good point, I didn’t consider that.

src/command/WorkspaceCommand.swift
```diff
@@ -27,6 +28,12 @@ struct WorkspaceCommand : Command {
         }
         check(workspace.monitor.setActiveWorkspace(workspace))
         focusedWorkspaceName = workspace.name
+
+        // Firing Notification for e.g Sketchybar Integration        
+        let name = NSNotification.Name("\(Bundle.appId).WorkspaceFocus")
```
nikitabobko, line 33: I will replace `Bundle.appId` with `bobko.aerospace` because I want to have the same notification id for debug and release builds

nikitabobko (commented):
Merged with the following patch:
```diff
diff --git a/src/command/WorkspaceCommand.swift b/src/command/WorkspaceCommand.swift
index 2a73731..9300ab2 100644
--- a/src/command/WorkspaceCommand.swift
+++ b/src/command/WorkspaceCommand.swift
@@ -28,12 +28,6 @@ struct WorkspaceCommand : Command {
         }
         check(workspace.monitor.setActiveWorkspace(workspace))
         focusedWorkspaceName = workspace.name
-
-        // Firing Notification for e.g Sketchybar Integration
-        let name = NSNotification.Name("\(Bundle.appId).WorkspaceFocus")
-        let userInfo = ["workspace": focusedWorkspaceName]
-        DistributedNotificationCenter.default().postNotificationName(name, object: nil, userInfo: userInfo)
-
         return true
     }

diff --git a/src/focused.swift b/src/focused.swift
index f5ad750..6d48f26 100644
--- a/src/focused.swift
+++ b/src/focused.swift
@@ -29,10 +29,14 @@ private var _focusedWorkspaceName: String = focusedMonitorInaccurate?.activeWork
 var focusedWorkspaceName: String { // todo change to focused monitor
     get { _focusedWorkspaceName }
     set {
-        if newValue != _focusedWorkspaceName {
+        let changed = newValue != _focusedWorkspaceName
+        _focusedWorkspaceName = newValue
+        if changed {
+            // Firing Notification for e.g Sketchybar Integration
+            DistributedNotificationCenter.default().postNotificationName(NSNotification.Name("bobko.aerospace.focusedWorkspaceChanged"), object: nil)
+
             previousFocusedWorkspaceName = _focusedWorkspaceName
         }
-        _focusedWorkspaceName = newValue
     }
 }
 var previousFocusedWorkspaceName: String? = nil
```

src/command/WorkspaceCommand.swift
MatthiasGrandl: I am fine with the change, although there is a noticable latency in the sketchybar update when querying the current active workspace with `aerospace list-workspaces --focused` instead of getting it from the notification.

nikitabobko: Well, on my machine `aerospace list-workspaces --focused` is quick
```
~/a/AeroSpace/LocalPackage $ time .build/debug/aerospace list-workspaces --focused > /dev/null
.build/debug/aerospace list-workspaces --focused  0.00s user 0.00s system 6% cpu 0.093 total
```
And this is a debug build without optimizations

Is it quick on your machine? If so then I suspect that the bug is somewhere in sketchybar, and the issue should be addressed on their side

MatthiasGrandl: It’s probably the debug build then. For me `time` shows 0.01s

nikitabobko: Sorry, I don't get what you mean. Are you saying that `aerospace list-workspaces` is noticeably slower in debug compared to release? For me, 0.093s is not noticeable

Issue #74, opened 2023-12-29
`gaps.outer.top` vs the notch


johnallen3d:
Long, _long_ time Yabai user here trying out AeroSpace. After about one day I'm pretty sure I'm a convert. Brilliant work @nikitabobko!

That said, I have a question regarding `gaps.outer.top`. It seems that on the display with a "notch", the padding is in relation to the bottom of the notch. Whereas displays without a notch, the padding is in relation to the actual top of the display. This makes the gap between the top of a window and my bar (SketchyBar) drastically larger/smaller on each display. Is there any way to account for this (eg. display specific setting)?

Maybe a screenshot will help illustrate better than my description. The left side is my MacBook Pro and the right is my external.

![CleanShot 2023-12-29 at 18 24 53@2x](https://github.com/nikitabobko/AeroSpace/assets/184915/1306976a-1a15-44ac-8e5f-38318cc569a4)

nikitabobko:
> Is there any way to account for this (eg. display specific setting)?

Now it's not possible. But I agree that it'd be nice to have display specific settings for the gaps

johnallen3d:
Ok, I'm glad to know there wasn't something I was missing. 

I will say, for my use case, an option to disregard the notch would also suffice. I'd take a crack at a PR but I have _zero_ experience with Swift to date.

wojciech-kulik:
Alternatively, it would be nice to be able to change gaps using hotkey. It would be useful when I'm using 2 displays and I switch to 1.

nikitabobko:
> Alternatively, it would be nice to be able to change gaps using hotkey. It would be useful when I'm using 2 displays and I switch to 1.

What is the use case for workspace-specific gaps?

In my head, there are two use cases for gaps:
1. Aesthetics. In this case, users use the same gaps on all workspaces
2. Practical. To make it easier to distinguish between windows

wojciech-kulik:
@nikitabobko to solve the problem mentioned in this issue. When I work on two displays I don't mind much that the top gap is bigger on my side (built-in)  display, but when I switch to mac only then the gap is annoying and takes to much space, so I need to adjust my config to reduce it.

As you said, right now it's not possible to fix this bug, so it could work as a workaround temporarily. If the bug is fixed, then changing gaps dynamically is not needed either.

wojciech-kulik:
@johnallen3d I fixed it here: #98

wojciech-kulik:
[...]

PR #98, opened 2024-01-03
Fixed issue #74 - support for custom menu bars

wojciech-kulik:
It resolves issue #74. 

The main assumptions of this solution:
- if the system menu bar is visible, we want to calculate the top gap from it.
- if the system menu bar is hidden, we want to calculate the top gap from the screen top.

Detailed description of possible cases:
1. System menu bar is used and is visible. In this case, we want to calculate padding from the menu bar, because the menu bar height could differ per screen (the status bar for screens with a notch is bigger). For this purpose, I'm using the top offset provided by `visibleRect` calculating it this way: `visibleRect.topLeftY - rect.topLeftY`.
2. Custom menu bar is used. In this case, we would like to calculate the padding always from the screen top (ignoring notch) to allow users to have a consistent top padding across all screens to avoid issues like #74 where custom menu bars have the same height on all screens.
3. System menu bar is used and is hidden. This is the only case that is not perfect because for screens with a notch and the config with `0` top gap windows will be placed slightly under the notch. After setting top gap to for example `40` now we will have the same padding on all screens, causing an unwanted gap on external screens if we adjusted it to our screen with notch.

I think the third case is quite rare and I don't see any way to solve it without having a padding option per screen, so for now I think this approach is still better than the current one. It allows us to use AeroSpace both with the system menu bar and custom menu bar having expected top gap.

### DIFF
--- a/src/model/Monitor.swift
+++ b/src/model/Monitor.swift
@@ -4,6 +4,7 @@ private struct MonitorImpl {
     let name: String
     let rect: Rect
     let visibleRect: Rect
+    let safeAreaInsets: NSEdgeInsets
 }
 
 extension MonitorImpl: Monitor {
@@ -18,6 +19,7 @@ protocol Monitor: AeroAny {
     var visibleRect: Rect { get }
     var width: CGFloat { get }
     var height: CGFloat { get }
+    var safeAreaInsets: NSEdgeInsets { get }
 }
 
 class LazyMonitor: Monitor {
@@ -25,6 +27,7 @@ class LazyMonitor: Monitor {
     let name: String
     let width: CGFloat
     let height: CGFloat
+    let safeAreaInsets: NSEdgeInsets
     private var _rect: Rect? = nil
     private var _visibleRect: Rect? = nil
 
@@ -32,6 +35,7 @@ class LazyMonitor: Monitor {
         self.name = screen.localizedName
         self.width = screen.frame.width // Don't call rect because it would cause recursion during mainMonitor init
         self.height = screen.frame.height // Don't call rect because it would cause recursion during mainMonitor init
+        self.safeAreaInsets = screen.safeAreaInsets
         self.screen = screen
     }
 
@@ -45,7 +49,9 @@ class LazyMonitor: Monitor {
 }
 
 private extension NSScreen {
-    var monitor: Monitor { MonitorImpl(name: localizedName, rect: rect, visibleRect: visibleRect) }
+    var monitor: Monitor {
+        MonitorImpl(name: localizedName, rect: rect, visibleRect: visibleRect, safeAreaInsets: safeAreaInsets)
+    }
 
     var isMainScreen: Bool {
         frame.minX == 0 && frame.minY == 0
@@ -64,7 +70,12 @@ private extension NSScreen {
 }
 
 private let testMonitorRect = Rect(topLeftX: 0, topLeftY: 0, width: 1920, height: 1080)
-private let testMonitor = MonitorImpl(name: "Test Monitor", rect: testMonitorRect, visibleRect: testMonitorRect)
+private let testMonitor = MonitorImpl(
+    name: "Test Monitor",
+    rect: testMonitorRect,
+    visibleRect: testMonitorRect,
+    safeAreaInsets: .init()
+)
 
 /// It's inaccurate because NSScreen.main doesn't work correctly from NSWorkspace.didActivateApplicationNotification &
 /// kAXFocusedWindowChangedNotification callbacks.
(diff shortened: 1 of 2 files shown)


nikitabobko:
Thanks for the patch and detailed reasoning! But...

The solution is too smart. It tries to guess what users want. But there always will be cases when you won't guess correctly. The config should do exactly what it says.

Consider this example: what if users hide the macOS bar, but configure their bar to be shown below the notch. They would want to configure the top gap to be calculated from the notch

IMO, #74 should be fixed by padding option per screen

Currently, I have this solution in mind:
```toml
[gaps]
inner.horizontal = 0
inner.vertical = 0
outer.left = 0
outer.bottom = 0
outer.top = 0
outer.right = 0

[gaps."built-in"] # "built-in" is a monitor pattern as in https://nikitabobko.github.io/AeroSpace/guide.html#assign-workspaces-to-monitors
inner.horizontal = 0
inner.vertical = 0
outer.left = 0
outer.bottom = 0
outer.top = 0
outer.right = 0
```

Feel free to try to implement the "per monitor" approach, if you feel adventurous

wojciech-kulik:
I agree. As I mentioned, I consider this solution as a temporary one before introducing gaps per screen. It fixes the most common use cases.

However, if you feel that we should go directly for "gaps per screen", feel free to take it over. Unfortunately, for the next 2 weeks, I will be unavailable to fix it :(.

nikitabobko:
> I consider this solution as a temporary one before introducing gaps per screen. It fixes the most common use cases.

Good to hear that we agree on that. I'm closing this PR then. I won't merge it, sorry

> However, if you feel that we should go directly for "gaps per screen", feel free to take it over. Unfortunately, for the next 2 weeks, I will be unavailable to fix it :(

No problem. I will fix this issue myself sooner or later. But until then anyone is welcome to try

nikitabobko:
> Currently, I have this solution in mind:

Different idea is smth like this (draft):
```
[gaps]
inner.horizontal = 10
inner.vertical = 10
outer.left = 10
outer.bottom = 10
outer.top = [{ "built-in" = 0 }, { "another-monitor-pattern" = 5 }, 10] # The first matching from left to right is applied. 10 would be the default
outer.right = 10
```

> It's a good first contribution issue

I take it back

wojciech-kulik:
I think I could try implementing: https://github.com/nikitabobko/AeroSpace/pull/98#issuecomment-1876045739

However, I also think that it would be nice to have some automatic implementation like the one I provided in this PR. It could be something like `menu-bar-auto-detection` (default: false). What do you think? I think this flag would cover 90% of cases without need of manually setting up gaps per monitor

nikitabobko:
You can try to implement this syntax **(3)**:
```
[gaps]
inner.horizontal = 10
inner.vertical = 10
outer.left = 10
outer.bottom = 10
outer.top = [{ monitor."built-in" = 0 }, { monitor."another-monitor-pattern" = 5 }, 10]
outer.right = 10
```

Pro: in the future, it can be extended to support workspace-specific gaps `[{ monitor."built-in" = 0 }, { workspace.1 = 5 }, 10]`

There are two big problems with syntax **(1)** https://github.com/nikitabobko/AeroSpace/pull/98#issuecomment-1876045739
1. sets/tables are unordered in toml https://github.com/toml-lang/toml/issues/162#issuecomment-1816795384 and toml++ https://github.com/marzer/tomlplusplus/issues/62 (toml++ is an AeroSpace dependency). You won't be able to understand what `gaps` block to choose when several monitor patterns match
2. In **(1)**, you need to decide on a strategy when some values are absent. In **(3)**, you don't have that problem
```
[gaps]
inner.horizontal = 0
inner.vertical = 0
outer.left = 0
outer.bottom = 0
outer.top = 0
outer.right = 0

[gaps."built-in"] # Note: inner.vertical is absent. Should it be inherited or should it be config error?
inner.horizontal = 0
outer.left = 0
outer.bottom = 0
outer.top = 0
outer.right = 0
```
3. IMO, **(3)** has loose coupling, and high cohesion. Contrary to **(1)** which has high coupling and low cohesion https://en.wikipedia.org/wiki/Coupling_(computer_programming)

> I also think that it would be nice to have some automatic implementation like the one I provided in this PR

Sorry, I'm strongly against it. This is too magical option, and it complicates the logic too much.

wojciech-kulik:
Ok, I'll try with the 3rd approach

wojciech-kulik:
@nikitabobko any tips how to parse something like `{ monitor."built-in" = 0 }`?

nikitabobko:
> Is it monitor table with key built-in and value 0?

yes

wojciech-kulik:
Done in #117

PR #109, opened 2024-01-10
Update goodness to include a link to JankyBorders.adoc

GraemeARobinson:


nikitabobko:
Thanks! I'm curious, would it be useful to mention that `borders` CLI can be called `after-startup-command` callback, like so?
```
after-startup-command = 'exec-and-forget borders active_color=0xffe1e3e4 inactive_color=0xff494d64 width=5.0'
```

JankyBorders readme mentions the same for `yabairc`, but I'm unfamiliar with yabai. Is `yabairc` run on every relaunch/config update? If so, don't users get two JankyBorders processes running?

johnallen3d:
It's not uncommon for a `yabairc` to include a `killall borders` before launching the process (same goes for SketchyBar).

nikitabobko:
There is no need to `killall borders`. It has a built-in detection of already running process

johnallen3d:
Yeah, good point! I must have been thinking of SketchyBar. There was a time when this approach (kill and restart) was commonly used, but that experience has also improved.

PR #113, opened 2024-01-12
Update esc binding on resize mode in i3-like-config-example.toml

binamkayastha:
Using the default i3 like config, when in resize mode, clicking esc would set the mode to "esc mode". This mode does not exist. This change fixes it so that pressing esc in the resize mode goes back to the main mode.

Sidenote: Absolutely love this project. I've been searching for something like this for years. Thank you so much!


nikitabobko:
Thanks!

PR #118, opened 2024-01-15
Add flag to skip empty workspaces in workspace next/prev

ab5trakkt:
This patch adds a flag to jump to a non-empty workspace when using workspace next or workspace prev.


nikitabobko:
A more composable approach is to introduce `--empty [no]` option to `list-workspaces` command. `workspace (next|prev)` accepts stdin, so you can query workspaces which you want to cycle through:
```bash
aerospace list-workspaces --monitor focused --empty no | aerospace workspace next
# or if you want to "continuously" switch away from empty workspaces
(aerospace list-workspaces --monitor focused --empty no && aerospace list-workspaces --focused) | sort | uniq | aerospace workspace next
```

I'm closing this PR in favor of c346333b8cabb7900c2eeafaa6b28635aca48e67

PR #198, opened 2024-04-09
Add "Open config" Menu Item

gobijan:
Hi, 

I found myself to be experimenting a lot with the config. It is super handy to have a "Reload config" menu item at hand but I found myself to always jump to the terminal and open the config file. Before having the config at hand I had to copy it from the app into my home folder using as recommended in the Readme.

This PR adds a menu item to open the config file directly. It also copies it initially from the App to the home folder when no config is present.

It's a quality of life improvement for me. Would be happy to have this in AeroSpace :)

Cheers Bijan


nikitabobko:
Thanks for the proposal!

Merged as:
1. https://github.com/nikitabobko/AeroSpace/commit/ca8fcaec8aa274e8d961ab72f8b6c96d77e939ab
2. https://github.com/nikitabobko/AeroSpace/commit/82c2291bd1b945b62a18a0a2a0f7a798522eb71c

PR #213, opened 2024-05-05
fix(docs): Fix Colemak keyboard distribution

martintdp:
The mapping seems to be flipped.
With the default configuration and only
including the mappings, `alt-i` would change to workspace "u" (by mapping to `alt-u`) instead of
moving focus right (as expected of `alt-l`)

This commit addresses the issue by inverting the mapping declaration.


nikitabobko:
Are you sure that your PR is correct? I have double-checked it and the mapping seems to be correct

Keys of the mapping are key notations and values of the mapping are key codes

E.g. for key `i` in Colemak has key code `l`

docs/goodness.adoc
```diff
@@ -163,37 +163,37 @@ cmd-h = []
...
 
 z = 'z'
 x = 'x'
 c = 'c'
 v = 'v'
 b = 'b'
-k = 'n'
-r = 'm'
+n = 'k'
+m = 'r'
```
omar25h, line 196: I'm using the Colemak layout and the current settings are almost correct. As a matter of fact, I only had to modify this line to avoid having duplicate mapping (`r = 's'` on line 179 and `r = 'm'` here
```suggestion
m = 'm'
```

nikitabobko: > I only had to modify this line to avoid having duplicate mapping (r = 's' on line 179 and r = 'm' here)

Thanks! That was a typo indeed. Should be fixed now https://github.com/nikitabobko/AeroSpace/commit/74dc45ec

PR #234, opened 2024-05-21
docs: add alfred workflow

gogamid:


nikitabobko:
Thanks for the PR!

But the workflow you created is far from production-ready. Running only commands that do not require options is not enough to be useful for other peoples

Feel free to resubmit the PR once it becomes more mature

PR #255, opened 2024-06-07
Fix typo + links to on-window-detected on default-config.toml

cabello:
Hello there,

Giving this app a try and I saw the typo in the docs and also I needed to make 1password and iterm2 (using in "quake" mode) float by default, so I added some docs to the default configuration as well.

Hope this is useful for future users.

Thank you for putting this app together. 💜


nikitabobko:
Merged the first commit

Skipped the second commit. All complicated options are mentioned in the guide, and there is no reason to pollute `default-config.toml`

PR #275, opened 2024-06-12
Reduce workspace switching flicker

jilyaluk:
Previously, workspace switch might flicker a bit upon changing to a higher number. This could be observed by switching repeatedly between first and second workspace, for example. 2->1 does not flicker, but 1->2 sometimes shows desktop. This fixes the issue: visible workspaces are first unhidden, and then invisible ones are hidden.


nikitabobko:
Thanks! Merged with minor changes from me

Issue #333, opened 2024-07-17
Long setup jitter when plugging in multiple monitors together


Solomonz:
Context:

❯ aerospace -v
aerospace CLI client version: 0.11.1-Beta a8fe49f8103c7840f2e2b5497bade7f63535c139
AeroSpace.app server version: 0.11.1-Beta a8fe49f8103c7840f2e2b5497bade7f63535c139

[.aerospace.toml](https://github.com/user-attachments/files/16273167/aerospace.toml.txt)

I have two monitors (both are Samsung LF32TU874VNXGO Thunderbolt T3/USB-C 3850x2160), A and B. My desk setup is I have A in a vertical orientation on the left (code monitor), connected via thunderbolt 4 daisychain to B in a horizontal orientation in the middle (main monitor), connected via thunderbolt 4 to my laptop on the right (messaging, spotify, etc). This means that when I sit down at my desk, I can just plug my laptop in to the single thunderbolt cable, and get charging, both monitors, ethernet (plugged into B), and my USB A keyboard (also plugged into B).

Repro:

Video: https://github.com/user-attachments/assets/2b9643af-ccf2-4ef0-9819-b2c5e7d02a6f

As you can see, B is detected first, and there's no issue there. However, A is quickly detected afterwards, and when Aerospace tries to redistribute windows at that point, it gets into this frustrating loop that takes a while to get past. In this example, it took around 16 seconds from detecting the third monitor to finishing the setup jitter, but other times it has taken even longer, sometimes even taking a full 45-60 seconds before settling down. I did find that unplugging A from B, plugging B into my laptop, waiting a few seconds, and then plugging A into B does significantly reduce the jitter time, but that isn't a great solution.

At first I thought this was related to #289, but I see that that seems to have been isolated to a change after 0.11.2, whereas I am having this issue on 0.11.1.

Let me know if there's any more info I can provide, and thank you for this (otherwise) amazing tool!

nikitabobko:
Wow, that's crazy. I use only a single external monitor and I regularly plug and unplug it, I have never seen a problem like this

1. Is the problem reproducible only with a single external monitor?
2. Is the problem reproducible if you disable "Displays have separate spaces"?
3. Is the problem reproducible if you drop `workspace-to-monitor-force-assignment` from the config?

Solomonz:
[...]

Issue #159, opened 2024-02-12
Assign application to specific workspace


nick79:
Hi,

Is it possible to somehow permanently assign some application (for example Slack) to the specific workspace (for example Workspace 2)? I checked both provided configurations but didn't find any example, nor did I find it in the provided documentation. Probably, I miss something.

Thanks in advance.

nikitabobko:
https://nikitabobko.github.io/AeroSpace/guide.html#on-window-detected-callback

```
[[on-window-detected]]
if.app-id = 'com.tinyspeck.slackmacgap'
run = 'move-node-to-workspace S'
```

PR #334, opened 2024-07-17
docs: Add Monitor resolution impacts into the guide

NielsPeeters:
I've been using Aerospace for the last couple months, while it worked great in office, it didn't at home where I have 2 different monitors with different screen resolutions.
As a result I've been having performance impacts, where it took long for my windows to switch between workspaces.
After stalking the the issues tab I noticed #333 and in that conversation #159 and wished this was somewhere included into the guide/docs, either as a warning or in the comparison. 

+ cleanup

### DIFF
--- a/docs/guide.adoc
+++ b/docs/guide.adoc
@@ -365,9 +365,13 @@ Overview of '`Displays have separate Spaces`'
 |😡 No
 |👍 Yes
 
-|macOS status bar ...
-|... is displayed on both monitors
-|... is displayed only on main monitor
+|macOS status bar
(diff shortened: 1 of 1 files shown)


NielsPeeters:
Initially it seemed to work fine with `Displays have separate Spaces`' disabled, but now I'm again hitting delays when pressing keyboard shortcuts and Aerospace hanging, (loading when hovering over the menubar) to then all of a sudden catch up to switch between all spaces.

Pr can be ignored for now, will create a defect with example, later today when I reproduce it again.

PR #362, opened 2024-08-06
default-config: Do not use -n for terminal

graywolf:
With -n, the terminal apps just keeps accumulating in the dock.

So use -a combined with argument of ~ to open new window of default terminal.


Grezzo:
I approve of this. I couldn't figure out why I was getting so many terminals in my dock.

tobi:
Solution: 

```
ctrl-alt-cmd-enter = ['''exec-and-forget osascript <<EOF
tell application "iTerm2"
    create window with default profile
    activate
end tell
EOF''']

```

Issue #224, opened 2024-05-10
[Feature] Alias for combinations


AlgusDark:
Hi, thanks for this awesome project, I just discovered and I'm feeling that it's super great at this stage and I'm excited for it's future. There is this feature I would like to know if it's possible. 

I have a hyper key (`ctrl+alt+shift+cmd`) that I would like to define so I can do `hyper-c` instead of defining the all 4 modifiers in every definition of a combo.

Thanks!

nikitabobko:
Hi, 

the issue is reasonable, but it's a minor priority for me. I myself will be occupied with more major issues.

The currently existing `key-notation-to-key-code` syntax in the config can be reused:

```
[key-mapping.key-notation-to-key-code]
hyper = 'ctrl-alt-shift-cmd'

[mode.main.binding]
hyper-c = 'workspace c'
```

gldtn:
> Hi,
> 
> the issue is reasonable, but it's a minor priority for me. External contributions are welcome.
> 
> The currently existing `key-notation-to-key-code` syntax in the config can be reused:
> 
> ```
> [key-mapping.key-notation-to-key-code]
> hyper = 'ctrl-alt-shift-cmd'
> 
> [mode.main.binding]
> hyper-c = 'workspace c'
> ```

Has this changed since your reply? Doesn't seem to work for me:
```
Failed to parse /Users/gldtn/.config/aerospace/aerospace.toml

key-mapping.key-notation-to-key-code.hyper: 'ctrl-alt-shift-cmd' is invalid key code

mode.main.binding.hyper-b: Can't parse modifiers in 'hyper-b' binding
```

Thanks!

bcremer:
@gldtn It's not a implemented feature yet.

@nikitabobko only suggested that the already existing config key (`key-notation-to-key-code`) can by used should anybody implement that feature.

gldtn:
> @gldtn It's not a implemented feature yet.
> 
> @nikitabobko only suggested that the already existing config key (`key-notation-to-key-code`) can by used should anybody implement that feature.

I see, I miss understood his reply. Thank you!

YANOUSHek:
I know this is not the solution proposed by @nikitabobko but I don't think there's a need for anything more than `hyper` and `meh` modifiers. So I created #389 to have the feature request done.

FrancoLab:
Any progress on this. Would be an awesome feature.

EttVenter:
[...]

PR #389, opened 2024-08-12
Added support for hyper and meh modifiers

YANOUSHek:
There's two popular custom modifier keys called `hyper` and `meh` which are `cmd+alt+shift+ctrl` and `alt+shift+ctrl` respectively.

I've added two new modifier types to the `modifiersMap` global variable. This allows for shorthand syntax in the config file: `hyper-h` instead of having to write `cmd-ctrl-alt-shift-h`.

### DIFF
--- a/Sources/AppBundle/config/keysMap.swift
+++ b/Sources/AppBundle/config/keysMap.swift
@@ -201,6 +201,8 @@ let modifiersMap: [String: NSEvent.ModifierFlags] = [
     "alt": .option,
     "ctrl": .control,
     "cmd": .command,
+    "meh": [.shift, .control, .option],
+    "hyper": [.command, .shift, .control, .option],
 ]
 
 extension NSEvent.ModifierFlags {


lentil32:
nice

bcremer:
Not sure how "standardized" the definitions for `hyper` and `meh` combinations are.

I'm using https://hyperkey.app/ and the options are `cmd+alt+ctrl` or `cmd+alt+ctrl+shift` but not `alt+shift+ctrl`.

 image

jakenvac:
I think traditionally OP is correct in their definitions for `hyp` and `meh`. It's been this way for as long as I can remember and is standard in multiple keyboard firmwares:
- https://github.com/qmk/qmk_firmware/blob/master/docs/keycodes.md (Look at `KC_HYPR` and `KC_MEH`)
- https://github.com/KMKfw/kmk_firmware/blob/main/docs/en/keycodes.md#modifiers (KC.HYPR, KC.MEH)
- https://ergodox-ez.com/our-open-source-firmware (Based on QMK, uses hyper & meh as selling points)

To me, that screenshot of the hyperkey app just seems like it's giving you the option of including shift in your hyper key combo or not. It doesn't seem concerned with the `meh` key at all (`alt+shift+ctrl`), so I wouldn't say that is evidence against it being standard.

That being said, do we really need a specific name for these key combos in every program as long as they handle `cmd-alt-ctrl-shift-*` just fine?

camiloaromero23:
@nikitabobko any updates on this? 🤔

nikitabobko:
Thanks for the proposal, but sorry, I don't like the idea to hardcode `hyper` and `meh`. The next thing users are going to ask is to hardcode another combination of keys.

The proper solution is suggested here: https://github.com/nikitabobko/AeroSpace/issues/224#issuecomment-2111556500

YANOUSHek:
@nikitabobko to be honest I think it's far easier to add all the combinations here rather than having some custom parsing logic code to maintain. Especially since all the APIs that are being used here are system ones and they support assigning multiple modifiers. But ultimately it's your call.

I'll try to find some time to do it through `key-notation-to-key-code`.

PR #480, opened 2024-08-30
Closes 479 - Remove launchctl calls in startAtLogin routine

### LINKED ISSUE #479: [BUG REPORT] `start-at-login` is technically broken, but preventing a worse bug
The startup at login routine for configuring the launch agent plist & loading/unloading it is technically not working as intended.

### The Bug

Both the commands to [load and unload](https://github.com/nikitabobko/AeroSpace/blob/main/Sources/AppBundle/config/startAtLogin.swift#L27-L31) the plist fail with IO errors as the paths given to them aren't valid - thus the agent is never loaded or unloaded by the app.

`url.absoluteString` keeps the `file://` scheme from the URL instance that is created to locate the plist. launchctl just expects a plain path. To *correct* this behavior `URL.path` should be used instead. However this introduces a messier bug - more on this below.

The reason this behavior still works as well as it does is because the plist is written/deleted when app starts up/reloads config depending on the `start-at-login` value. When you next log in, the plist is automatically loaded by launchd or is missing so cannot be launched.

Running AeroSpaceApp from the terminal shows the following errors:
```sh
# output when the load fails:
Load failed: 5: Input/output error
Try running `launchctl bootstrap` as root for richer errors.

# output when unload fails:
Unload failed: 5: Input/output error
Try running `launchctl bootout` as root for richer errors.
```
You can see slightly more detailed info in the launchd logs:
```sh
# here's one such example for a failed 'load'
2024-08-30 16:58:28.718503 (gui/501 [100018]) <Notice>: Bootstrap by launchctl[21729] for //file:///Users/jake/Library/LaunchAgents/bobko.aerospace.plist failed (2: No such file or directory)
```

### The Messy Bug
When `launchctl load some.plist` is ran and `RunAtLoad` is `true` in the plist, the program is immediately executed - this creates a second instance of AeroSpace and then everything gets wild. Hotkeys are handled twice, windows are moved erratically, and things become generally unpredictable. In my specific scenario, I had things set up in such a way that both instances were fighting over a window and I couldn't focus my terminal to kill the app. Don't ask me how because I'm not sure.

There are many solutions to this problem, such as a seperate startup helper app, creating a lockfile to ensure single instances etc. However I feel this is overkill, at least for a first pass solution.

I think for now, it's safe to remove the calls to launchctl and just rely on the [lack of] presence of the plist and let macos do it's thing on login.

As an aside, truly fixing the launchd integration could bring benefits such as auto restarting the app after an update. #352

Sorry for the verbosity of this bug report. It's been a long day :)

## Checklist

- [X] I've searched for duplicates. My search queries were (please don't forget to search in closed issues too):
  - "log in"
  - "start-at-login"
  - "launchd"
- [ ] I've read the documentation https://nikitabobko.github.io/AeroSpace/guide & https://nikitabobko.github.io/AeroSpace/commands
  - N/A
- [ ] I've searched in the documentation. My search queries were:
  - N/A

```
aerospace CLI client version: 0.14.1-Beta 5861e85b65cb7dc914a0a68027c56e2bc90510fd
AeroSpace.app server version: 0.14.1-Beta 5861e85b65cb7dc914a0a68027c56e2bc90510fd
```
Also happens on main as of 9d8362d04f449c4fdaab0fb2d305e6fe0f4ca48a

jakenvac:
closes #479 

As described in #479, the calls to launchctl are broken due to invalid file paths, but *fixing* them introduces a worse bug.

As macos/launchd will automatically load a plist if it's present in the launch agents directory on load, it's enough that we simply write/delete the file as needed. For now at least, as there is no regression in functionality.

In future, if we wanted tighter control over the app as a launch agent we could explore enforcing a single instance of the app or introducing a startup helper to manage the lifecycle of AeroSpaceApp.

No hard feelings if you don't like this solution, I know we've not discussed it prior.

Update:
I'd not thought of this sooner, so it's not mentioned in the issue, but if you did want to fix the launchctl calls and keep them - a simple option to prevent the duplicate app bug would be to update the launch agent plist to run:
```sh
open -W /Applications/AeroSpace.app --args --started-at-login
``` 
instead of 
```sh
/Applications/Aerospace.app/Contents/MacOS/AeroSpace --started-at-login
```

As by default, Mac OS app bundles only allow one instance to run at a time.

However with this option, without some kind of guard, you'd still see errors on the unload when the plist file doesn't exist and start-at-login is false. And you'd see the load errors when the agent is already loaded and `start-at-login` is true.

### DIFF
--- a/Sources/AppBundle/config/startAtLogin.swift
+++ b/Sources/AppBundle/config/startAtLogin.swift
@@ -26,9 +26,7 @@ func syncStartAtLogin() {
             """
         Result { try plist.write(to: url, atomically: false, encoding: .utf8) }
             .getOrThrow("Can't write to \(url) ")
-        Result { try Process.run(URL(filePath: "/bin/launchctl"), arguments: ["load", url.absoluteString]) }.getOrThrow()
     } else {
-        Result { try Process.run(URL(filePath: "/bin/launchctl"), arguments: ["unload", url.absoluteString]) }.getOrThrow()
         try? FileManager.default.removeItem(at: url)
     }
 }


nikitabobko:
Merged

I knew that something wrong is going on here, but I was too lazy to properly investigate, so your comments were helpful, thanks!

While we are here, IMO it's better to put your great PR description directly to the commit message (the PR description itself can be empty). I consider the commits being the source of truth since they are more likely to stay for future generations than the GitHub hosting.

BTW, there is a new API since macOS 13 `SMAppService.register`. I don't know how it compares to the `LaunchAgents` solution that we use here

jakenvac:
noted, i'll try to remember to put the descriptions in the PR. Thanks for merging.

I also saw that new API while digging into this, i didn't look into much but i *think* it might only work for sandboxed apps. I'll do some more digging.

Issue #103, opened 2024-01-06
Feature request: introduce `if.is-dialog` in `on-window-detected` callback


nikitabobko:
Users tend to write callbacks like this:
```toml
[[on-window-detected]]
if.app-id = 'com.jetbrains.intellij'
run = 'move-node-to-workspace 2'
```

The problem with callbacks like this is that if during the work, users move the window to a different workspace, and then decide to open some dialog, the dialog will be moved back to workspace 2

It would be nice to make it possible to ignore these callbacks for dialogs. 

1. One way of doing this could be to introduce `if.is-dialog`. The problem with this solution is that it is not "enabled" by default. Users would need to remember that they always need to write `if.is-dialog = false`
2. Another crazy solution could be to introduce an option for `move-node-to-workspace` command `[--if-dialog [no]]`. And say that `--if-dialog no` is enabled by default in `on-window-detected` callback

nikitabobko:
Right now, I lean towards the solution 2 rather than the solution 1

I like to think that `on-window-detected` callback is the thing that considers different cases of [ADT](https://en.wikipedia.org/wiki/Algebraic_data_type). `if.is-dialog = false` will skip the callback, and AeroSpace will try to apply further callbacks. But the expected behavior is to stop searching for further callbacks

nikitabobko:
On the other hand, if users use "floating" by default, and then per selected apps say that the apps are tiling then `if.is-tiled-by-default = true` would be exactly what they need

Maybe both solutions need to be implemented

BTW, it would be cool to come up with a better name. `if.is-tiled-by-default` is ugly

wojciech-kulik:
As you mentioned
> The problem with callbacks like this is that if during the work, users move the window to a different workspace, and then decide to open some dialog, the dialog will be moved back to workspace 2

The problem is with child windows.

Wouldn't be better to implement `if.first-window` or something like that? (as mentioned in #71)

nikitabobko:
> The problem is with child windows.

Not necessarily. Consider the case when users have more than one projects opened in Xcode or IntelliJ. It's more than one windows that are perceived as "main" windows but the algorithm of `is-leader-window` ("child" windows) suggested in #71 will only match one of them. 

What also makes me suspicious about `is-leader-window` is that I've never heard about analogical features in other tiling WMs

wojciech-kulik:
> What also makes me suspicious about is-leader-window is that I've never heard about analogical features in other tiling WMs

[...]

PR #565, opened 2024-10-04
Add is-dialog function to on-window-detected callback

GeorgeGo:
Add handling for is-dialog to parseOnWindowDetected.swift 
Add shouldFloat and !isWindow handling to WindowDetectedCallbackMatcher in MacWindow.swift

In response to Issue #103


nikitabobko:
Unanswered design questions in #103 should be resolved before writing any code

PR #569, opened 2024-10-05
Echo failed dependency in add-to-bin

nickcurran:
`build-debug.sh` and `run-debug.sh` fail silently when a dependency is missing, e.g., `fish`, `rust`, `xcbeautify`. This adds an output message indicating when a dependency is missing, e.g., `Dependency xcbeautify not found`.

### DIFF
--- a/script/setup.sh
+++ b/script/setup.sh
@@ -4,14 +4,19 @@ set -u # Treat unset variables and parameters other than the special parameters 
 set -o pipefail # Any command failed in the pipe fails the whole pipe
 # set -x # Print shell commands as they are executed (or you can try -v which is less verbose)
 
+# Look for a given dependency, failing the whole script if not found.
+# If found, create a mapping script to that dependency in the .deps/bin folder
 add-to-bin() {
-    /usr/bin/which "$1" &> /dev/null && \
-        cat > ".deps/bin/${2:-$1}" <<EOF
+    /usr/bin/which "$1" &> /dev/null || (echo "Dependency $1 not found" && exit 1)
+    
+    cat > ".deps/bin/${2:-$1}" <<EOF
 #!/bin/bash
 exec '$(/usr/bin/which "$1")' "\$@"
 EOF
 }
 
+# Delete the dependency mapping scripts from .deps/bin, recreate them,
+# and establish our PATH with .deps/bin in it
 nuke-path() {
     /bin/rm -rf .deps/bin
     /bin/mkdir -p .deps/bin


nikitabobko:
It's by design that it doesn't fail. All of those dependencies are optional (e.g. you only need `fish` if you want to build shell completion)

> build-debug.sh and run-debug.sh fail silently when a dependency is missing

Are you sure? I just tried to run `run-debug.sh` with the following patch, and it run without problems. Maybe you mean `build-release.sh`?

```diff
diff --git a/script/setup.sh b/script/setup.sh
index 49bb2998..0b51f141 100644
--- a/script/setup.sh
+++ b/script/setup.sh
@@ -6,7 +6,7 @@ set -o pipefail # Any command failed in the pipe fails the whole pipe
 
 add-to-bin() {
     /usr/bin/which "$1" &> /dev/null && \
-        cat > ".deps/bin/${2:-$1}" <<EOF
+        /bin/cat > ".deps/bin/${2:-$1}" <<EOF
 #!/bin/bash
 exec '$(/usr/bin/which "$1")' "\$@"
 EOF
@@ -16,18 +16,18 @@ if /bin/test -z "${NUKE_PATH:-}"; then
     /bin/rm -rf .deps/bin
     /bin/mkdir -p .deps/bin
 
-    add-to-bin bash not-outdated-bash # build-shell-completion.sh
-    add-to-bin brew # install-from-sources.sh
-    add-to-bin bundle # Ruby, asciidoc
-    add-to-bin bundler # Ruby, asciidoc
-    add-to-bin cargo
-    add-to-bin fish
-    add-to-bin git
-    add-to-bin rustc
-    add-to-bin xcbeautify
+    # add-to-bin bash not-outdated-bash # build-shell-completion.sh
+    # add-to-bin brew # install-from-sources.sh
+    # add-to-bin bundle # Ruby, asciidoc
+    # add-to-bin bundler # Ruby, asciidoc
+    # add-to-bin cargo
+    # add-to-bin fish
+    # add-to-bin git
+    # add-to-bin rustc
+    # add-to-bin xcbeautify
 
     export PATH="${PWD}/.deps/bin:/bin:/usr/bin"
-    chmod +x .deps/bin/*
+    # chmod +x .deps/bin/*
     export NUKE_PATH=1
 fi
```

nickcurran:
Yes, I'm very sure which commands I'm running that fail before my suggested changes: run-debug.sh and build-debug.sh.

Is it your intent that I comment out the dependencies that I don't intend to install? It would be easier if add-to-bin didn't fail silently - maybe don't fail at all and show a warning?

PR #605, opened 2024-10-20
[focus] add `prev` and `next` as directions to focusCommand

raisjn:
#### Description

Adds `prev` and `next` as directions to `focus` command. The specific benefit here is that they don't rely on cardinal directions and that they wrap around when they get to the first or last window. 

This is similar to how xmonad's [window cycling feature behaves](https://hackage.haskell.org/package/xmonad-contrib-0.18.1/docs/XMonad-Actions-CycleWindows.html)

#### Test Plan

* Load a workspace with multiple windows, verify `next` and `prev` cycle between them and don't stop at the end.
  * Verify in accordion and tiles
* Load a workspace with no windows, verify program doesn't crash

#### Notes

* Please edit the commit / commit message as desired
* I did not regen docs, wasn't able to get it to work properly

### DIFF
--- a/Sources/AppBundle/command/impl/FocusCommand.swift
+++ b/Sources/AppBundle/command/impl/FocusCommand.swift
@@ -17,6 +17,9 @@ struct FocusCommand: Command {
 
         switch args.target {
             case .direction(let direction):
+                if (direction == .next || direction == .prev) {
+                    return focusPrevOrNext(target, direction, io)
+                }
                 let window = target.windowOrNil
                 if let (parent, ownIndex) = window?.closestParent(hasChildrenInDirection: direction, withLayout: nil) {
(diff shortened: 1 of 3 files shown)


nikitabobko:
I appreciate the effort, but the PR doesn't meet the desired quality bar

raisjn:
no problem, thanks for looking!

btctrl:
Bump, I was just trying to find this feature as I forget all the time whether i need to focus left or right when stacked.

Is this being tracked elsewhere?

jakeswenson:
> I appreciate the effort, but the PR doesn't meet the desired quality bar

Curious **what** is missing in the quality of this? documentation updates? tests? How could someone revive it and take it further? 

Or is it not a feature that would be accepted? 

I personally would love to have this for my setup.

PR #606, opened 2024-10-20
[draft][summon-workspace] summon workspace works on non-focused monitors

raisjn:
#### Description

If the workspace has an assignedMonitor, it pulls that workspace onto its assigned monitor but the focus is retained on the current monitor. If the workspace is visible, nothing happens and focus stays on the current monitor

**use case**

You may keep a terminal on one monitor and want to cycle what is visible on the other monitor while staying in the terminal. This fixes the issue of rapidly cycling workspaces causing the monitor focus to shift. When wanting to focus a monitor, `focus-monitor` is used instead. 

#### Test Plan

[1,2,3] are main, [8,9,0] are secondary, [U,Z] have no assigned monitor

* [1, 0*] - summon 2 -> [2, 0*] (summons 2 to pinned monitor)
* [1, 0*] - summon U -> [1, U*] (summons U to current monitor)
* [U, 0*] - summon U -> [U, 0*] (nothing happens because 0 is pinned)
* [U, Z*] - summon U -> [Z, U*] (swap workspaces)

#### Notes

* feel free to edit the commit
* feel free to ignore this commit, this is just how i configured the behavior for myself

### DIFF
--- a/Sources/AppBundle/command/impl/SummonWorkspaceCommand.swift
+++ b/Sources/AppBundle/command/impl/SummonWorkspaceCommand.swift
@@ -7,15 +7,23 @@ struct SummonWorkspaceCommand: Command {
     func run(_ env: CmdEnv, _ io: CmdIo) -> Bool {
         check(Thread.current.isMainThread)
         let workspace = Workspace.get(byName: args.target.val.raw)
-        let monitor = focus.workspace.workspaceMonitor
+        let onMonitor = monitors.first { $0.activeWorkspace == workspace }
+        if onMonitor != nil {
+            io.err("Workspace '\(workspace.name)' is already visible on a monitor, returning")
+            return !args.failIfNoop
+        }
+
+        let monitor = focus.workspace.forceAssignedMonitor ?? focus.workspace.workspaceMonitor
         if monitor.activeWorkspace == workspace {
             io.err("Workspace '\(workspace.name)' is already visible on the focused monitor. Tip: use --fail-if-noop to exit with non-zero code")
             return !args.failIfNoop
         }
-        if monitor.setActiveWorkspace(workspace) {
-            return workspace.focusWorkspace()
+
+        if isValidAssignment(workspace: workspace, monitor: monitor) {
+            _ = monitor.setActiveWorkspace(workspace)
+            return monitor.activeWorkspace.focusWorkspace()
         } else {
-            return io.err("Can't move workspace '\(workspace.name)' to monitor '\(monitor.name)'. workspace-to-monitor-force-assignment doesn't allow it")
+            return workspace.forceAssignedMonitor?.setActiveWorkspace(workspace) != nil
         }
     }
 }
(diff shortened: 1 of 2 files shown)


nikitabobko:
> If the workspace has an assignedMonitor, it pulls that workspace onto its assigned monitor but the focus is retained on the current monitor

It's yet unclear to me if it should be the default behavior. At least, it should be configurable to add possibility to focus the "pinned" workspace

```toml
alt-1 = 'summon-workspace 1 --reveal-if-force-assigned' # just make it visible, as you suggest
alt-1 = 'summon-workspace 1 --focus-if-force-assigned' # focus it
```

How does Xmonad behave? Does Xmonad even have a notion of force assigned workspaces?

Right now, I lean towards: it should be an error by default (as it is now), and the error message should mention `--reveal-if-force-assigned` and `--focus-if-force-assigned` flags to make the flags more discoverable.

> [U, 0*] - summon U -> [U, 0*] (nothing happens because 0 is pinned)

Thanks for bringing this example! I didn't think about it. Alternatively, it could work this way: [U, 0*] - summon U -> [1, U*]

But given: "workspaces try to swap if both of them are visible" mantra, the feature interaction of `workspace-to-monitor-force-assignment` and `summon-workspace` becomes tricky, I don't have an answer myself yet

---

And sorry to mention it, but the quality of this PR is subpar too. I've not yet tried to understand the code you wrote, but the PRs should at least keep the CI green

nikitabobko:
> I don't have an answer myself yet

Maybe it should be a separate flag `--swap-if-both-are-visible` (the flag should be incompatible with `--focus-if-force-assigned`/`--reveal-if-force-assigned`)

Your 3rd example got me thinking that existing `workspace-to-monitor-force-assignment` feature makes it impossible to make AeroSpace behave like Xmonad by default.

raisjn:
> And sorry to mention it, but the quality of this PR is subpar too. I've not yet tried to understand the code you wrote, but the PRs should at least keep the CI green

No problem - I should have marked this draft - my intent is to start a conversation + demonstrate the behavior I am using. I will push my changes to a fork in the future

raisjn:
> Your 3rd example got me thinking that existing workspace-to-monitor-force-assignment feature makes it impossible to make AeroSpace behave like Xmonad by default. 

Yes, that is what I have been thinking about and why I implemented it this way - I've never used Xmonad with pinned screens.  I came up with this solution because it feels closest to how I was using Xmonad. I was trying to accomplish: "when using workspace switching, it should never move the mouse off the current monitor". 

> And sorry to mention it, but the quality of this PR is subpar too. I've not yet tried to understand the code you wrote, but the PRs should at least keep the CI green

No problem - I should have marked this draft - my intent is to start a conversation + demonstrate the behavior I am using. I will push my changes to a fork in the future if I need to demonstrate any code. 

A hanging semi-colon which caused the linter to fail in CI. I ran tests locally and they pass (using `swift test`), but the scripts in your repo don't work for me. They silently exit during the `source ./script/setup.sh` step - this is a point of friction which is making it difficult to install dependencies, automatically run tests, generate docs and run the linters. I'm not sure if it is because I'm running zsh or some other reason.

nikitabobko:
> No problem - I should have marked this draft - my intent is to start a conversation + demonstrate the behavior I am using. I will push my changes to a fork in the future if I need to demonstrate any code.

Thanks for understanding, and sorry if I sounded too harsh. If someody wants to show a snippet, it's better to attach it as patch in the issue comment, or give a link to the commit in their fork. PRs create an impression that the person expects the commit to be merged, and I expect PRs to be polished to an acceptable degree. Red CI is a big red flag that the person didn't even run tests.

> but the scripts in your repo don't work for me. They silently exit during the source ./script/setup.sh step

Thanks for mentioning! Should be fixed in the `main` branch

Issue #592, opened 2024-10-14
Add `%{app-exec-path}` and `%{app-bundle-path}` interpolation variable to `--format` flag in query commands


nikitabobko:
Query commands to add it to:
- `list-apps`
- `list-windows`

```
$ aerospace list-apps --format '%{app-exec-path}'
/Applications/Google Chrome.app/Contents/MacOS/Google Chrome
/System/Library/CoreServices/Finder.app/Contents/MacOS/Finder

$ aerospace list-apps --format '%{app-bundle-path}'
/Applications/Google Chrome.app
/System/Library/CoreServices/Finder.app
```

Use case: Get app icon for spotlight-like applications https://github.com/raycast/extensions/pull/14908/files#diff-bb1e762a64d9a33559e2e0481c4980cea9f69d54fa3dcc1fd9e5fac95e0bb5e5R21

nikitabobko:
Related: #359

nikitabobko:
Released in 0.16.0-Beta

PR #612, opened 2024-10-24
Add interpolation variables to --format flag

dzanotto:
Add %{app-exec-path} and %{app-bundle-path} interpolation variable to --format flag in query commands list-apps and list-windows
#592


nikitabobko:
Thanks! Merged

Issue #614, opened 2024-10-25
[Suggestion] Using Window Identifiers to determine special windows


hackrmomo:
I'm using [ghostty](https://mitchellh.com/writing/ghostty-is-coming) as my terminal and it supports a mode for a Quick Terminal (Quake-Style Terminal) that needs to be floated within AeroSpace. There's no easy way of specifying this behaviour at the moment. My current suggestion I've proposed is to effectively attach a window identifier and then filter them inside of various other apps (including AeroSpace) for more specific functionality. In the case of ghostty, it would simply float the Quick Terminal window despite belonging to the `com.mitchellh.ghostty` app id.

Edit: This would also potentially be useful for apps like Arc Browser for #611 

Draft PR: #615 

## Checklist

- [x] I've searched for duplicates. My search queries were (please don't forget to search in closed issues too):
  - ghostty
  - identifiers
  - identifier
- [x] I've read the guide https://nikitabobko.github.io/AeroSpace/guide
- [x] I've searched in the documentation (https://nikitabobko.github.io/AeroSpace/guide & https://nikitabobko.github.io/AeroSpace/commands). My search queries were:
  - N/A - this is a suggestion

```
(custom build - not referring to a current build version of AeroSpace)
```

hackrmomo:
(Also if you have access to the ghostty repository, this is under pull request #2489)

nikitabobko:
> there is a discussion around this under pull request #2489

https://github.com/ghostty-org/ghostty/pull/2489

nikitabobko:
Closing since it was fixed on Ghostty side. Thanks for the fix!

thuvasooriya:
@nikitabobko @m-ahdal sorry for the ping but is there a way to make on-window-detected run and move only the normal ghostty windows to a certain workspace.
I've tried a bunch of builtin if combinations from the docs site but it also seems to also move the quick terminal popup. Maybe I'm missing something.

thuvasooriya:
and on another note, this is a common suggestion and not restricted to ghostty right?
this would possibly solve the floating of arc previews and mini arc windows as well, I think Orion browser also introduced link previews recently. So I'm not sure why this was marked as completed. Sorry if I'm missing something correct me if I'm wrong please.

sfc-gh-jroes:
I'm running 0.17.1-beta and ghostty 1.1.3 and still seeing this. If I have two tabs in ghostty, I get an empty tile. Did it regress?

[...]

PR #615, opened 2024-10-25
Add a window-specific filter for ghostty's quick terminal

hackrmomo:
Discussion: #614


nikitabobko:
Closing since it was fixed on Ghostty side. Thanks for the fix!

Issue #603, opened 2024-10-18
[feature] summon-workspace: don't switch monitors


raisjn:
This feature is very cool! Aerospace is almost perfect for me now :-D 

Is it possible to add a parameter to `summon-workspace` so that when focusing to an active workspace on a separate monitor, it brings that workspace to the current monitor? This is how my xmonad is currently setup. 

Current behavior: the other monitor will become focused. If we are at [0,1*], it becomes [0*, 1]
Desired behavior: the two workspaces will be swapped. If it was [0,1*] before, it would be [1, 0*]. 

I have modified the source to accomplish this for myself, but would appreciate it being included in your source for future releases. Thank you! 

## Checklist

- [x] I've searched for duplicates. My search queries were (please don't forget to search in closed issues too):
  - xmonad
  - summon-workspace
- [x] I've read the guide https://nikitabobko.github.io/AeroSpace/guide
- [x] I've searched in the documentation (https://nikitabobko.github.io/AeroSpace/guide & https://nikitabobko.github.io/AeroSpace/commands). 

My search queries were:
  - summon-workspace

nikitabobko:
Yes, I realized that AeroSpace behavior isn't identical to Xmonad after 0.15.0 was released. I agree that swapping the workspaces should be the default.

As an Xmonad user, can you please tell how is the appropriate operation called in Xmonad?

raisjn:
> When you have multiple workspaces visible and mod-n to a different visible workspace, your current one swaps with the other one. We'll see how to change that below, if you don't like the swapping -- simply change 'greedyView' to 'view' in your workspace key bindings

from https://wiki.haskell.org/Xmonad/Frequently_asked_questions#Multi_head_and_workspaces_.28desktops.29

in short: swapping the workspaces (and keeping focus on current monitor) is called "greedyView" and is default. not swapping workspace + moving focus is known as "view"

raisjn:
i'm going to daily drive this and try to implement + document the missing features from xmonad that i use frequently. a quick list so far (please don't feel the need to implement these, just putting this here for myself for now)

1. is there a keybinding for swapping the contents of the monitor. i suspect there may be a sequence of commands to accomplish it
2. a command to focus the next and previous windows (instead of directional only) with wrap around. 
3. using `layout accordion` doesn't seem to embed the current window back into the container, but `layout tiling` does. i typically bind one key for pushing a window back into the tiling order and have separate keys for switching between layouts. 
[...]

PR #681, opened 2024-11-09
Add `--when-visible (focus|swap)` option to the summon-workspace command

lynaghk:
First, let me say AeroSpace is really cool! I've been trying to relive my XMonad memories on Mac [for years](https://kevinlynagh.com/organizing-windows/) and am excited about the possibilities with AeroSpace.

This PR implements the swapping behavior discussed in https://github.com/nikitabobko/AeroSpace/issues/603.

I could be misunderstanding the AeroSpace internals, but it doesn't seem like this feature is blocking on workspace renaming. I've tested it locally with multiple monitors and tiling / accordion layouts and it seemed to work as intended. (Note: I have not attempted to wire up any automated tests, though would be happy to do so if you'd like me to.)

Even though in that issue you said you wanted `swap` to be the default behavior, in the PR I have `focus` as the default for backwards compatibility.

Finally, I live and work in Oud-West, so if you would prefer to discuss in person --- beverages are on me =D

### DIFF
--- a/Sources/AppBundle/command/impl/SummonWorkspaceCommand.swift
+++ b/Sources/AppBundle/command/impl/SummonWorkspaceCommand.swift
@@ -7,15 +7,34 @@ struct SummonWorkspaceCommand: Command {
     func run(_ env: CmdEnv, _ io: CmdIo) -> Bool {
         check(Thread.current.isMainThread)
         let workspace = Workspace.get(byName: args.target.val.raw)
-        let monitor = focus.workspace.workspaceMonitor
-        if monitor.activeWorkspace == workspace {
+        let focusedMonitor = focus.workspace.workspaceMonitor
+
+        if focusedMonitor.activeWorkspace == workspace {
             io.err("Workspace '\(workspace.name)' is already visible on the focused monitor. Tip: use --fail-if-noop to exit with non-zero code")
             return !args.failIfNoop
         }
-        if monitor.setActiveWorkspace(workspace) {
-            return workspace.focusWorkspace()
+
+        if !workspace.isVisible {
+            // then we just need to summon the workspace to the focused monitor
+            if focusedMonitor.setActiveWorkspace(workspace) {
+                return workspace.focusWorkspace()
+            } else {
+                return io.err("Can't move workspace '\(workspace.name)' to monitor '\(focusedMonitor.name)'. workspace-to-monitor-force-assignment doesn't allow it")
+            }
         } else {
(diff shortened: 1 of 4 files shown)


lynaghk:
I've been driving this today and have noticed a bug: I have

```
on-focus-changed = ['move-mouse window-lazy-center']
on-focused-monitor-changed = ['move-mouse monitor-lazy-center']
```

enabled and when I summon a workspace the mouse jumps to the other monitor. I'll investigate and see if I can push up a fix.

raisjn:
@lynaghk: I also implemented a similar summon workspace feature - https://github.com/nikitabobko/AeroSpace/pull/606 - see the discussion for some of the nuances to keep in mind. the reference to renaming workspaces was for a separate feature idea: xmonad supports swap workspaces on monitors. 

your other PR (https://github.com/nikitabobko/AeroSpace/pull/682) is similar to another PR i implemented (https://github.com/nikitabobko/AeroSpace/pull/605), so it looks like we have been thinking similarly 😄

lynaghk:
Thanks for the pointers @raisjn! I didn't see your PR referenced from #603 so I totally missed it. I agree with you that:

> "when using workspace switching, it should never move the mouse off the current monitor".

I'll update this PR once I understand more what's happening and can resolve.

raisjn:
fwiw, the code in my PR maintains the "should never move the mouse off the current monitor" maxim and correctly works with pinned workspaces (`forceAssignedMonitor`). i'm not sure if you are using pinned workspaces, but consider the case: if you summon a pinned workspace, you don't want to move the mouse to that monitor.

also consider this case: `[U, 0*] - summon U -> [U, 0*] (nothing happens because 0 is pinned)` and what the behavior should be.

cogliostro:
what's the status on this?

nikitabobko:
Status: I'm too lazy and overloaded to take a look at the PR. And I don't know how people really use `summon-workspaces`, so I don't know what should be the default

And I hate `setActiveWorkspace` implementation (not this PR's the problem). I know that there are bugs in the implementation that manifest themselves in some special conditions, but I'm too lazy to think about them (AFAIK, right now, the conditions are such, so that the bugs don't manifest themselves). And it's all not covered with tests. The problem is that the PR naturally uses the API, because it has to (again, not this PR's problem).

And I plan to rework the whole tree structure #1215. Part of which would be to make monitors real part of the tiling tree, and cover it with tests.

So, unfortunately, the PR is not the priority. Not until, #1215 is fixed

awwaiid:
I updated #1349 to be rebased off of current `main`. This small change has continued to work fantastically for me, thank you again to @lynaghk and @nikitabobko !

picosquare:
Fantastic feature. Would like to see this merged.

Issue #248, opened 2024-05-30
Support `focus dfs-next` / `focus dfs-prev`


patrikkj:
Thank you for building a great tool!

One thing that I've missed is the ability to `focus`, `move` and `join-with` using  clockwise/counterclockwise navigation, akin to `mod1 + j`, `mod1 + k` etc. in [Amethyst](https://ianyh.com/amethyst/). Alternative names could be `next`/`previous`.

Features:
- allows for movements using one/two directional keys instead of four
- requires significantly fewer actions in certain cases
- is unambiguous, as in its behaviour does not change based on the focus cache.

Example:
```
h_tiles
├── window_1 (focused)
└── v_tiles
    ├── window_2 (prev. focused)
    └── window_3
```

navigating to `window_3` in this case requires two actions,  `right` which would take you to `window_2` followed by `down`. This could have been done using a single ccw action.

nikitabobko:
How should more complicated layouts work in this case?
```
h_tiles
  window1
  v_tiles
    window2
    window3
  window4
```

patrikkj:
Vaild point - I think an angle-based heuristic would make sense:
<img width="862" alt="image" src="https://github.com/nikitabobko/AeroSpace/assets/35503519/80543412-4d8a-4c85-b48f-e279eb744561">

I very rarely find myself having more than two "columns" in either direction though, in which case the cw/ccw navigation would be very intuitive.

nffdiogosilva:
I would also like to have this feature :)

codemac:
Would love this option as well. In many cases I have 2, maybe 3 windows.

The complex layout options is when hjkl works better, but for 90% of the time I basically want "some other" window of which is either one or two windows.

nikitabobko:
Given AeroSpace tree nature, a more well-defined algorithm would be to add possibility to focus next/prev window in DFS (Depth-first search) order

For simple layouts, it will work exactly as clockwise/counterclockwise

But it's yet unclear how to make it work with `move` and `join-with` commands (maybe it's not needed)

Though `next-dfs`/`prev-dfs` works good with upcoming `swap` command #8

ntrp:
Regarding `move` and `join-with` I also agree it is not needed, the cycle command is really useful when you simply have 2-3 windows per workspace and you want to switch focus between them without needing to think if the layout is vertical or horizontal. 
[...]

PR #682, opened 2024-11-10
Allow specifying relative DFS index for focus command

lynaghk:
I also wanted dfs next/prev as discussed in https://github.com/nikitabobko/AeroSpace/issues/248.

I figured I could implement without adding additional flags or breaking backwards compatibility by having the parser look for an optional +/- suffix on the existing dfs-index argument. The next/previous behavior can be had via:

focus --dfs-index +1
focus --dfs-index -1

respectively.

As with my other PR, I haven't implemented any automated tests around this but would be happy to look into doing so if desired.


nikitabobko:
Thanks for the idea, but I think it should be implemented as separate `dfs-next|dfs-prev` arguments as mentioned in the issue, feel free to submit a different PR

You also forgot to update the docs and shell completion

PR #684, opened 2024-11-10
Feature: added option `cross-workspace-floating-windows` to allow floating windows to exist across workspaces

nmder:
Hi there,

Thank you for this super useful tool!

I personally find this feature useful: to have the floating windows exist across the workspaces (without switching to their original workspaces), as I often have these windows bound to some keybindings.

Works fine for me thus far.

Thanks!!


nikitabobko:
Thanks, but it's not the way it should be implemented. The right way is #2. Quite extensive refactorings should be made to support it the way I want it to.

I reject the PR

nmder:
`Virtual monitor` would be a more robust solution. I agree.

Issue #56, opened 2023-12-15
Monospaced font for menu bar?


MatthiasGrandl:
So I really love the subtle menubar that shows the current space. However the default font doesn't lend itself very well to this purpose as the different numbers/letters, have different widths, causing other menubar items to jump when switching spaces. The use of a monospaced font would alleviate this. 

For example the MacOS default clock in the right of the menu bar seems to be using the equivalent of CSS: `font-variant-numeric: tabular-nums;` to prevent jumping.

nikitabobko:
I'm 100% agree that it needs to be monospaced. The menu bar is programmed in Swift UI and [I tried](https://github.com/nikitabobko/AeroSpace/blob/main/src/AeroSpaceApp.swift#L78) to make it monospaced but it didn't work.

I suspect that macOS simply doesn't allow changing the font of the tray items. Are there any known apps that show their tray item in monospaced? macOS clock doesn't count. It's a system thing, it can have more privileges

MatthiasGrandl:
So could it be worked around with icons? I am not at all familiar with SwiftUI MacOS programming. 

But yeah I expected this to not be super straightforward :)

nikitabobko:
> So could it be worked around with icons?

The tray icon shows workspace names. Workspace names are not necessarily single-lettered

rami3l:
When I was using Amethyst, I also used a companion app called SpaceId, and the latter seems to be using Swift code to generate the tray icon: https://github.com/dshnkao/SpaceId/blob/b6d452c818c37403d471de32d7a9044e88a06b14/SpaceId/Application/ButtonImage.swift

It works pretty well actually: 
![](https://raw.githubusercontent.com/dshnkao/SpaceId/master/images/i.png)

ELLIOTTCABLE:
SpaceId looks _fantastic_, and of course I find it via this thread, right as it becomes useless to me since I'm switching to Aerospace.

:+1: to adding a _list_ of spaces to the titlebar, that doesn't shift around often.

(My vote would be: show all spaces that have ≥1 window; with the current one highlighted dynamically switch around between them.)

Yixuan-Wang:
I tried copying SpaceId's source code in and here is how it looks
<img width="330" alt="SpaceId style tray icon" src="https://github.com/user-attachments/assets/9f1d145b-3b93-48bf-9af1-7b95ce4b8aa4">

> > So could it be worked around with icons?
> 
[...]

PR #685, opened 2024-11-10
Monospace menubar

frostplexx:
This PR addresses issue #56 by introducing three new options:
```toml
# Font size for menu bar
font-size = 12.0

# Font family for the menu bar items
# Use the Font Book to find fonts
font-family = 'JetBrainsMono Nerd Font Mono'

# Possible values: icon|horizontal_full|horizontal_min
# icon: shows only the current workspace
# horizontal_full: shows all workspaces, with the current one highlighted
# horizontal_min: shows all workspaces with open windows,and the current one
workspace-indicator-style = 'icon'
```
To accomplish this, I had to rewrite all menu bar items to use AppKit, as pure SwiftUI lacks the capability to display arbitrary views.

Here are some pictures demonstrating the feature in action:
image
image
image
image


cvlvxi:
Can we get a video demo? This seems pretty interesting

frostplexx:
https://github.com/user-attachments/assets/4041861f-58e7-4ae1-9c6e-9d03ecac1150

Emojis are pushed to the end, which is the same behavior that exists.

Sources/AeroSpaceApp/AppDelegate.swift
```diff
@@ -0,0 +1,237 @@
+import AppBundle
+import AppKit
+import Cocoa
+import Combine
+import Common
+import Foundation
+import SwiftUI
+
+class AppDelegate: NSObject, NSApplicationDelegate {
```
nikitabobko, line 9: `Sources/AeroSpaceApp` should stay as small as possible, because this "module" is shared between Xcode and SPM. All the code must be pushed to `Sources/AppBundle` module

Sources/AeroSpaceApp/AppDelegate.swift
```diff
@@ -0,0 +1,237 @@
...
+
+        // Add reload config item if enabled
+        if viewModel.isEnabled {
+            let reloadConfigItem = NSMenuItem(
+                title: "Reload config", action: #selector(reloadConfigAction(_:)), keyEquivalent: "r")
+            reloadConfigItem.target = self
+            menu.addItem(reloadConfigItem)
+        }
+
+        // Add quit item
```
nikitabobko, line 157: Please avoid obvious comments

docs/config-examples/default-config.toml
```diff
@@ -43,6 +43,19 @@ on-focused-monitor-changed = ['move-mouse monitor-lazy-center']
 # Also see: https://nikitabobko.github.io/AeroSpace/goodness#disable-hide-app
 automatically-unhide-macos-hidden-apps = false
 
+# Font size for menu bar
+font-size = 12.0
+
+# Font family for the menu bar items
+# Use the Font Book to find fonts
+font-family = 'Andale Mono'
```
nikitabobko, line 51: I don't think that font should be configurable. We should just pick a good default and stick with it.

nikitabobko (commented):
Thanks for your work, but I am not ready to accept such big PR. The PR is hardly reviewerable.

Please, fix problem by problem in small incremental accurate commits. The good first step is to only make the indicator monospaced and nothing else. Everything else can be discussed after that.

nikitabobko:
Also see: https://github.com/nikitabobko/AeroSpace/issues/711#issuecomment-2541128213

PR #851, opened 2024-12-12
Make menu font monospaced using ImageRenderer

mobile-ar:
Here is quite a small change to enable the menu to display the monospaced font. The trick is achieved using `ImageRenderer` that will create an image from a content view.

### DIFF
--- a/Sources/AppBundle/MenuBar.swift
+++ b/Sources/AppBundle/MenuBar.swift
@@ -51,8 +51,35 @@ public func menuBar(viewModel: TrayMenuModel) -> some Scene {
             terminateApp()
         }.keyboardShortcut("Q", modifiers: .command)
     } label: {
-        // .font(.system(.body, design: .monospaced)) doesn't work unfortunately :(
-        Text(viewModel.isEnabled ? viewModel.trayText : "⏸️")
+        if viewModel.isEnabled {
+            menuLabel(viewModel: viewModel)
+                .id(viewModel.trayText.hashValue)
+        } else {
+            Text("⏸️")
+        }
+    }
+}
+
+struct menuLabel: View {
+    var viewModel: TrayMenuModel
+
+    var body: some View {
+        let renderer = ImageRenderer(content: imageContent)
+        if let cgImage = renderer.cgImage {
+            Image(cgImage, scale: 2, label: Text(viewModel.trayText))
+                .renderingMode(.template)
+                .resizable()
+                .aspectRatio(contentMode: .fit)
+        } else {
+            // In case image can't be rendered fallback to plain text
+            Text(viewModel.trayText)
+        }
+    }
+
+    // I used a largeTitle font and then use a scale of 2 to make the image look smoother
+    private var imageContent: some View {
+        Text(viewModel.trayText)
+            .font(.system(.largeTitle, design: .monospaced))
     }
 }
 
--- a/Sources/AppBundle/MenuBar.swift
+++ b/Sources/AppBundle/MenuBar.swift
@@ -61,13 +61,14 @@ public func menuBar(viewModel: TrayMenuModel) -> some Scene {
 }
 
 struct menuLabel: View {
+    @Environment(\.colorScheme) var colorScheme
+
     var viewModel: TrayMenuModel
 
     var body: some View {
         let renderer = ImageRenderer(content: imageContent)
         if let cgImage = renderer.cgImage {
             Image(cgImage, scale: 2, label: Text(viewModel.trayText))
-                .renderingMode(.template)
                 .resizable()
                 .aspectRatio(contentMode: .fit)
         } else {
@@ -80,6 +81,7 @@ struct menuLabel: View {
     private var imageContent: some View {
         Text(viewModel.trayText)
             .font(.system(.largeTitle, design: .monospaced))
+            .foregroundStyle(colorScheme == .light ? Color.black : Color.white)
     }
 }
 


Sources/AppBundle/MenuBar.swift
```diff
@@ -51,8 +51,35 @@ public func menuBar(viewModel: TrayMenuModel) -> some Scene {
...
+            // In case image can't be rendered fallback to plain text
+            Text(viewModel.trayText)
+        }
+    }
+
+    // I used a largeTitle font and then use a scale of 2 to make the image look smoother
+    private var imageContent: some View {
+        Text(viewModel.trayText)
+            .font(.system(.largeTitle, design: .monospaced))
+            .foregroundStyle(Color.white)
```
nikitabobko, line 82: I neither checked out, nor tested the code. Doesn't this line make the text constantly white? What if the user has light wallpaper?

mobile-ar: True that. I tested light and dark appearence with the same wallpaper(my dark wallpaper) and both displayed same color, but my bad assuming they where the same, appearance applies to the apps windows not to the menu bar of macOS.
I can push a simple fix, removing the color of the text and adding: `.renderingMode(.template)` to the image in that way the image will be considered a template and it will apply the tint color same as the Os displays. I can fix this commit, just tested it:

image
image

mobile-ar: Fixed. Also was thinking that in the future making the font `.bold` will make it more visible. Also I was thinking in extending this using this ImageRenderer functionality to implement something like this proposal #844

nikitabobko:
To verify: do emoji-named workspaces still work?

mobile-ar:
> To verify: do emoji-named workspaces still work?

I wasn't aware that emojis were supported on workspace name, not is not working, but I have a fixed this. I will update this PR soon.

After fixing this issue I did some profiling of the app regarding memory usage and I find out that this implementation the image rendering is using and quickly releasing around 100 Mb of memory (like you can see in the image below).

image

The unmodified app on debug mode uses around 50 Mb. This new implementation uses the same ~50 Mb + the peaks of extra ~100 Mb for rendering the image.

So please take in consideration this extra memory consumption in your review.

Sources/AppBundle/MenuBar.swift
```diff
@@ -51,8 +51,37 @@ public func menuBar(viewModel: TrayMenuModel) -> some Scene {
             terminateApp()
         }.keyboardShortcut("Q", modifiers: .command)
     } label: {
-        // .font(.system(.body, design: .monospaced)) doesn't work unfortunately :(
-        Text(viewModel.isEnabled ? viewModel.trayText : "⏸️")
+        if viewModel.isEnabled {
+            menuLabel(viewModel: viewModel)
+                .id(viewModel.trayText.hashValue)
```
nikitabobko, line 56: What is this `id` for? I think it for some optimization purposes, like to avoid rendering twice, right?

But if I drop `.id` and put `let _ = print("render")` inside `var body: some View`, I don't see any unnecessary prints in stdout

mobile-ar: Yes, to avoid rendering twice and also to force it to render when the value changes on the viewModel. Some time ago I faced some SwiftUI issues not displaying correct values, because the view did not refresh after value changed.

nikitabobko: But if I drop `.id` and put `let _ = print("render")` inside `var body: some View`, I don't see any unnecessary prints in stdout

nikitabobko: Does `.id` have any effects in our case?

mobile-ar: Yes, if you remove the `.id` you will not see the correct workspaces changes reflected in the image. (As I mentioned "... mostly to force it to render when the value changes on the viewModel")

Given that the full display is now an Image, if something changes in top level, SwiftUI is not triggering the render of it internally because the top level view is not observing the `trayText` property change. It's only doing `menuLabel(viewModel: viewModel)` and might not get aware of the internal changes of the properties.

You can try this by running it without `.id` and you will notice that if you start in workspace 1 and you move to 2 it won't be displayed correctly in the menu bar (menu bar will still display 1).

nikitabobko: Ok, I missed that thank you. I'm going to replace `var viewModel: TrayMenuModel` to `var text: String` in that case `.id` is not necessary

nikitabobko:
Thanks, merged with the following patch on top:
```diff
diff --git a/Sources/AppBundle/MenuBar.swift b/Sources/AppBundle/MenuBar.swift
index e15eb2bb..479749f0 100644
--- a/Sources/AppBundle/MenuBar.swift
+++ b/Sources/AppBundle/MenuBar.swift
@@ -52,37 +52,32 @@ public func menuBar(viewModel: TrayMenuModel) -> some Scene {
         }.keyboardShortcut("Q", modifiers: .command)
     } label: {
         if viewModel.isEnabled {
-            menuLabel(viewModel: viewModel)
-                .id(viewModel.trayText.hashValue)
+            MonospacedText(viewModel.trayText)
         } else {
-            Text("⏸️")
+            MonospacedText("⏸️")
         }
     }
 }
 
-struct menuLabel: View {
-    @Environment(\.colorScheme) var colorScheme
-
-    var viewModel: TrayMenuModel
+struct MonospacedText: View {
+    @Environment(\.colorScheme) var colorScheme: ColorScheme
+    var text: String
+    init(_ text: String) { self.text = text }
 
     var body: some View {
-        let renderer = ImageRenderer(content: imageContent)
+        let renderer = ImageRenderer(
+            content: Text(text)
+                .font(.system(.largeTitle, design: .monospaced))
+                .foregroundStyle(colorScheme == .light ? Color.black : Color.white)
+        )
         if let cgImage = renderer.cgImage {
-            Image(cgImage, scale: 2, label: Text(viewModel.trayText))
-                .resizable()
-                .aspectRatio(contentMode: .fit)
+            // Using scale: 1 results in a blurry image for unknown reasons
+            Image(cgImage, scale: 2, label: Text(text))
         } else {
             // In case image can't be rendered fallback to plain text
-            Text(viewModel.trayText)
+            Text(text)
         }
     }
-
-    // I used a largeTitle font and then use a scale of 2 to make the image look smoother
-    private var imageContent: some View {
-        Text(viewModel.trayText)
-            .font(.system(.largeTitle, design: .monospaced))
-            .foregroundStyle(colorScheme == .light ? Color.black : Color.white)
-    }
 }
 
 func getTextEditorToOpenConfig() -> URL {

```

Issue #752, opened 2024-11-24
Softly assign workspace to monitor


Amparose:
Workspaces with intent is great, i.e. "Workspace (C)omms for email and messaging apps". I _usually_ like this one on Monitor D but sometimes I want to focus it by moving it to Monitor A. To have this flexibility I can't use `workspace-to-monitor-force-assignment` but then I have to manually move all my workspaces into position on every start, which is annoying.

I'd like a `workspace-to-monitor-assignment` (no `force`) that assigns a workspace to a monitor, if it exists, at launch but doesn't force it to stay there. A "reset this workspace to assigned monitor" would be nice to avoid the workspace shuffle of moving one monitor-at-a-time.

nikitabobko:
Duplicate of #123

PR #853, opened 2024-12-12
Add soft default assigned monitor

lordalisat:
Added the possibility to add a default assigned monitor to a workspace, as suggested in https://github.com/nikitabobko/AeroSpace/issues/752#issue-2687386754.
Default assigned monitor works similarly to force assigned monitor, though the move-workspace-to-monitor command will work in this case. The move-workspace-to-monitor reset command can be used to reset to the default monitor (defaults to main). Force assigned monitor will always have priority over default assigned monitor.


nikitabobko:
Close due to inactivity

PR #863, opened 2024-12-16
doc: Add Shottr app-id

lukeramljak:
Added Shottr to the apps list


nikitabobko:
Thanks, but I don't want to pull it into the repo

PR #864, opened 2024-12-16
feat: add `--boundaries` options to move command

### LINKED ISSUE #183: [Feature Request] Return status of an aerospace command after executing it
# Motivation
I'm currently trying to do something like:

>  If you can move the current window to the left in the current workspace, do it! Otherwise, move the window to the monitor in the left (wrapping around)

So I thought that the following simple zsh script would do it for me:

```zsh
aerospace move left || aerospace move-node-to-monitor --wrap-around left
```

Unfortunately, that doesn't work, since `aerospace move left` does not fail when it can't move. 

# Feature request
I think it would be nice for aerospace commands to return something to indicate if they "failed" to do what they were supposed to do. It would be enough to solve my problem.

# Alternative
Another way to solve it would be to make `move` to work similarly to the [focus command](https://nikitabobko.github.io/AeroSpace/commands#focus) (with boundaries and boundaries actions).

EcmaXp:
## Summary
- Resolves #183

This PR introduces a new `--boundaries` option to the move command, allowing windows to move beyond their current workspace.

By default, window movements remain within a single workspace. With `--boundaries all-monitors-outer-frame`, windows can now be moved across multiple monitors for increased flexibility.

## Key Changes

`move` Command Updates:

```diff
- aerospace move [-h|--help] [--window-id <window-id>] (left|down|up|right)
+ aerospace move [-h|--help] [--window-id <window-id>] [--focus-follows-window]
+                [--fail-if-noop] [--boundaries <boundary>] [--wrap-around] (left|down|up|right)
```

- `--boundaries <boundary>`: Choose between staying within the `workspace` or `all-monitors-outer-frame`.
- `--wrap-around`: Cycle through monitors when reaching an edge.
- `--focus-follows-window:` Automatically focus the moved window.
- `--fail-if-noop`: Exit with a non-zero status if the window cannot move (e.g., hitting the boundary without `--boundaries all-monitors-outer-frame`).

## Why?

These enhancements provide more control over window movement in multi-monitor setups and improve scripting scenarios by offering fail-fast behavior, wrap-around cycling, and automatic refocusing.

## Example

```toml
# aerospace.toml
ctrl-alt-left = "move left --boundaries all-monitors-outer-frame --focus-follows-window"
ctrl-alt-down = "move down --boundaries all-monitors-outer-frame --focus-follows-window"
ctrl-alt-up = "move up --boundaries all-monitors-outer-frame --focus-follows-window"
ctrl-alt-right = "move right --boundaries all-monitors-outer-frame --focus-follows-window"
```

### DIFF
--- a/Sources/AppBundle/command/impl/MoveCommand.swift
+++ b/Sources/AppBundle/command/impl/MoveCommand.swift
@@ -56,7 +56,6 @@ private func moveOut(_ io: CmdIo, window: Window, direction: CardinalDirection) 
                  .macosHiddenAppsWindowsContainer, .macosPopupWindowsContainer: true
         }
     }) as! TilingContainer
-    let lastAppliedLayoutVirtualRect: Rect? = window.lastAppliedLayoutVirtualRect
     let bindTo: TilingContainer
     let bindToIndex: Int
     switch innerMostChild.parent.nodeCases {
@@ -65,6 +64,10 @@ private func moveOut(_ io: CmdIo, window: Window, direction: CardinalDirection) 
             bindTo = parent
             bindToIndex = innerMostChild.ownIndex + direction.insertionOffset
         case .workspace(let parent): // create implicit container
+            if !canMoveOutInDirection(window: window, direction: direction) {
+                return false;
+            }
+
             let prevRoot = parent.rootTilingContainer
             prevRoot.unbindFromParent()
             // Force tiles layout
@@ -87,7 +90,24 @@ private func moveOut(_ io: CmdIo, window: Window, direction: CardinalDirection) 
         adaptiveWeight: WEIGHT_AUTO,
         index: bindToIndex
     )
-    return window.lastAppliedLayoutVirtualRect != lastAppliedLayoutVirtualRect;
+    return true;
+}
+
+private func canMoveOutInDirection(window: Window, direction: CardinalDirection) -> Bool {
+    let rootTilingContainer = window.nodeWorkspace?.rootTilingContainer
+    if window.parent != rootTilingContainer {
+        return true
+    }
+    if rootTilingContainer?.orientation != direction.orientation {
+        return true
+    }
+
+    switch direction {
+        case .left, .up:
+            return window.ownIndex != 0
+        case .right, .down:
+            return window.ownIndex != window.parent.children.count - 1
+    }
 }
 
 private func deepMoveIn(window: Window, into container: TilingContainer, moveDirection: CardinalDirection) {
--- a/Sources/AppBundle/command/impl/MoveCommand.swift
+++ b/Sources/AppBundle/command/impl/MoveCommand.swift
@@ -26,13 +26,13 @@ struct MoveCommand: Command {
                     return true
                 } else {
                     if moveOut(io, window: currentWindow, direction: direction) {
-                        return true;
+                        return true
                     } else if args.boundaries == .allMonitorsUnionFrame {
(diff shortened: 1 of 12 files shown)


dvrkoo:
Currently switching between linux with Hyprland and macOS with Aerospace a lot and this is one of features I was missing the most, thanks a lot

Sources/AppBundle/command/impl/MoveCommand.swift
```diff
@@ -56,6 +64,10 @@ private func moveOut(_ io: CmdIo, window: Window, direction: CardinalDirection)
             bindTo = parent
             bindToIndex = innerMostChild.ownIndex + direction.insertionOffset
         case .workspace(let parent): // create implicit container
+            if !canMoveOutInDirection(window: window, direction: direction) {
```
nikitabobko, line 67: From what I read in the code, you disallow creating implicit containers when the window is moved along the axis of the rootTilingNode. This is wrong approach. It should be conditional behavior of `--boundaries-action`

By default it should always be possible to move out window out of container. Please, read #183 thread

Sources/Common/cmdArgs/impl/MoveCmdArgs.swift
```diff
@@ -6,21 +8,51 @@ public struct MoveCmdArgs: CmdArgs {
...
     public var direction: Lateinit<CardinalDirection> = .uninitialized
-    public var windowId: UInt32?
+    public var windowId: UInt32? { // Forward to moveNodeToMonitor
+        get { moveNodeToMonitor.windowId }
+        set(newValue) { moveNodeToMonitor.windowId = newValue }
+    }
     public var workspaceName: WorkspaceName?
+    public var failIfNoop: Bool = false
+    public var rawBoundaries: Boundaries? = nil
+    public var moveNodeToMonitor = MoveNodeToMonitorCmdArgs(rawArgs: [])
```
nikitabobko, line 31: I didn't think a lot about it, but it may not be the right place to reuse `move-node-to-monitor`

nikitabobko (commented):
Thanks for the effort, but I reject the PR. Your understanding of the mental model of moving windows is incorrect. The PR is fundamentally wrong.

EcmaXp:
> Thanks for the effort, but I reject the PR. Your understanding of the mental model of moving windows is incorrect. The PR is fundamentally wrong.

Thanks for taking the time to review my PR. I appreciate your feedback and insights.

I’ll revisit the approach based on your comments and open a new PR that aligns more closely with the intended mental model.

Sources/Common/cmdArgs/impl/MoveCmdArgs.swift
EcmaXp: I don’t fully agree. `MoveNodeToMonitorCommand` already calls `MoveNodeToWorkspaceCommand` in a similar way. There might be other approaches, but I’m not sure yet.

https://github.com/nikitabobko/AeroSpace/blob/2891df5e316ececb3f4f57499616cad64f3b9c76/Sources/AppBundle/command/impl/MoveNodeToMonitorCommand.swift#L19-L20

docs/aerospace-move.adoc
```diff
@@ -25,6 +26,17 @@ Deprecated name: `move-through`
 include::./util/conditional-options-header.adoc[]
 
 -h, --help:: Print help
+--fail-if-noop:: Exit with a non-zero status if the window cannot move because it’s at the boundary
```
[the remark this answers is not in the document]
EcmaXp: I’ve reviewed #183 and your feedback.

I’m a bit unsure about how to best design the arguments, but I’ll reconsider the approach.

Issue #999, opened 2025-01-18
Introduce `force-keep-workspaces-alive` config option


nikitabobko:
```toml
force-keep-workspaces-alive = ['A', 'B', 'C', 'D']
```

The mental model is described here: https://github.com/nikitabobko/AeroSpace/discussions/875#discussioncomment-11618465

The workspaces in the menubar should be sorted the same as workspaces in the `force-keep-workspaces-alive` array (to cover #830)

To make all workspaces "phantom" (https://github.com/nikitabobko/AeroSpace/discussions/996), one can use the empty array:
```toml
force-keep-workspaces-alive = []
```

nikitabobko:
UPD: renamed to `persistent-workspaces`

nikitabobko:
Instead of `auto-infer-from-config` https://github.com/nikitabobko/AeroSpace/discussions/875#discussioncomment-11618465, I am eager to make a breaking change here and force users to explicitly write `persistent-workspaces` in their config (whether it's empty array or some values)

dezren39:
hey how is this going? :-) i saw the i3-style status bar and it reminded me of this. thank you again!

greggh:
I am also wondering how this is going. I was just searching for a way to keep my workspaces alive. I don't care if they are empty or not. Thanks for all the awesome work.

nikitabobko:
Implemented in 0.20.0-Beta.

`persistent-workspaces` config option is only available if you specify `config-version = 2` in your config. See the docs & default config for example

McHeisenburglar:
@nikitabobko With this change, is it possible to *create* workspaces on-the-fly without explicit names? I'm coming from a macOS Spaces workflow, so I create and delete workspaces as I see fit. I have enabled phantom workspaces in my AeroSpace, so now I don't have any empty/redundant spaces, but I would like to somehow do "move window to new workspace" (without having to explicitly say "move to workspace 1", "move to workspace 2").

lutzseverino:
I'm very new to AeroSpace and I want to understand: Is this the correct way of "hiding" the workspaces that I'm not using? E.g. I only use numbered workspaces, I don't like having the a-z clutter. I set the `persistent-workspaces` to only the numbered workspaces and the result seems what I intended. 

Regardless, I wanna make sure that's the intended usage, maybe help others also understand this config value.

Issue #711, opened 2024-11-15
taskbar icon for each active workspace


dezren39:
i'm not sure if mac allows this,

i would like each workspace with active windows to be displayed as an icon.

if i have windows in workspaces 1,2,6,9
then the task bar will have
- 4 icons
- one for each workspace
- active workspace is underlined
- left click on non-active will open the workspace on-click
- left click on active workspace will behave like current taskbar icon

cynicaljoy:
I'm solving this with [SketchyBar](https://github.com/FelixKratz/SketchyBar)

of course some people will want this feature w/the standard macOS menu bar, just making sure people know there is an option available right now.

mickey-doyle:
I am 99% sure this would be possible because [Spaceman](https://github.com/Jaysce/Spaceman) already exists for native macOS desktop spaces and works exactly how you describe. I'd love this feature as well, would be super awesome if this happened - or at least a little more customization on the indicator we have now. I use a 1440p monitor so the small white number is a bit annoying to visually seek out - would love color & size options one day.

nikitabobko:
A note: I'm a big UI tinkering hater. I don't care how things look, all I care is how things work - what workflows are provided is the only important thing, UI doesn't matter. And I want AeroSpace users to accept this mantra, I don't want people to waste their time on "ricing" their computer (though they already do that).

So if someone decides to push the initiative with different menubar styles forward, there a lot of things you will have to explain:
1. You have to describe a clear, organized and strict mental model of all possible menubar styles
2. You have to describe pros and cons of each menubar style
3. You have to describe use-cases for each menubar style (like for example, it's unclear to me why the indication of the currently active workspace is not enough)
4. You have to take into consideration the fact that MacBook notch doesn't allow the indicator being too wide
5. You have to optimize number of possible menubar styles (make it as small as possible)
6. For every knob and configuration option you provide, you have to explain the importance in terms of workflows and use-cases
7. etc...

Just as an example: on my Linux machine, I could have different dpi for different windows, so that my mouse changes size when I move it from one window to another, and I couldn't care less. It's an extreme example, and I obviously agree that dpi should be the same across windows, nevertheless, it shows how much I don't care about UIs, and I will pushback hard for every knob and additional menubar style, since I'm the one who would need to support the UI configuration madness in the end.

Another example: gaps are useless, and they waste screen estate, period. There are two reasons why AeroSpace provides gaps:
1. It's too much popular request
[...]

PR #1049, opened 2025-01-29
New menu bar option for i3-like functionality #844

mobile-ar:
New menu-bar-style configuration to add extra menu functionality ('text' monospaced default, 'squares' images indicator and 'i3' squares + not active workspaces with windows in them)

Feel free to suggest better naming for the options. I will update all needed suggestions/recomendations

Full details of this implementation in #844

### DIFF
--- a/Sources/AppBundle/ExperimentalUISettings.swift
+++ b/Sources/AppBundle/ExperimentalUISettings.swift
@@ -1,0 +1,68 @@
+import AppKit
+import SwiftUI
+
+struct ExperimentalUISettings {
+    var displayStyle: MenuBarStyle {
+        get {
+            if let value = UserDefaults.standard.string(forKey: ExperimentalUISettingsItems.displayStyle.rawValue) {
+                return MenuBarStyle(rawValue: value) ?? .monospacedText
+            } else {
+                return .monospacedText
+            }
+        }
+        set {
+            UserDefaults.standard.setValue(newValue.rawValue, forKey: ExperimentalUISettingsItems.displayStyle.rawValue)
+            UserDefaults.standard.synchronize()
+        }
+    }
+}
+
+enum ExperimentalUISettingsItems: String {
+    case displayStyle
+}
+
+@MainActor
+func getExperimentalUISettingsMenu(viewModel: TrayMenuModel) -> some View {
+    let appearence = NSApplication.shared.effectiveAppearance.name
+    let color = (appearence == .vibrantDark || appearence == .darkAqua) ? Color.white : Color.black
+    return Menu {
+        Text("These settings are EXPERIMENTAL.")
+        Text("No stability guarantees are provided.")
+        Divider()
+        Text("Menu bar style (macOS 14 or later):")
+        Text(MenuBarStyle.systemText.title)
+        Button {
+            viewModel.experimentalUISettings.displayStyle = .systemText
+        } label: {
+            Toggle(isOn: .constant(viewModel.experimentalUISettings.displayStyle == .systemText)) {
+                MenuBarLabel(viewModel.trayText, textStyle: .system, color: color)
+            }
+        }
+        Text(MenuBarStyle.monospacedText.title)
+        Button {
+            viewModel.experimentalUISettings.displayStyle = .monospacedText
+        } label: {
+            Toggle(isOn: .constant(viewModel.experimentalUISettings.displayStyle == .monospacedText)) {
+                MenuBarLabel(viewModel.trayText, color: color)
+            }
+        }
+        Text(MenuBarStyle.squares.title)
+        Button {
+            viewModel.experimentalUISettings.displayStyle = .squares
+        } label: {
+            Toggle(isOn: .constant(viewModel.experimentalUISettings.displayStyle == .squares)) {
+                MenuBarLabel(viewModel.trayText, color: color, trayItems: viewModel.trayItems)
+            }
+        }
+        Text(MenuBarStyle.i3.title)
+        Button {
+            viewModel.experimentalUISettings.displayStyle = .i3
+        } label: {
+            Toggle(isOn: .constant(viewModel.experimentalUISettings.displayStyle == .i3)) {
+                MenuBarLabel(viewModel.trayText, color: color, trayItems: viewModel.trayItems, workspaces: viewModel.workspaces)
+            }
+        }
+    } label: {
+        Text("Experimental UI Settings")
+    }
+}
--- a/Sources/AppBundle/MenuBar.swift
+++ b/Sources/AppBundle/MenuBar.swift
@@ -26,6 +26,8 @@ public func menuBar(viewModel: TrayMenuModel) -> some Scene { // todo should it 
             }
             Divider()
         }
+        getExperimentalUISettingsMenu(viewModel: viewModel)
+        Divider()
         Button(viewModel.isEnabled ? "Disable" : "Enable") {
             Task {
                 try await runSession(.menuBarButton, .forceRun) { () throws in
@@ -62,25 +64,51 @@ public func menuBar(viewModel: TrayMenuModel) -> some Scene { // todo should it 
         }.keyboardShortcut("Q", modifiers: .command)
     } label: {
         if viewModel.isEnabled {
-            MonospacedText(viewModel.trayText)
+            switch viewModel.experimentalUISettings.displayStyle {
+                case .systemText:
+                    MenuBarLabel(viewModel.trayText, textStyle: .system)
+                case .monospacedText:
+                    MenuBarLabel(viewModel.trayText)
+                case .squares:
+                    MenuBarLabel(viewModel.trayText, trayItems: viewModel.trayItems)
+                case .i3:
+                    MenuBarLabel(viewModel.trayText, trayItems: viewModel.trayItems, workspaces: viewModel.workspaces)
+            }
         } else {
-            MonospacedText("⏸️")
+            MenuBarLabel("⏸️")
         }
     }
 }
 
-struct MonospacedText: View {
+@MainActor
+struct MenuBarLabel: View {
     @Environment(\.colorScheme) var colorScheme: ColorScheme
     var text: String
-    init(_ text: String) { self.text = text }
+    var textStyle: MenuBarTextStyle
+    var color: Color?
+    var trayItems: [TrayItem]?
+    var workspaces: [WorkspaceViewModel]?
+
+    let hStackSpacing = CGFloat(4)
+    let itemHeight = CGFloat(40)
+    let itemBorderSize = CGFloat(4)
+    let itemCornerRadius = CGFloat(6)
+
+    var finalColor: Color {
+        return color ?? (colorScheme == .dark ? Color.white : Color.black)
+    }
+
+    init(_ text: String, textStyle: MenuBarTextStyle = .monospaced, color: Color? = nil, trayItems: [TrayItem]? = nil, workspaces: [WorkspaceViewModel]? = nil) {
+        self.text = text
+        self.textStyle = textStyle
+        self.color = color
+        self.trayItems = trayItems
+        self.workspaces = workspaces
+    }
 
     var body: some View {
         if #available(macOS 14, *) { // https://github.com/nikitabobko/AeroSpace/issues/1122
-            let renderer = ImageRenderer(
-                content: Text(text)
-                    .font(.system(.largeTitle, design: .monospaced))
-                    .foregroundStyle(colorScheme == .light ? Color.black : Color.white)
-            )
+            let renderer = ImageRenderer(content: menuBarContent)
             if let cgImage = renderer.cgImage {
                 // Using scale: 1 results in a blurry image for unknown reasons
                 Image(cgImage, scale: 2, label: Text(text))
@@ -91,6 +119,129 @@ struct MonospacedText: View {
         } else { // macOS 13 and lower
             Text(text)
         }
+    }
+
+    var menuBarContent: some View {
+        return ZStack {
+            if let trayItems {
+                HStack(spacing: hStackSpacing) {
+                    ForEach(trayItems, id: \.id) { item in
+                        itemView(for: item)
+                        if item.type == .mode {
+                            Text(":")
+                                .font(.system(.largeTitle, design: textStyle.design))
+                                .foregroundStyle(finalColor)
+                                .bold()
+                        }
+                    }
+                    if workspaces != nil {
+                        let otherWorkspaces = Workspace.all.filter { workspace in
+                            !workspace.isEffectivelyEmpty && !trayItems.contains(where: { item in item.type == .monitor && item.name == workspace.name })
+                        }
+                        if !otherWorkspaces.isEmpty {
+                            Group {
+                                Text("|")
+                                    .font(.system(.largeTitle, design: textStyle.design))
+                                    .foregroundStyle(finalColor)
+                                    .bold()
+                                    .padding(.bottom, 2)
+                                ForEach(otherWorkspaces, id: \.name) { item in
+                                    itemView(for: TrayItem(type: .monitor, name: item.name, isActive: false))
+                                }
+                            }
+                            .opacity(0.6)
+                        }
+                    }
+                }
+                .frame(height: itemHeight)
+            } else {
+                HStack(spacing: hStackSpacing) {
+                    Text(text)
+                        .font(.system(.largeTitle, design: textStyle.design))
+                        .foregroundStyle(finalColor)
+                }
+            }
+        }
+    }
+
+    @ViewBuilder
+    fileprivate func itemView(for item: TrayItem) -> some View {
+        if item.name.containsEmoji() {
+            // If workspace name contains emojis we use the plain emoji in text to avoid visibility issues scaling the emoji to fit the squares
+            Text(item.name)
+                .font(.system(.largeTitle, design: textStyle.design))
+                .foregroundStyle(finalColor)
+        } else {
+            if let imageName = item.systemImageName {
+                Image(systemName: imageName)
+                    .resizable()
+                    .aspectRatio(contentMode: .fit)
+                    .symbolRenderingMode(.monochrome)
+                    .foregroundStyle(finalColor)
+            } else {
+                let text = Text(item.name)
+                    .font(.system(.largeTitle, design: textStyle.design))
+                    .bold()
+                    .padding(.horizontal, item.isActive ? itemBorderSize * 2 : itemBorderSize * 1.5)
+                    .frame(height: itemHeight)
+                if item.isActive {
+                    ZStack {
+                        text.foregroundStyle(.clear)
+                            .overlay(
+                                RoundedRectangle(cornerRadius: itemCornerRadius, style: .circular)
+                            )
+                        text.blendMode(.destinationOut)
+                    }
+                    .compositingGroup()
+                    .foregroundStyle(finalColor)
+                } else {
+                    text
+                        .padding(.horizontal, itemBorderSize)
+                        .overlay(
+                            RoundedRectangle(cornerRadius: itemCornerRadius, style: .circular)
+                                .strokeBorder(lineWidth: itemBorderSize)
+                        )
+                        .foregroundStyle(finalColor)
+                }
+            }
+        }
+    }
+}
+
+enum MenuBarTextStyle: String {
+    case monospaced
+    case system
+    var design: Font.Design {
+        switch self {
+            case .monospaced:
+                return .monospaced
+            case .system:
+                return .default
+        }
+    }
+}
+
+enum MenuBarStyle: String, CaseIterable, Identifiable, Equatable, Hashable {
+    case monospacedText
+    case systemText
+    case squares
+    case i3
+    var id: Int {
+        return self.hashValue
+    }
+    var title: String {
+        switch self {
+            case .monospacedText: "Monospaced font"
+            case .systemText: "System font"
+            case .squares: "Square images"
+            case .i3: "i3 style"
+        }
+    }
+}
+
+private extension String {
+    func containsEmoji() -> Bool {
+        unicodeScalars.contains { $0.properties.isEmoji && $0.properties.isEmojiPresentation }
     }
 }
 
--- a/Sources/AppBundle/TrayMenuModel.swift
+++ b/Sources/AppBundle/TrayMenuModel.swift
@@ -7,9 +7,11 @@ public class TrayMenuModel: ObservableObject {
     private init() {}
 
     @Published var trayText: String = ""
+    @Published var trayItems: [TrayItem] = []
     /// Is "layouting" enabled
     @Published var isEnabled: Bool = true
     @Published var workspaces: [WorkspaceViewModel] = []
+    @Published var experimentalUISettings: ExperimentalUISettings = ExperimentalUISettings()
 }
 
 @MainActor func updateTrayText() {
@@ -25,10 +27,51 @@ public class TrayMenuModel: ObservableObject {
         let monitor = $0.isVisible || !$0.isEffectivelyEmpty ? " - \($0.workspaceMonitor.name)" : ""
         return WorkspaceViewModel(name: $0.name, suffix: monitor, isFocused: focus.workspace == $0)
     }
+    var items = sortedMonitors.map {
+        TrayItem(type: .monitor, name: $0.activeWorkspace.name, isActive: $0.activeWorkspace == focus.workspace && sortedMonitors.count > 1)
+    }
+    let mode = activeMode?.takeIf { $0 != mainModeId }?.first?.lets { TrayItem(type: .mode, name: $0.uppercased(), isActive: true) }
+    if let mode {
+        items.insert(mode, at: 0)
+    }
+    TrayMenuModel.shared.trayItems = items
 }
 
-struct WorkspaceViewModel {
+struct WorkspaceViewModel: Hashable {
     let name: String
     let suffix: String
     let isFocused: Bool
+}
+
+enum TrayItemType: String, Hashable {
+    case mode
+    case monitor
+}
+
+private let validNumbers = "0" ... "9"
+private let validLetters = "A" ... "Z"
+
+struct TrayItem: Hashable, Identifiable {
+    let type: TrayItemType
+    let name: String
+    let isActive: Bool
+    var systemImageName: String? {
+        // System image type is only valid for single number and single capital char workspace name
+        guard name.count == 1 else { return nil }
+        guard validNumbers.contains(name) || validLetters.contains(name) else { return nil }
+        let lowercasedName = name.lowercased()
+        switch type {
+            case .mode:
+                return "\(lowercasedName).circle"
+            case .monitor:
+                if isActive {
+                    return "\(lowercasedName).square.fill"
+                } else {
+                    return "\(lowercasedName).square"
+                }
+        }
+    }
+    var id: String {
+        return type.rawValue + name
+    }
 }
--- a/Sources/AeroSpaceApp/AeroSpaceApp.swift
+++ b/Sources/AeroSpaceApp/AeroSpaceApp.swift
@@ -8,6 +8,7 @@ import SwiftUI
 struct AeroSpaceApp: App {
     @MainActor // macOS 13
     @StateObject var viewModel = TrayMenuModel.shared
+    @StateObject var appearance = Appearance.shared
 
     init() {
         initAppBundle()
@@ -15,6 +16,6 @@ struct AeroSpaceApp: App {
 
     @MainActor // macOS 13
     var body: some Scene {
-        menuBar(viewModel: viewModel)
+        menuBar(viewModel: viewModel, appearance: appearance)
     }
 }
--- a/Sources/AppBundle/ui/Appearance.swift
+++ b/Sources/AppBundle/ui/Appearance.swift
@@ -1,0 +1,28 @@
+//import AppKit
+import SwiftUI
+
+enum AppearanceType {
+    case light
+    case dark
+}
+
+@MainActor
+public class Appearance: ObservableObject {
+    public static let shared = Appearance()
+
+    private var appearanceObserver: NSKeyValueObservation?
+
+    private init() {
+        appearanceObserver = NSApp.observe(\.effectiveAppearance) { [weak self] app, _ in
+            Task { @MainActor in
+                let name = app.effectiveAppearance.name
+                let isDarkAppearance = name == .vibrantDark || name == .darkAqua || name == .accessibilityHighContrastDarkAqua || name == .accessibilityHighContrastVibrantDark
+                self?.appAppearance = isDarkAppearance ? .dark : .light
+            }
+        }
+    }
+
+    /// System Settings -> Appearance -> Light/Dark
+    /// This is the theme representing how the UI should look inside the app (this might be different than the menu bar color)
+    @Published var appAppearance: AppearanceType = .dark
+}
--- a/Sources/AppBundle/ui/Appearance.swift
+++ b/Sources/AppBundle/ui/Appearance.swift
@@ -1,4 +1,3 @@
-//import AppKit
 import SwiftUI
 
 enum AppearanceType {
--- a/Sources/AppBundle/ui/ExperimentalUISettings.swift
+++ b/Sources/AppBundle/ui/ExperimentalUISettings.swift
@@ -1,4 +1,3 @@
-import AppKit
 import SwiftUI
 
 struct ExperimentalUISettings {
--- a/Sources/AppBundle/ui/ExperimentalUISettings.swift
+++ b/Sources/AppBundle/ui/ExperimentalUISettings.swift
@@ -22,9 +22,8 @@ enum ExperimentalUISettingsItems: String {
 }
 
 @MainActor
-func getExperimentalUISettingsMenu(viewModel: TrayMenuModel) -> some View {
-    let appearence = NSApplication.shared.effectiveAppearance.name
-    let color = (appearence == .vibrantDark || appearence == .darkAqua) ? Color.white : Color.black
+func getExperimentalUISettingsMenu(viewModel: TrayMenuModel, appearance: Appearance) -> some View {
+    let color = appearance.appAppearance == .dark ? Color.white : Color.black
     return Menu {
         Text("Menu bar style (macOS 14 or later):")
         Button {
--- a/Sources/AppBundle/ui/ExperimentalUISettings.swift
+++ b/Sources/AppBundle/ui/ExperimentalUISettings.swift
@@ -26,43 +26,40 @@ func getExperimentalUISettingsMenu(viewModel: TrayMenuModel) -> some View {
     let appearence = NSApplication.shared.effectiveAppearance.name
     let color = (appearence == .vibrantDark || appearence == .darkAqua) ? Color.white : Color.black
     return Menu {
-        Text("These settings are EXPERIMENTAL.")
-        Text("No stability guarantees are provided.")
-        Divider()
         Text("Menu bar style (macOS 14 or later):")
-        Text(MenuBarStyle.systemText.title)
-        Button {
-            viewModel.experimentalUISettings.displayStyle = .systemText
-        } label: {
-            Toggle(isOn: .constant(viewModel.experimentalUISettings.displayStyle == .systemText)) {
-                MenuBarLabel(viewModel.trayText, textStyle: .system, color: color)
-            }
-        }
-        Text(MenuBarStyle.monospacedText.title)
         Button {
             viewModel.experimentalUISettings.displayStyle = .monospacedText
         } label: {
             Toggle(isOn: .constant(viewModel.experimentalUISettings.displayStyle == .monospacedText)) {
                 MenuBarLabel(viewModel.trayText, color: color)
+                Text(" -  " + MenuBarStyle.monospacedText.title)
             }
         }
-        Text(MenuBarStyle.squares.title)
+        Button {
+            viewModel.experimentalUISettings.displayStyle = .systemText
+        } label: {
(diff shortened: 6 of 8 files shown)


danfbfern:
I understand that this topic is not something important for @nikitabobko, but since it still allows to have the current behaviour for users that like that simplicity, is there any constraint to prevent this from being merged? Any drawbacks of having it?
@mobile-ar keeps rebasing it to keep the topic alive for a few months

Sources/AppBundle/TrayMenuModel.swift
```diff
@@ -15,20 +16,52 @@ public class TrayMenuModel: ObservableObject {
...
+        activeMode?.takeIf { $0 != mainModeId }?.first?.lets {
+            items.append(TrayItem(type: .mode, name: String($0), isActive: true))
+            return "[\($0.uppercased())] "
+        } ?? ""
+    ) +
         sortedMonitors
         .map {
-            ($0.activeWorkspace == focus.workspace && sortedMonitors.count > 1 ? "*" : "") + $0.activeWorkspace.name
+            items.append(TrayItem(type: .monitor, name: $0.activeWorkspace.name, isActive: $0.activeWorkspace == focus.workspace && sortedMonitors.count > 1))
+            return ($0.activeWorkspace == focus.workspace && sortedMonitors.count > 1 ? "*" : "") + $0.activeWorkspace.name
```
nikitabobko, line 29: This code is written in a functional style. Please avoid side effects (`items.append`) from the lambda of `map`

nikitabobko:
I've reflected on this a bit more: Sorry, I just don't care. I hate programming UIs, I hate maintaining UIs, I hate providing configuration options for UIs – the configurations options space for UIs is enormous. For me, it's hard to come up with reasonable mental model which is powerful yet not excessive, and I just don't want to waste my mental energy on that.

I've read diagonally [the response](https://github.com/nikitabobko/AeroSpace/discussions/844#discussioncomment-11624441) from @mobile-ar and IDK, I still don't care, sorry 🤷 I appreciate the detailed response though!

The reason why I can't just merge this PR is because it touches things that I care about.
1. I care about the config format. The format should be stable, and its mental model should be clear.
2. I care about the code maintenance. It's a less of critical concern since the code in this PR is approachable and looks easily maintainable. @mobile-ar thanks for that!

**So I suggest an alternative** – please extract UI settings into a UI settings window + UserDefaults (or whatever default technology macOS apps usually use to manage settings) and don't pollute the config with UI settings. If this PR add a button like "Experimental Settings" to the dropdown, I'm ok to merge it.

Inside the window, please add a disclaimer like "These settings are EXPERIMENTAL. No stability guarantees are provided". Please also make it clear that the settings only work for macOS >= 14

Maybe, if eventually this UI logic irons out, I will be ok to make it configurable through the TOML config.

But be aware that if the UI logic turns out being too problematic to maintain, I will just drop it (hence the "experimental" in the name). I will be reasonable – I will do my minimal effort to maintain it (do the migrations if necessary), or ping @mobile-ar to ask for help (hopefully, he will be willing to help)

zhenpewu:
Hi,

I’ve been following this PR for a while and noticed the maintainer’s stance on UI over-customization, which I totally agree with. However, the monospaced menu bar feels excessively wide for multi-monitor workspaces. Would it be possible to provide an option to use the previous font instead of mono?

https://github.com/nikitabobko/AeroSpace/discussions/1112

Sources/AppBundle/TrayMenuModel.swift
mobile-ar: Sure, I was just trying to reuse the iteration through the collection for performance reasons, to avoid scanning the collection twice to get the values needed for the TrayItem. I can decouple the logic from this map, but I might need to introduce a quite similar one for populating the items collection.

Sources/AppBundle/MenuBar.swift
```diff
@@ -83,6 +94,87 @@ struct MonospacedText: View {
...
+extension String {
+    func parseMenuBarStyle() -> MenuBarStyle? {
+        if let parsed = MenuBarStyle(rawValue: self) {
+            return parsed
+        } else {
+            return nil
+        }
+    }
+
+    func containsEmoji() -> Bool {
```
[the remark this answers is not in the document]
mobile-ar: Because an emoji inside the square images are way too small to be visible, then when someone uses emojis for the name of a workspace I just display the emoji, instead of the representation of the letter for the workspace. (Added this for support, emojis in the menu bar are quite annoying on my opinion 😄)

mobile-ar:
@nikitabobko Sure, I will be keen to move this to an "Experimental UI Settings" in the menu option and remove the TOML config for this feature. To be honest I thought you preferred all configuration for the app to be in the TOML file, that's why I implemented in that way. 

I'm totally ok with the experimental UI settings approach and save some sort of mac "UserDefauls".
The only issue might be people who likes to keep their .config files won't be able to save this config when moving between computers. (Image personal one, work one, or even on a new machine).

Sources/AppBundle/TrayMenuModel.swift
mobile-ar: Also I use the same approach in line 23, to get the item related to the mode. I store all those inside `items`, given that in that way I have all the items to display (modes + workspaces) in an object structure instead of only text.

nikitabobko: > but I might need to introduce a quite similar one for populating the items collection.

That's fine

> Also I use the same approach in line 23

Thanks for pointing out. Please fix it too.

Sources/AppBundle/MenuBar.swift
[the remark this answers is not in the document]
nikitabobko: Please leave a comment in the code that they are way too small to be visible in square images

nikitabobko:
> To be honest I thought you preferred all configuration for the app to be in the TOML file, that's why I implemented in that way.

Yes, normally, I prefer plain text configuration. This case is different

> The only issue might be people who likes to keep their .config files won't be able to save this config when moving between computers

Yes, the suggested solution is a compromise

mobile-ar:
@nikitabobko I have updated the PR. I tried to extract the Experimental UI Settings as much as possible into the `ExperimentalUISettings.swift` in case you want to remove in the future.
I added 1 extra option to filter the empty workspaces from the menu list (because I think is was a nice to have, also was requested in comments under #711)

Looks like this now when filtered:
image

And like this when not filtered:
image

mobile-ar:
> Inside the window, please add a disclaimer like "These settings are EXPERIMENTAL. No stability guarantees are provided". Please also make it clear in the window that the settings work only for macOS >= 14

I just added this into the setting. I forgot this part in previous commit.
Now looks like this with the requested disclaimer:

image

Another option for the `menu bar style` submenu is to remove it completely putting it under `if #available(macOS 14, *)` so it won't even be shown in older versions of macOS.

manuelgcsousa:
Any news on this PR?

zhenpewu:
@mobile-ar I forked your changes https://github.com/zhenpewu/AeroSpace/actions/runs/14506543264 and publish a build artifacts for macOS 15.

cc: @manuelgcsousa

zhenpewu:
Screenshot 2025-04-16 at 19 33 38

One nit-picky bit of feedback: the build feels a bit too large for i3-like or square layouts. I think it would be better to adjust the sizing—maybe wrap the Workspace in a square that matches the size of a monospace character. Thank you.

mobile-ar:
@zhenpewu both i3 style and squares are using an image representation using the native system images (you can find the reference in the SF symbols app) each square is same size as the other. These are not monospaced characters wrapped in any sort of square. 
image

It's quite noticeable the empty space with numbers like 1. But I usually prefer letters, and it's less noticeable there.

image

zhenpewu:
Hey!

Just to clarify—I didn’t mean the spacing inside the square, but the size of the square itself in the menu. My Swift knowledge is pretty limited, but I tried a quick tweak on my end: adding `.frame(width: 32, height: 32)` to the Image made it look a lot better. Without that, it defaults to 40, which feels a bit too big—especially since the menu bar caps at 44.

https://github.com/zhenpewu/AeroSpace/commit/d41b1eb02db78a3e8683cea9f0266d5bc7699e4d

Hope that helps!

mobile-ar:
@zhenpewu By Apple Design Human Interface Guideline (https://developer.apple.com/design/human-interface-guidelines/the-menu-bar#Menu-bar-extras) the menu bar extra height is 24pt (times 2 as per retina), that's why I set the height to 40 (because the image will be scaled with a factor of 2) that will make the image 20pt. I feel more than enough padding while maintaining good readability inside the square.

Sources/AppBundle/ExperimentalUISettings.swift
```diff
@@ -0,0 +1,62 @@
...
+    }
+}
+
+enum ExperimentalUISettingsItems: String {
+    case displayStyle
+    case filterEmptyWorkspacesFromMenu
+}
+
+@MainActor
+func getExperimentalUISettingsMenu(viewModel: TrayMenuModel) -> some View {
```
nikitabobko, line 35: UX problem: users don't know how the tray is going to look like without clicking it. But when they click it, the menu disappears, which is not nice

I see several ways to fix it:
1. Change the tray icon, when the mouse is hovered over the item (it may introduce tray icon and menu jaggering or not)
2. Don't hide the menu when user clicks on items (it can also potentially may introduce tray icon and menu jaggering)
3. Extract this submenu into a real NSWindow
4. Maybe something else...

You are free to choose yourself

Sources/AppBundle/ExperimentalUISettings.swift
```diff
@@ -0,0 +1,62 @@
...
+            } else {
+                return .monospacedText
+            }
+        }
+        set {
+            UserDefaults.standard.setValue(newValue.rawValue, forKey: ExperimentalUISettingsItems.displayStyle.rawValue)
+            UserDefaults.standard.synchronize()
+        }
+    }
+    var filterEmptyWorkspacesFromMenu: Bool {
```
nikitabobko, line 18: Please drop this config option for now. I plan to address the problem with `persistent-workspaces` config option: https://github.com/nikitabobko/AeroSpace/issues/999

Sources/AppBundle/TrayMenuModel.swift
```diff
@@ -25,10 +27,42 @@ public class TrayMenuModel: ObservableObject {
...
+        switch type {
+            case .mode:
+                return "\(lowercasedName).circle"
+            case .monitor:
+                if isActive {
+                    return "\(lowercasedName).square.fill"
+                } else {
+                    return "\(lowercasedName).square"
+                }
+        }
```
nikitabobko, line 66: This is fragile

Sources/AppBundle/MenuBar.swift
```diff
@@ -92,6 +110,96 @@ struct MonospacedText: View {
...
+        return self.hashValue
+    }
+    var title: String {
+        switch self {
+            case .monospacedText:
+                "Monospaced font"
+            case .systemText:
+                "System font"
+            case .squares:
+                "Square images"
```
nikitabobko, line 192: The problem with this style is that the tray icon disappears if users do smth like this: `aerospace workspace dev`

nikitabobko (commented):
Sorry, I don't have enough time lately to review all the PRs, and I didn't take a thorough look at the code yet, but here are my comments so far

I also somehow managed to make the tray icon completely disappear when running your code. Can you please make it clear how to clear UserDefaults settings from the CLI. I tried something like `defaults delete bobko.aerospace.debug` but it didn't work.

Sources/AppBundle/MenuBar.swift
```diff
@@ -92,6 +110,96 @@ struct MonospacedText: View {
...
+            case .i3:
+                "i3 style"
+        }
+    }
+}
+
+extension String {
+    func containsEmoji() -> Bool {
```
nikitabobko, line 200:
```suggestion
    // If workspace name contains emojis we use the plain emoji in text to avoid visibility issues scaling the emoji to fit the squares
    fileprivate func containsEmoji() -> Bool {
```

Sources/AppBundle/ExperimentalUISettings.swift
mobile-ar: Removed.

Sources/AppBundle/ExperimentalUISettings.swift
mobile-ar: So I went with something else 😄 (given that not closing the menu on selection looks not possible in SwiftUI as per my investigation) to keep it simple and to use the same UI to display how the menu bar will look in real time (reusing the MenuLabel).
Looks like this:

image

Sources/AppBundle/TrayMenuModel.swift
mobile-ar: True that. I improved this by using this only for the charset supported with system images (1 to 9 and A to Z). So this is now optional and if nil the UI is layout in a custom way.

Sources/AppBundle/MenuBar.swift
mobile-ar: Fixed / improved how long name workspaces are displayed for `squares` and `i3` styles

Sources/AppBundle/MenuBar.swift
mobile-ar: Fixed

mobile-ar:
> Sorry, I don't have enough time lately to review all the PRs, and I didn't take a thorough look at the code yet, but here are my comments so far
> 
> I also somehow managed to make the tray icon completely disappear when running your code. Can you please make it clear how to clear UserDefaults settings from the CLI. I tried something like `defaults delete bobko.aerospace.debug` but it didn't work.

You can delete it using `defaults delete AeroSpaceApp` for the debug one. (Remember that you need to trigger some change for the menu to refresh the value, UserDefaults it's not an observable object)

Sources/AppBundle/ExperimentalUISettings.swift
```diff
@@ -0,0 +1,68 @@
...
+}
+
+@MainActor
+func getExperimentalUISettingsMenu(viewModel: TrayMenuModel) -> some View {
+    let appearence = NSApplication.shared.effectiveAppearance.name
+    let color = (appearence == .vibrantDark || appearence == .darkAqua) ? Color.white : Color.black
+    return Menu {
+        Text("These settings are EXPERIMENTAL.")
+        Text("No stability guarantees are provided.")
+        Divider()
```
nikitabobko, line 31:
```suggestion
```

Sources/AppBundle/ExperimentalUISettings.swift
```diff
@@ -0,0 +1,68 @@
...
+        Text(MenuBarStyle.i3.title)
+        Button {
+            viewModel.experimentalUISettings.displayStyle = .i3
+        } label: {
+            Toggle(isOn: .constant(viewModel.experimentalUISettings.displayStyle == .i3)) {
+                MenuBarLabel(viewModel.trayText, color: color, trayItems: viewModel.trayItems, workspaces: viewModel.workspaces)
+            }
+        }
+    } label: {
+        Text("Experimental UI Settings")
```
nikitabobko, line 66:
```suggestion
        Text("Experimental UI Settings (No stability guarantees)")
```

Sources/AppBundle/ExperimentalUISettings.swift
nikitabobko: For some reason if I drop these lines, I get a segmentation fault if I run aerospace like `./run-debug.sh`

Sources/AppBundle/ExperimentalUISettings.swift
```diff
@@ -0,0 +1,68 @@
...
+            }
+        }
+        Text(MenuBarStyle.monospacedText.title)
+        Button {
+            viewModel.experimentalUISettings.displayStyle = .monospacedText
+        } label: {
+            Toggle(isOn: .constant(viewModel.experimentalUISettings.displayStyle == .monospacedText)) {
+                MenuBarLabel(viewModel.trayText, color: color)
+            }
+        }
```
nikitabobko, line 48: I think monospaced text should be the first as its the default

Sources/AppBundle/ExperimentalUISettings.swift
nikitabobko: Actually, if I do almost any changes to this file, then the next time I run the debug build from CLI, I get a segmentation fault.

I have learnt that I can resolve the segmentation fault via `rm -rf .build/arm64-apple-macosx`, but it's not a nice dev experience. I understand that it's probably a swift bug, but we have to come up with some workaround (rewriting code in some way to avoid these incremental compilation bugs would be sufficient)

I have the following swift version installed via swiftly

```
Apple Swift version 6.1 (swift-6.1-RELEASE)
Target: arm64-apple-macosx15.0
```

I observe the same problem with other Swift setups:
- If I run debug build from CLI using swift 6.1 from Xcode
- If I run debug build from Xcode itself

nikitabobko: Example of a crash:

```
💣 Program crashed: Bad pointer dereference at 0x0000000000000009

Thread 0 crashed:

 0 0x0000000196e74fdc swift_retain + 24 in libswiftCore.dylib
 1 0x0000000196ea36cc swift::OpaqueValue* tuple_initializeWithCopy<false, false>(swift::OpaqueValue*, swift::OpaqueValue*, swift::TargetMetadata<swift::InProcess> const*) + 104 in libswiftCore.dylib
 2 0x00000001ebe40b34 <deduplicated_symbol> + 64 in SwiftUICore
 3 0x00000001b590c618 initializeWithCopy for Menu + 156 in SwiftUI
 4 menuBar(viewModel:) + 136 in AeroSpaceApp at /Users/bobko/a/AeroSpace/Sources/AppBundle/MenuBar.swift:7:5

     5│ @MainActor
     6│ public func menuBar(viewModel: TrayMenuModel) -> some Scene { // todo should it be converted to "SwiftUI struct"?
     7│     MenuBarExtra {
      │     ▲
     8│         let shortIdentification = "\(aeroSpaceAppName) v\(aeroSpaceAppVersion) \(gitShortHash)"
     9│         let identification      = "\(aeroSpaceAppName) v\(aeroSpaceAppVersion) \(gitHash)"

 5 AeroSpaceApp.body.getter + 228 in AeroSpaceApp at /Users/bobko/a/AeroSpace/Sources/AeroSpaceApp/AeroSpaceApp.swift:18:9

    16│     @MainActor // macOS 13
    17│     var body: some Scene {
    18│         menuBar(viewModel: viewModel)
      │         ▲
    19│     }
    20│ }

 6 0x00000001ec12c6e8 BodyAccessor.setBody(_:) + 1204 in SwiftUICore
 7 0x00000001b668ef70 AppBodyAccessor.updateBody(of:changed:) + 152 in SwiftUI
 8 0x00000001ec12edd8 closure #1 in DynamicBody.updateValue() + 600 in SwiftUICore
 9 0x00000001ec12e804 DynamicBody.updateValue() + 928 in SwiftUICore
10 0x00000001b6920728 AG::Graph::UpdateStack::update() + 524 in AttributeGraph
11 0x00000001b6920f94 AG::Graph::update_attribute(AG::data::ptr<AG::Node>, unsigned int) + 420 in AttributeGraph
12 0x00000001b692911c AG::Graph::input_value_ref_slow(AG::data::ptr<AG::Node>, AG::AttributeID, unsigned int, unsigned int, AGSwiftMetadata const*, unsigned char&, long) + 712 in AttributeGraph
13 0x00000001b693faf8 AGGraphGetValue + 228 in AttributeGraph
14 0x00000001ec12e714 DynamicBody.updateValue() + 688 in SwiftUICore
15 0x00000001b6920728 AG::Graph::UpdateStack::update() + 524 in AttributeGraph
16 0x00000001b6920f94 AG::Graph::update_attribute(AG::data::ptr<AG::Node>, unsigned int) + 420 in AttributeGraph
17 0x00000001b692911c AG::Graph::input_value_ref_slow(AG::data::ptr<AG::Node>, AG::AttributeID, unsigned int, unsigned int, AGSwiftMetadata const*, unsigned char&, long) + 712 in AttributeGraph
18 0x00000001b693faf8 AGGraphGetValue + 228 in AttributeGraph
19 0x00000001b53ffe7c ResolvedMenuBarExtra.MakeList.updateValue() + 548 in SwiftUI
20 0x00000001b53d7528 specialized implicit closure #1 in closure #1 in closure #1 in Attribute.init<A>(_:) + 24 in SwiftUI
21 0x00000001b6920728 AG::Graph::UpdateStack::update() + 524 in AttributeGraph
22 0x00000001b6920f94 AG::Graph::update_attribute(AG::data::ptr<AG::Node>, unsigned int) + 420 in AttributeGraph
23 0x00000001b69287b0 AG::Graph::value_ref(AG::AttributeID, unsigned int, AGSwiftMetadata const*, unsigned char&) + 272 in AttributeGraph
24 0x00000001b693fb48 AGGraphGetValue + 308 in AttributeGraph
25 0x00000001b562a444 AppGraph.filteredPrimarySceneSummaries.getter + 3496 in SwiftUI
26 0x00000001b668e76c AppGraph.instantiateOutputs() + 480 in SwiftUI
27 0x00000001ec303b34 GraphHost.instantiate() + 224 in SwiftUICore
28 0x00000001b569ce88 closure #1 in closure #1 in runApp<A>(_:) + 88 in SwiftUI
29 0x00000001ec132d28 closure #1 in static Update.ensure<A>(_:) + 56 in SwiftUICore
30 0x00000001ec131b04 static Update.ensure<A>(_:) + 96 in SwiftUICore
31 0x00000001b569ce1c runApp<A>(_:) + 124 in SwiftUI
32 0x00000001b59fcd14 static App.main() + 224 in SwiftUI

Backtrace took 1.50s

/Users/bobko/a/AeroSpace/run-debug.sh: line 6: 74831 Segmentation fault: 11  ./.debug/AeroSpaceApp "$@"
```

Sources/AppBundle/ExperimentalUISettings.swift
nikitabobko: Thanks, that's also a nice option. Now the problem is that when it takes 2 rows, it looks too bulky, it'd be nice to merge 2 rows into a single row

Sources/AppBundle/TrayMenuModel.swift
```diff
@@ -25,10 +27,51 @@ public class TrayMenuModel: ObservableObject {
...
-struct WorkspaceViewModel {
+struct WorkspaceViewModel: Hashable {
     let name: String
     let suffix: String
     let isFocused: Bool
 }
+
+enum TrayItemType: String, Hashable {
+    case mode
+    case monitor
```
nikitabobko, line 48: Why do you call it `monitor` if it's for workspaces?

Anyway, I think better names for these two would be: `circle` and `square`

nikitabobko:
Can you please everything related to UI to `Sources/AppBundle/ui`?

Sources/AppBundle/MenuBar.swift
```diff
@@ -26,6 +26,8 @@ public func menuBar(viewModel: TrayMenuModel) -> some Scene { // todo should it
             }
```
nikitabobko, line 26: IDK if it's something you are interested, but I'd be happy to refactor this `func menuBar` function to a more canonical Swift UI code (in a separate PR). I know nothing about Swift UI and I just wrote it as a function because it's the way people program in Jetpack Compose (Android reactive UI framework, which I am happen to be familiar with)

But I have a feeling that a canonical way to split UI into a components is via structs in Swift UI. I have a feeling that you might know Swift UI better

nikitabobko: The same about `func getExperimentalUISettingsMenu` & `MenuBarLabel.itemView` which you introduce in this PR

Sources/AppBundle/TrayMenuModel.swift
nikitabobko: Ok, given this line: `if item.type == .mode`, I see how `mode` & `workspace` names are better

Sources/AppBundle/MenuBar.swift
```diff
@@ -62,25 +64,51 @@ public func menuBar(viewModel: TrayMenuModel) -> some Scene { // todo should it
...
+            switch viewModel.experimentalUISettings.displayStyle {
+                case .systemText:
+                    MenuBarLabel(viewModel.trayText, textStyle: .system)
+                case .monospacedText:
+                    MenuBarLabel(viewModel.trayText)
+                case .squares:
+                    MenuBarLabel(viewModel.trayText, trayItems: viewModel.trayItems)
+                case .i3:
+                    MenuBarLabel(viewModel.trayText, trayItems: viewModel.trayItems, workspaces: viewModel.workspaces)
+            }
```
nikitabobko, line 76: Right now, you unconditionally pass `viewModel.trayText` even if it's not used inside (for square and i3 styles)

I feel like pairs of systemText/monospacedText styles and pairs of squares/i3 styles are significantly different from each other to dedicate separate `MenuBarLabel`s for the two groups

Sources/AppBundle/MenuBar.swift
```diff
@@ -62,25 +64,51 @@ public func menuBar(viewModel: TrayMenuModel) -> some Scene { // todo should it
...
-            MonospacedText("⏸️")
+            MenuBarLabel("⏸️")
         }
     }
 }
 
-struct MonospacedText: View {
+@MainActor
+struct MenuBarLabel: View {
     @Environment(\.colorScheme) var colorScheme: ColorScheme
```
nikitabobko, line 85: Those tricks with `@Environment(\.colorScheme)` and `NSApplication.shared.effectiveAppearance` inside `getExperimentalUISettingsMenu` is something that makes me think if it's the the way it should be done. They are related yet apparently they mean different things

Do you have any pointers?

Sources/AppBundle/MenuBar.swift
```diff
@@ -92,6 +120,129 @@ struct MonospacedText: View {
...
+                    ForEach(trayItems, id: \.id) { item in
+                        itemView(for: item)
+                        if item.type == .mode {
+                            Text(":")
+                                .font(.system(.largeTitle, design: textStyle.design))
+                                .foregroundStyle(finalColor)
+                                .bold()
+                        }
+                    }
+                    if workspaces != nil {
```
nikitabobko, line 137: You pass `workspaces` array but you don't use it in `MenuBarLabel`

nikitabobko:
Two digit workspaces sometimes don't look nice (on my screenshot, it's fine for square style, but not ok for i3 style)

image

Sources/AppBundle/ExperimentalUISettings.swift
mobile-ar: I'm not able to reproduce this issue. I can edit this file and never get any issues. I also run this this trough the CLI all the time.
image


I have also swift 6.1
image

I will try to do some more investigation to see if I can reproduce this.

Sources/AppBundle/MenuBar.swift
mobile-ar: Yes, this was quite tricky, I couldn't find a lot of documentation. But when using the `MenuBarLabel` inside the menu, the color of the colorScheme was not reflecting the change of color.

There is one color for System -> Appearance (`NSApplication.shared.effectiveAppearance`)
image

And other for System -> Wallpaper
image

When you start combining this 2 macOS system configurations `@Environment(\.colorScheme)` is not getting the correct value inside the menu. But when combining `colorScheme` and `effectiveAppearance`) it started working for all the combinations.

Sources/AppBundle/MenuBar.swift
mobile-ar: It's used to check if I need to display the workspaces and as a trick redraw the view on changes when the workspace gets a window added.
What I noticed is that I can add 1 extra property `isEffectivelyEmpty` to `WorkspaceViewModel` and then use workspaces instead of `Workspace.all.filter ...`

Sources/AppBundle/MenuBar.swift
mobile-ar: TrayText is always used as a fallback for the imageRenderer in case of failure and as a label for the image. It's called text here:
`Image(cgImage, scale: 2, label: Text(text))` line 114 (body of `MenuBarLabel`)

Maybe should I rename the parameter to be `trayText` to better reflect this?

Sources/AppBundle/MenuBar.swift
nikitabobko: Thank you, so it wasn't some guide on the internet, but you figured it out with series of trials and errors.

Do I get it correctly that you have figured what combinations of `@Environment(\.colorScheme)` and `NSApplication.shared.effectiveAppearance` correspond to which Light/Dark Appearance and Light/Dark Wallpaper?

If so, I'd suggest to refactor the code to make it self-documenting (if Apple doesn't make a nice API, we will do it ourselves):

```swift
// System Settings -> Appearance -> Light/Dark
enum AppearanceTheme {
  case light
  case dark

  static var current: AppearanceTheme {
    // @Environment(\.colorScheme) and NSApplication.shared.effectiveAppearance magic...
  }
}

// System Settings -> Wallpaper -> Light/Dark
enum WallpaperTheme {
  case light
  case dark

  static var current: WallpaperTheme {
    // @Environment(\.colorScheme) and NSApplication.shared.effectiveAppearance magic...
  }
}
```

Would smth like this make sense? Would this API be easier to use?

Sources/AppBundle/MenuBar.swift
```diff
@@ -92,6 +120,129 @@ struct MonospacedText: View {
...
+                        let otherWorkspaces = Workspace.all.filter { workspace in
+                            !workspace.isEffectivelyEmpty && !trayItems.contains(where: { item in item.type == .monitor && item.name == workspace.name })
```
nikitabobko, line 139: Realization. It's unnecessary O(n^2):
```swift
/*...*/.filter {
  /*...*/ && !trayItems.contains(/*...*/)
}
```

Please avoid it. I think it will be fixed automatically once this comment is resolved https://github.com/nikitabobko/AeroSpace/pull/1049/files#r2059296197

Sources/AppBundle/MenuBar.swift
nikitabobko: >  What I noticed is that I can add 1 extra property isEffectivelyEmpty to WorkspaceViewModel and then use workspaces instead of Workspace.all.filter ...

Yes, that would be better

Sources/AppBundle/TrayMenuModel.swift
nikitabobko: But I still have a question why `monitor` and not `workspace` name

Sources/AppBundle/MenuBar.swift
nikitabobko: >  What I noticed is that I can add 1 extra property isEffectivelyEmpty to WorkspaceViewModel and then use workspaces instead of Workspace.all.filter ...

That would be better, thanks

Sources/AppBundle/MenuBar.swift
mobile-ar: I have seen both implementations in SwiftUI both functions and structs, usually is common to use functions when the view receives parameters. 

Sure I will be interested in doing the refactoring later in another PR. I think I tried that in one of my views and I was having some weird behavior (some changes not refreshing the UI) but I could take a look at this.

Sources/AppBundle/ExperimentalUISettings.swift
mobile-ar: Menu is not giving a lot of options to make it nicer. I will prefer a spacer between the image and the text and align one to leading and other to trailing but it's not possible.
This is the best looking 1 line I could think of: 
image

Sources/AppBundle/ExperimentalUISettings.swift
mobile-ar: Fixed

Sources/AppBundle/ExperimentalUISettings.swift
mobile-ar: Done

Sources/AppBundle/MenuBar.swift
mobile-ar: Fixed

Sources/AppBundle/MenuBar.swift
mobile-ar: I still need this to filter the active monitors workspaces, to not display it duplicated ones. Maybe I need to filter this in some other way that I'm not aware.
I could not filter workspaces inside TrayMenuModel, because they are used in the main menu to display all workspaces.

Also trayItems array should be small (macOS only supports 5 monitors in it best hardware) + 1 mode <- (Not a justification, I will try to think how to improve this)

Sources/AppBundle/MenuBar.swift
nikitabobko: Makes sense, thanks

mobile-ar:
> Two digit workspaces sometimes don't look nice (on my screenshot, it's fine for square style, but not ok for i3 style)
> 
> Especially, given that `dev` workspace looks nice
> 
> image

I added a fix for this. There is support for numbers in system images until 50. I added that fix. Still not sure why this 2 digits numbers did not display correctly. I will investigate this more.

Sources/AppBundle/MenuBar.swift
mobile-ar: I don't think this is possible, `@Environment` is a `DynamicProperty`
```swift
/// An interface for a stored variable that updates an external property of a
/// view.
///
/// The view gives values to these properties prior to recomputing the view's
/// ``View/body-swift.property``.
@available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.0, *)
public protocol DynamicProperty {
```
And also from `@Environment` init documentation
```swift
...
///     struct MyView: View {
///         @Environment(\.colorScheme) var colorScheme: ColorScheme
///
///         // ...
///     }
///
/// SwiftUI automatically updates any parts of `MyView` that depend on
/// the property when the associated environment value changes.
...
```
Also even if we try to access `EnvironmentValues` manually, we will only get defaults values, and not the live updates
```swift
@available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.0, *)
public struct EnvironmentValues : CustomStringConvertible {

/// Creates an environment values instance.
///
/// You don't typically create an instance of ``EnvironmentValues``
/// directly. Doing so would provide access only to default values that
/// don't update based on system settings or device characteristics.
/// Instead, you rely on an environment values' instance
/// that SwiftUI manages for you when you use the ``Environment``
/// property wrapper and the ``View/environment(_:_:)`` view modifier.
public init()
```

I don't think this will be possible without some hack like having an empty view that uses this property and hook on change of `colorScheme` to update a `@Published` property inside the our custom `Appearance: ObservableObject`

Sources/AppBundle/MenuBar.swift
mobile-ar: Fixed by adding isVisible to the WorkspaceViewModel

Sources/AppBundle/TrayMenuModel.swift
mobile-ar: Fixed

Sources/AppBundle/ExperimentalUISettings.swift
mobile-ar: @nikitabobko  I tried adding changes to that same file with Zed and vim, but still running normally for me with `./run-debug.sh`. 
I have installed Xcode 16.2 but I'm actively using Xcode 16.3.
I can't reproduce this issue. Also it's super weird issue for a fairly common file, there is nothing complex on it to produce something like this as per my understanding.

nikitabobko:
@mobile-ar thanks for your work and patience! Merged

mobile-ar:
> @mobile-ar thanks for your work and patience! Merged

All good 😄 I'm happy to contribute. Thank you for creating and maintaining this awesome tool.

mobile-ar:
> I didn't understand what the complication around `appearanceObserver` is for, so I cleaned up your code [9f49504](https://github.com/nikitabobko/AeroSpace/commit/9f49504049d0bc90826e87a33ff82db877284483)
> 
> Please let me know if it was important for some reason

It's important if you want to avoid an small case where the UI color inside the menu is incorrect.

If you open System Sesttings -> Appearance -> and you change the current Appearance. And without any other screen refresh or windows change you open the Aerospace menu -> Experimental UI, you will see the color is the previous one. 

It will show incorrectly like this:

image

That's why I made Appearance to be observable, to avoid this small issue. By making it observable, you ensure that on change of Appearance, the UI is refreshed correctly every time.

Sources/AppBundle/MenuBar.swift
mobile-ar: Yes, I don't recall any difference between both. Could be reusability at some point depending the scope of the functions, but it's not the case here. 

I think there is some small gain in performance from the function version, but it might be minimal.
Some source: https://www.swiftbysundell.com/articles/swiftui-views-versus-modifiers/

nikitabobko:
@mobile-ar thanks for the explanation. I think the appearance doesn't change that often, and I prefer to trade correctness for more straightforward code in this place.

It's unfortunate that Apple makes us write this boilerplate to avoid this bug.

volker-fr:
I want to express my disappointment that this PR was closed.

Would it be possible to have it as "permanently experimental" with the notion of "if it breaks, there will be zero support"?


In my humble opinion, this feature would resolve a major pain point that was worked around by various people in often not so nice or buggy ways. The sole reason why I have to use SketchyBar is due to the lack of seeing the workspace -- something this PR would solve. And Sketchybar has its own issues / conflicts with the official macOS bar.

I understand that this implementation is a bit hacky and not as smooth as hoped. At the same time, my impression is that the way AeroSpace has to work in a few areas to get around Apples limitation. One example here is requirement for [Proper monitor arrangement](https://nikitabobko.github.io/AeroSpace/guide#proper-monitor-arrangement). I personal would prefer a little bit more inconvenience here, over the lack of having this big missing feature.


In the end the decision is up to @nikitabobko , I still think that this feature would bring tremendous value.

zhenpewu:
@volker-fr this PR is closed, becasue @nikitabobko have merged it in his end.

PR #1057, opened 2025-02-03
Update guide.adoc - add missing word

wojtekolesinski:
Not sure if this should be on or off, but there is a word missing here :)


nikitabobko:
Merged, thanks

PR #1117, opened 2025-02-20
Add automatically-unhide-macos-hidden-apps-exceptions config option.

### LINKED ISSUE #1041: Add `automatically-unhide-macos-hidden-apps-exceptions = [...]` config option
The proposal:
```toml
automatically-unhide-macos-hidden-apps = true
automatically-unhide-macos-hidden-apps-exceptions = ['com.github.wez.wezterm']
```

1. Apps in the list should be excluded from the `automatically-unhide-macos-hidden-apps` logic.
2. Specifying non-empty `automatically-unhide-macos-hidden-apps-exceptions` array while `automatically-unhide-macos-hidden-apps = false` should be config error

Use cases:
1. Discord app is misbehaving. Instead of closing windows for real, it uses "hide app" when you try to close its window
2. Raycast supposedly uses "hide app" to toggle windows https://github.com/nikitabobko/AeroSpace/discussions/1040

MediosZ:
Fix #1041 

I tested both Raycast shortcuts and Discord, and they will not be unhidden automatically with `automatically-unhide-macos-hidden-apps-exceptions=['com.github.wez.wezterm', 'com.hnc.Discord']`

Currently, apps in exceptions will remember the workspace where they launched. When you reopen the app, it always appears on that workspace and focuses on that workspace.

### DIFF
--- a/Sources/AppBundle/GlobalObserver.swift
+++ b/Sources/AppBundle/GlobalObserver.swift
@@ -19,11 +19,22 @@ class GlobalObserver {
                    // "Hide app" (cmd-h) -> force focus
                    MacApp.allAppsMap.values.filter({ $0.nsApp.isHidden }).count == 1
                 {
-                    // Force focus
-                    _ = w.focusWindow()
-                    _ = w.nativeFocus()
+                    if let identifier = w.macAppUnsafe.nsApp.bundleIdentifier,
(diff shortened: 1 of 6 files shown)


MediosZ:
I think these "bugs" are caused by the hidden window being focused automatically. Not sure if we should address them.

Issue #1064, opened 2025-02-04
SIGABORT while parsing an invalid config


nikitabobko:
```
[[[on-window-detected]]
if.app-id = 'com.apple.findmy'
run = 'layout floating'

[on-window-detected]]
if.app-id = 'com.openai.chat'
run = 'layout floating'
```

The problem should be further investigated and reported to https://github.com/LebJe/TOMLKit and/or https://github.com/marzer/tomlplusplus/

Originally reported at: https://github.com/nikitabobko/AeroSpace/discussions/1061

fullmetalsheep:
Hi, 

I've debugged this for a bit and reported my findings against in the [TOMLKit Repo ](https://github.com/LebJe/TOMLKit/issues/30)

Additionally made a PR which prevents this from crashing AeroSpace here: https://github.com/nikitabobko/AeroSpace/pull/1134

nikitabobko:
Fixed by migrating to https://github.com/dduan/TOMLDecoder

PR #1134, opened 2025-02-24
ISSUE-#1064 Prevent crash caused by TOMLKit Parsing error

fullmetalsheep:
- Prevents TOMLKit table header parsing error from crashing AeroSpace by doing a pre-check for invalid table headers.

- Additionally throws a helpful user error showing the line number + content of the invalid config line

### DIFF
--- a/Sources/AppBundle/config/parseConfig.swift
+++ b/Sources/AppBundle/config/parseConfig.swift
@@ -1,5 +1,6 @@
 import AppKit
 import Common
+import Foundation
 import HotKey
 import TOMLKit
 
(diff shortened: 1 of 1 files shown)


Sources/AppBundle/config/parseConfig.swift
```diff
@@ -153,9 +157,26 @@ func parseCommandOrCommands(_ raw: TOMLValueConvertible) -> Parsed<[any Command]
     }
 }
 
+// Note: this is only required due to a TOML parsing bug in TOMLKit which throws SIGABORT in parsing failures
+// Should be removed once TOMLKit is patched
+private func validateTOMLTableHeaderSyntax(in input: String) throws {
+    // Matches [[ followed by at least one more [ While ignoring commented lines #
+    let pattern = #"^(?!\s*#)\s*\[\[(?=\[)"#
```
nikitabobko, line 164: This is fragile

nikitabobko (commented):
Thanks, but I don't think it's necessary to circumvent the problem on AeroSpace side. The problem is minor, and the workaround is fragile

PR #1148, opened 2025-02-27
Add 'fail' action to boundaries handling.

### LINKED ISSUE #1073: Feature: Support `--boundaries-action fail`
`fail` is the same as `stop`, but makes the command exit with non-zero exit code

Use case:
`aerospace focus right --boundaries-action fail || aerospace workspace next`
https://github.com/nikitabobko/AeroSpace/discussions/1072

MediosZ:
Fix #1073 

Output:
```bash
bash-3.2$ .debug/aerospace focus right --boundaries-action stop && echo succeed
succeed
bash-3.2$ .debug/aerospace focus right --boundaries-action fail || echo fail
fail
```


nikitabobko:
Thanks, merged

PR #1156, opened 2025-02-28
Support resizing floating windows.

### LINKED ISSUE #9: resize command: support floating windows
Right now, the command does nothing when invoked on floating windows

MediosZ:
May close #9 

I’ve noticed that resizing floating windows can be done independently of other cases, so I’ve written some separate code to handle it. 

I’m introducing a new concept called the ‘leading axis’ to handle resizing options like ‘smart | smart-opposite’. The leading axis is the longer axis of the floating window, like the X-axis if the width is greater than the height.

With ‘resize smart’, the leading axis will be adjusted while keeping the aspect ratio of the window the same.

And with ‘resize smart-opposite’, the non-leading axis will be adjusted.


fahadbr:
would love to have this! currently need to work around resizing using hammerspoon

Jaaco:
Really nice, was looking for this! @nikitabobko can we make this happen? :)

randomowo:
is there anything stops from making it happens? really need this feature

Issue #104, opened 2024-01-06
Optimize query commands performance


nikitabobko:
`aerospace list-workspaces` is slow. Well, `aerospace list-workspaces` is ~100ms which by itself is fine.

**(1)** But 100ms is the result when AeroSpace is idle. Currently AeroSpace performs all operations in one thread. Considering that Ax calls are blocking, 100ms becomes worse when a lot of Ax calls are performed. 

It can be tested with `alt-w = ['workspace W', 'exec-and-forget bash -c "time ~/.bin/aerospace list-workspaces" 2> ~/log']`. I have 9 windows opened, and I get 300ms in `~/log`. 300ms is not that nice

**(2)** The problem becomes worse when users need to call `aerospace list-workspaces` several times (E.g. to get all workspaces and focused workspaces). `list-workspaces` is designed to be called several times. Maybe the design could be reconsidered, or `--json` flag could be provided. With JSON format, AeroSpace could dump out all the info it knows about workspaces in a structured way.

Reported at: https://github.com/nikitabobko/AeroSpace/pull/81

liamaharon:
Note this gets extremely bad when CPU usage is saturated (e.g. when compiling a large Rust project), Aerospace becomes completely unresponsive.

The 'fix' I have in place for now is limiting the number of CPUs available to rustc.

CheariX:
I am using numerous aerospace CLI commands for my daily workflows (e.g., cycling through windows of active applications, focusing particular windows, etc.). The ~100 ms CLI became somewhat annoying to me since I started to feel the delay when I have ~300 ms in my scripts.

For this purpose, I experimented with some _hacky_ cache:

```
on-focus-changed = [
    'exec-and-forget aerospace list-windows --focused --format "%{window-id}" > /tmp/aerospace_cache_focused.json', 
    'exec-and-forget aerospace list-windows --all --json --format "%{window-id} %{window-title} %{app-bundle-id} %{app-name} %{app-pid} %{workspace} %{monitor-id} %{monitor-name}"  > /tmp/aerospace_cache_windows.json'
]
```

From there on, my scripts simply use `cat $file | jq`.
The impact is remarkable. My scripts now feel totally different.

I was wondering if some internal caching mechanism like that would be useful in general. For example, the focused window will not change until the "on-focus-changed" event is triggered. However, if I run it 3 times in a row, each run takes approximately 100 milliseconds.
[...]

Issue #278, opened 2024-06-15
Feature: shell-like combinators for conditional computation and sending several commands to server in a single batch


nikitabobko:
Currently it's impossible to consume stdout and exit codes of commands in toml config. One has to use `exec-and-forget`
```toml
alt-w = 'exec-and-forget aerospace workspace W || aerospace workspace-back-and-forth'
atl-tab = 'exec-and-forget aerospace list-workspaces --all | aerospace workspace next'
```

But the problem that it's slow. Communicating with the server back and forth from CLI client can take additional 100ms which becomes noticeable. Even if we fix the slowness somehow, annoying flickering will still remain an issue

It'd be cool if AeroSpace supported basic shell combinators (`||`, `&&`, `;`, `( )`, `!`)
```toml
alt-w = 'workspace W || workspace-back-and-forth'
atl-tab = 'list-workspaces --all | workspace next'
```

That's a big feature that lays the foundation for a lot of things (basically the combinators allow programming custom logic):
- It becomes possible to send a batch of commands from CLI to server at once, resulting in reduced flickering (because server already implements "double buffering" to reduce flickering)
  https://nikitabobko.github.io/AeroSpace/goodness#use-trackpad-gestures-to-switch-workspaces
  ```shell
  aerospace eval - <<< 'list-workspaces --monitor mouse --visible | xargs workspace; aerospace workspace next'
  ```
- It fixes the ugliness of `on-window-detected` TOML callback. The new syntax is much more compact and powerful. It remains readable for people familiar with shell combinators:
  ```toml
  on-window-detected = '''
      (test %{app-bundle-id} == intellij || test %{app-bundle-id} == intellij.community) && move-workspace-to-monitor I
          || test %{app-bundle-id} == com.google.Chrome && move-node-to-workspace W
          || test %{app-bundle-id} == com.apple.dt.Xcode && (move-node-to-workspace X; exec-async 'echo hi!')
  '''
  ```
- It opens up a possibility for even more powerful conditional gaps #60:
  ```
  [gaps]
  outer.left = 'test %{workspace-tiling-windows-count} <= 2 && echo 100'
  outer.right = 'test %{workspace-tiling-windows-count} <= 2 && echo 100'
  ```
- It makes everything more universal and reusable. E.g. there won't be need to fix #107 if `config` command will allow to query `on-window-detected` command:
  ```shell
  aerospace eval - <<< 'config --get on-window-detected | eval -'
  ```

[...]

Issue #131, opened 2024-01-22
Delay/freeze when switching focus & workspace


fmasa:
Hi!
Thanks for all the new features.

After updating to 0.8.0, I had AeroSpace disabled before I fixed the config file and I noticed that when AerosSpace is enabled there is a noticeable delay (I have no measurements, but it feels like hundreds of milliseconds at least) when either moving focus or changing workspaces. This also happens when I focus windows via `cmd+tab`, doing this without AeroSpace is instantaneous.

I know it's still beta, but I wanted to ask:

- Do other people experience this delay or am I missing some configuration
- Is this a performance issue in AeroSpace or some inherent limitation of Accessibility API?
- Is there any way to collect logs/diagnostics to find out more?

I'm on M1 Pro + Sonoma.
 
---

I'm not trying to bash AeroSpace - the feature set is great and I use it as a daily driver, but the MacOS felt laggy for a while and now I found out it's only when using AeroSpace and I want to help to make it better.

---

Note by @nikitabobko

Related: #242

nikitabobko:
**(1)** How many windows do you have opened? (you can use `aerospace list-windows --all | wc -l`) 
**(2)** Does the AeroSpace still behave poorly if you restart the app?

I saw on my colleague's machine that AeroSpace may behave poorly if there are unreasonably many windows opened (maybe 100 or even more, we didn't count)

I tried to open 20 windows and it performs well, I don't see any delay compared to when AeroSpace is not running

> Do other people experience this delay or am I missing some configuration

I can't see any delay on my machine (M1 Pro + Sonoma). I normally don't have more than 10 windows opened

> or some inherent limitation of Accessibility API?

I heard several times that Accessibility API is slow. Though, I can't confirm it yet. The only real slowness of Accessibility API I noticed is when some apps block their UI thread _in some special way_ (cause I couldn't reproduce the slowness when I wrote a small app that displays a simple window and blocks the UI thread).

For example, Spotify and Firefox sometimes block the UI thread _in a special way_ for several seconds when I try to quit (`cmd+q`) either of them, Accessibility API is blocked by these apps until they finally quit. I've not yet debugged this problem. 

**(3)** Some of the apps that you are running may have similar problems, so you can try to quit apps one by one to see when AeroSpace delay goes away

Kamyil:
Hi! I'm experiencing this issue on my Macbook Pro 14 with M1 Pro on macOS Sonoma, but a little bit differently.
[...]

Issue #16, opened 2023-11-12
Support more CLI commands


nikitabobko:
Reference: https://man.archlinux.org/man/extra/i3-wm/i3-msg.1.en

```
list-workspaces
list-windows [--all-workspaces]
list-apps
list-displays
tree [--json] [--include-floating-windows]
focus window-id 312
```

wojciech-kulik:
Great work! But why I don't see `list-workspaces` when I'm using aerospace via homebrew?

MatthiasGrandl:
Not released yet. You have to build it yourself.

alper:
Can we get a release? This would be great.

wojciech-kulik:
@nikitabobko why releases are not done regularly? The last release is one month old.

nikitabobko:
First versions of`list-workspaces` and `list-windows` are now released in 0.8.0-Beta

nikitabobko:
Would be nice to make `tree` command "interactive" #236

nikitabobko:
It's not a replacement for `tree` command, but in 0.12.0-Beta, `list-windows` comes with the `--format` flag that can be used to print to windows with their workspaces

TJTorola:
Would love this tree command. I want to take a stab at implimenting https://github.com/AdamWagner/stackline for aerospace but I don't think I can without being able to introspect on the tree structure first with a command like this.

jrk:
I was curious before digging into this myself:

Is this blocked on #1215 or is there something else that makes this nontrivial?

nikitabobko:
> Is this blocked on https://github.com/nikitabobko/AeroSpace/issues/1215 or is there something else that makes this nontrivial?

Not sure if it's blocked by https://github.com/nikitabobko/AeroSpace/issues/1215 but they indeed related, and I'd prefer to fix https://github.com/nikitabobko/AeroSpace/issues/1215 first

But the primary open question is the exact synopsis of the `tree` command and the examples of the command output (e.g. see https://github.com/nikitabobko/AeroSpace/issues/1403 to understand how I expect it to be formulated)

pfilipp:
I had a go at it in my fork. 

## Proposed synopsis

```
aerospace tree [-h|--help] [--json] [--workspace <workspace>] [--focused]
               [--floating-windows-as-tiling-windows]
```

## Options

| Option | Description |
|--------|-------------|
| `-h`, `--help` | Print help |
| `--json` | Output as JSON instead of human-readable tree |
[...]

PR #1191, opened 2025-03-07
Add more query data types

xiamaz:
Currently querying aerospace is rather slow (approx 50ms per call on my machine). I believe this is mostly caused by the usage of unix domain sockets for IPC, which seems to be much slower than mach ports, which is used by sketchybar (approx 1ms per call).

A common use-case, where this is noticeable is for updating bars and there has been some complex workarounds. Most of these can be removed by just creating a single call that contains all necessary information.

This modification allows us to get a list of workspaces including visibility and focus information, as well as names of apps. This is composable, so now e.g. window and app variables can also be used for workspace queries.

This can e.g. be used in the following command: `./.debug/aerospace list-workspaces --all --format '%{workspace-focused}%{workspace-visible}%{app-name}' --json`

Which will return e.g. the following json response now:

```
...
  {
    "app-name" : [
      "Mail",
      "iTerm2"
    ],
    "workspace-visible" : true,
    "workspace-focused" : true
  },
...
```

This allows for statusbar updates without noticeable latency without any additional caching or variables.

@nikitabobko Please let me know if you are interested in potentially merging this. I can make this a bit more complete and add some testing and documentation.

Related issues:
https://github.com/nikitabobko/AeroSpace/issues/104

These changes are used in the following sketchybar plugin: https://github.com/xiamaz/sketchybar-aerospace-plugin

### DIFF
--- a/Sources/AppBundle/command/format.swift
+++ b/Sources/AppBundle/command/format.swift
@@ -1,4 +1,5 @@
 import Common
+import Foundation
 
 enum AeroObj {
     case window(Window)
@@ -82,6 +83,8 @@ enum FormatVar: Equatable {
 
     enum WorkspaceFormatVar: String, Equatable, CaseIterable {
         case workspaceName = "workspace"
+        case workspaceFocused = "workspace-focused"
+        case workspaceVisible = "workspace-visible"
     }
 
     enum AppFormatVar: String, Equatable, CaseIterable {
@@ -105,6 +108,7 @@ enum Primitive: Encodable {
     case int32(Int32)
     case uint32(UInt32)
     case string(String)
+    indirect case array([Primitive])
 
     func toString() -> String {
         switch self {
@@ -113,6 +117,7 @@ enum Primitive: Encodable {
             case .int32(let x): x.description
             case .uint32(let x): x.description
             case .string(let x): x
+            case .array(let x): x.map{$0.toString()}.joined(separator: ",")
         }
     }
 
@@ -123,6 +128,7 @@ enum Primitive: Encodable {
             case .int32(let x): x
             case .uint32(let x): x
             case .string(let x): x
+            case .array(let x): x
         }
         var container = encoder.singleValueContainer()
         try container.encode(value)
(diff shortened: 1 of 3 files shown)


Sources/AppBundle/command/format.swift
```diff
@@ -169,6 +174,22 @@ extension String {
 
             case (.workspace(let ws), .monitor):
                 return expandFormatVar(obj: AeroObj.monitor(ws.workspaceMonitor))
+            case (.workspace(let ws), .app):
```
nikitabobko, line 177: I'm ok to merge `workspace-focused` and `workspace-visible` (though, I'd rename them to `workspace-is-focused` and `workspace-is-visible`)

I won't merge the support for `list-workspaces --all --format '%{app-name}'`. It's an incorrect approach to traverse children of the tree in `--format` flags, it's only allowed to traverse parents in `--format` flags. Otherwise, the output is not structural. E.g. consider `list-workspaces --all --format '%{app-name}%{app-bundle-id}'`

For printing the structure of the tree, a separate command should be designed and implemented https://github.com/nikitabobko/AeroSpace/issues/16

Performance related problems will be addressed by:
- `eval` command which will allow to run several commands in one request https://github.com/nikitabobko/AeroSpace/issues/278
- https://github.com/nikitabobko/AeroSpace/issues/104 & https://github.com/nikitabobko/AeroSpace/issues/131 with the approach of "thread per application" which will avoid blockage of the server main thread. I'm slowly working on that issue

nikitabobko (changes requested):

Sources/AppBundle/command/format.swift
xiamaz: Sure, thanks for the thorough response and I can see the issues with including the app and window fields in the workspace command.

I can take a look at the tree command and possibly create something along the line.

xiamaz:
@nikitabobko I have updated the docs and addressed your comments.

nikitabobko:
Thanks, merged

PR #1223, opened 2025-03-19
Add key-mapping.mod-notation-to-mod-flags config option

kaldius:
Hi, first off, I just want to say that AeroSpace is really awesome, so thanks for creating and maintaining this project! In particular, I think virtual workspaces was a great decision, and one of the main factors for me to switch over from yabai. I would also like to acknowledge the excellently organized codebase, helper scripts and docs, as they really make development such a smooth experience.

This is my first PR for an open-source project, ever, so please let me know if I have overlooked any formalities/best practicees, I'd be glad to make any necessary changes.

### About this PR

This PR addresses the introduction of custom modifier keys discussed in #224

Similar to `key-mapping.key-notation-to-key-code`, the new `key-mapping.mod-notation-to-mod-flags` alias allows any arbitrary modifier notation to represent one or more modifier keys.

This implementation differs slightly from the suggestion in https://github.com/nikitabobko/AeroSpace/issues/224#issuecomment-2111556500 due to how keys and mods are fundamentally treated differently, as well as the use of different types to represent each of them (i.e. `Key` and `NSEvent.ModifierFlags`). Because of this, I found it suitable to introduce a new TOML key for this functionality. 

This implementation also addresses the main concern raised in https://github.com/nikitabobko/AeroSpace/pull/389#issuecomment-2298994576 -that hardcoding certain combinations of modifiers is not extensible.

Looking forward to your feedback!

### DIFF
--- a/Sources/AppBundle/config/HotkeyBinding.swift
+++ b/Sources/AppBundle/config/HotkeyBinding.swift
@@ -70,7 +70,7 @@ struct HotkeyBinding: Equatable, Sendable {
     }
 }
 
-func parseBindings(_ raw: TOMLValueConvertible, _ backtrace: TomlBacktrace, _ errors: inout [TomlParseError], _ mapping: [String: Key]) -> [String: HotkeyBinding] {
+func parseBindings(_ raw: TOMLValueConvertible, _ backtrace: TomlBacktrace, _ errors: inout [TomlParseError], _ keyMapping: KeyMapping) -> [String: HotkeyBinding] {
     guard let rawTable = raw.table else {
         errors += [expectedActualTypeError(expected: .table, actual: raw.type, backtrace)]
         return [:]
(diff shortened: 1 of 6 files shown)


nikitabobko:
Thanks, but I vote against adding a new `key-mapping.mod-notation-to-mod-flags` section for such a minor feature. Yes, right now different types are used, it's part of the task to unify them.

nikitabobko:
Thanks for writing down your reasoning

> Unifying key codes and modifier flags under key-mapping.key-notation-to-key-code could lead to unintended expectations. A user might attempt to combine them like this

Frankly speaking, I don't think that it will cause any confusions. Combining mods and keys is slightly harder to program - true, but it it's not confusing.

```toml
[key-mapping.key-notation-to-key-code]
  foo = 'k' # when only keys are used, the notation must be used only as a suffix
  hyper = 'ctrl-alt-cmd-shift' # when only mods are used, the notation must be used only as a prefix
  bar = 'ctrl-a' # when flags and keys are combined, the notation must be used only as a suffix

[mode.main.binding]
  alt-foo = '' # green
  foo-k = '' # red. 'foo' must be a suffix

  alt-bar = '' # green
  bar-k = '' # red. 'bar' must be a suffix

  hyper-a = '' # green
  alt-hyper-k = '' # red. 'hyper' must be a prefix
  alt-hyper = '' # red. 'huper' must be a prefix
```

Nevertheless, I'm ok with your suggestion to allow only mods or keys in right-hand side. It's a reasonable limitation. And it's better to limit something than giving something that will be hard to deprecate.

> if using key-notation-to-key-code is still your preference

Yes, I think, it's still a preference

> One last point—since this is indeed a minor feature, perhaps the earlier solution of hard-coding all combinations wouldn’t be a bad compromise.

How would you call these combinations? `ctrlaltshift`? It is not better than writing `ctrl-alt-shift`.

kaldius:
@nikitabobko Thanks for your input. I understand your stance better now.

> I'm ok with your suggestion to allow only mods or keys in right-hand side.

Given the size of this feature, I think I'll go ahead with this. I now agree with you there is not much confusion, but the added complexity of checking prefix/suffix is probably not worth it.

> ctrlaltshift? It is not better than writing ctrl-alt-shift.

Good point. Hadn't thought about that. 🫣

I'll close this PR for now and raise another one soon.

Issue #60, opened 2023-12-17
[Feature Request] Large Monitor, Single App Window Size Concern


juniusfree:
First of all, thank you for creating AeroSpace.

Apologies if this is already solved. I am using a 28-inch monitor. When only one application is open, the window stretches across the entire screen, making it too wide and challenging to navigate efficiently.

It might be helpful to have a feature that allows for managing the window size or position in such scenarios.

This is the result I am aiming for. The window could be also be on the left or right, not just center.
![image](https://github.com/nikitabobko/AeroSpace/assets/8164667/e25a6f64-ec7b-40af-b6e3-7a47237e779d)

Marty-W:
Also endorsing this one. In yabai I am using space specific padding to achieve this, so maybe that could be a possible solution without restricting window sizes etc?

nikitabobko:
> In yabai I am using space specific padding to achieve this

What is the point of using ultra-wide screen, if you effectively cut left and right gaps from it?

I'm not joking, it's real question to better understand use case for workspace-specific gaps

Marty-W:
My current setup on an ultrawide: 

```
1 work: docs/terminal | code | localhost
2 leisure: (gap) | browser | (gap)
3 focused work: (gap) | browser | code | (gap)
```

The thing is I am still trying to divide the workspaces for the task at hand. So when I'm just browsing the web I don't really need all that screen real estate and having one window span the entire screen makes it harder to navigate (as @juniusfree  pointed out). The idea behind workspace three is that sometimes having too many windows open makes it harder for me to focus, so having just two, or only an editor makes it easier.

nikitabobko:
@Marty-W thanks for the explanation, I get it now

When you mentioned workspace-specific gaps, I assumed that you'd just make all workspaces that are assigned to the ultrawide monitor have big gaps, but I get it now. You only put big gaps for workspaces that only have one or so windows

That's an interesting use case for workspace-specific gaps

nikitabobko:
Currently, it's possible to have monitor-specific gaps
```
gaps.outer.top = [{ monitor."built-in" = 0 }, 24] # (1)
```

That's how it can be extended to support workspace-specific gaps:
```
gaps.outer.top = [{ monitor."built-in" = 0 }, { workspace.1 = 10 }, 24] # (2)
```

But I think we can do even better to support a pattern like "window count" (I'm brainstorming, it's not a final design, though I encourage you to share your thoughts)
```
# (3)
[...]

PR #1229, opened 2025-03-22
Implement smart gaps like i3-gaps

ntrp:
I did a simple implementation for smart gaps that behaves more or less as i3-gaps.

I tested the various scenarios:
- property missing: standard behaviour, default false
- smart = true: if only one tiled window is present remove outer gaps, once more than one are there add outer gaps

let me know if some tests are needed

### DIFF
--- a/Sources/AppBundle/config/parseGaps.swift
+++ b/Sources/AppBundle/config/parseGaps.swift
@@ -2,10 +2,11 @@ import Common
 import TOMLKit
 
 struct Gaps: Copyable, Equatable, Sendable {
+    var smart: Bool
     var inner: Inner
     var outer: Outer
 
-    static let zero = Gaps(inner: .zero, outer: .zero)
+    static let zero = Gaps(smart: false, inner: .zero, outer: .zero)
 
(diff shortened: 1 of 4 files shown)


nikitabobko:
https://github.com/nikitabobko/AeroSpace/discussions/1009#discussioncomment-11877426

ntrp:
Oh I missed that one. What is the state of the discussion regarding the conditional gaps? I would be also fine using that approach. 
If we are going in that direction I should close this one then?

nikitabobko:
Do you mean this proposal - https://github.com/nikitabobko/AeroSpace/issues/60#issuecomment-1933133848 ?

I think it's mostly rejected. The initial use-case it was suggested for (better support for ultra-wide monitors) is better covered by `max-width` config option

[My stance on smart gaps](https://github.com/nikitabobko/AeroSpace/discussions/1009#discussioncomment-11877426) hasn't changed - I don't understand their purpose. You can try to change my mind in the thread.

I'm fine if it becomes possible to express smart gaps in the future as a side-effect of a more abstract feature (like the suggested conditional gaps).

I'm closing the PR for now

PR #1243, opened 2025-03-28
Implement focus dfs-next/dfs-prev.

rickyz:
This allows switching to the next/prev window from the current one in
the depth-first order of the windows in the current workspace tree.

This is convenient, as it allows users to shift focus through a
workspace's windows in a more predictable way than the directional
movement commands, where the move target can depend on invisible
most-recently-used state.

_fixes https://github.com/nikitabobko/AeroSpace/issues/248

### DIFF
--- a/Sources/AppBundle/command/impl/FocusCommand.swift
+++ b/Sources/AppBundle/command/impl/FocusCommand.swift
@@ -162,7 +162,7 @@ private struct FloatingWindowData {
 
 extension TreeNode {
     @MainActor
-    fileprivate func findFocusTargetRecursive(snappedTo direction: CardinalDirection) -> Window? {
+    func findFocusTargetRecursive(snappedTo direction: CardinalDirection) -> Window? {
         switch nodeCases {
             case .workspace(let workspace):
                 return workspace.rootTilingContainer.findFocusTargetRecursive(snappedTo: direction)
--- a/Sources/AppBundle/command/impl/FocusCommand.swift
+++ b/Sources/AppBundle/command/impl/FocusCommand.swift
@@ -36,6 +36,24 @@ struct FocusCommand: Command {
                 } else {
                     return io.err("Can't find window with DFS index \(dfsIndex)")
                 }
+            case .dfsRelative(let nextPrev):
+                let windows = target.workspace.rootTilingContainer.allLeafWindowsRecursive
+                guard let currentIndex = windows.firstIndex(where: { $0 == target.windowOrNil }) else {
+                    return false
+                }
+                var targetIndex = switch nextPrev {
+                    case .dfsNext: currentIndex + 1
+                    case .dfsPrev: currentIndex - 1
+                }
+                if targetIndex < 0 || targetIndex >= windows.count {
+                    switch args.boundariesAction {
+                        case .stop: return true
+                        case .fail: return false
+                        case .wrapAroundTheWorkspace: targetIndex = (targetIndex + windows.count) % windows.count
+                        case .wrapAroundAllMonitors: return dieT("Must be discarded by args parser")
+                    }
+                }
+                return windows[targetIndex].focusWindow()
         }
     }
 }
--- a/Sources/AppBundle/command/impl/SwapCommand.swift
+++ b/Sources/AppBundle/command/impl/SwapCommand.swift
@@ -1,0 +1,55 @@
+import AppKit
+import Common
+
+struct SwapCommand: Command {
+    let args: SwapCmdArgs
+
+    func run(_ env: CmdEnv, _ io: CmdIo) async throws -> Bool {
+        guard let target = args.resolveTargetOrReportError(env, io) else {
+            return false
+        }
+
+        guard let currentWindow = target.windowOrNil else {
+            return io.err(noWindowIsFocused)
+        }
+
+        let targetWindow: Window?
+        switch args.target.val {
+            case .direction(let direction):
+                if let (parent, ownIndex) = currentWindow.closestParent(hasChildrenInDirection: direction, withLayout: nil) {
+                    targetWindow = parent.children[ownIndex + direction.focusOffset].findFocusTargetRecursive(snappedTo: direction.opposite)
+                } else if args.wrapAround {
+                    targetWindow = target.workspace.findFocusTargetRecursive(snappedTo: direction.opposite)
+                } else {
+                    return false
+                }
+            case .dfsRelative(let nextPrev):
+                let windows = target.workspace.rootTilingContainer.allLeafWindowsRecursive
+                guard let currentIndex = windows.firstIndex(where: { $0 == target.windowOrNil }) else {
+                    return false
+                }
+                var targetIndex = switch nextPrev {
+                    case .dfsNext: currentIndex + 1
+                    case .dfsPrev: currentIndex - 1
+                }
+                if targetIndex < 0 || targetIndex >= windows.count {
+                    if !args.wrapAround {
+                        return false
+                    }
+                    targetIndex = (targetIndex + windows.count) % windows.count
+                }
+                targetWindow = windows[targetIndex]
+        }
+
+        guard let targetWindow else {
+            return false
+        }
+
+        swapWindows(currentWindow, targetWindow)
+
+        if args.swapFocus {
+            return targetWindow.focusWindow()
+        }
+        return currentWindow.focusWindow()
+    }
+}
--- a/Sources/Common/cmdArgs/impl/FocusCmdArgs.swift
+++ b/Sources/Common/cmdArgs/impl/FocusCmdArgs.swift
@@ -12,20 +12,20 @@ public struct FocusCmdArgs: CmdArgs {
             "--window-id": ArgParser(\.windowId, upcastArgParserFun(parseArgWithUInt32)),
             "--dfs-index": ArgParser(\.dfsIndex, upcastArgParserFun(parseArgWithUInt32)),
         ],
-        arguments: [ArgParser(\.direction, upcastArgParserFun(parseCardinalDirectionArg))],
+        arguments: [ArgParser(\.cardinalOrDfsDirection, upcastArgParserFun(parseCardinalOrDfsDirection))],
     )
 
     public var rawBoundaries: Boundaries? = nil // todo cover boundaries wrapping with tests
     public var rawBoundariesAction: WhenBoundariesCrossed? = nil
     public var dfsIndex: UInt32? = nil
-    public var direction: CardinalDirection? = nil
+    public var cardinalOrDfsDirection: CardinalOrDfsDirection? = nil
     public var floatingAsTiling: Bool = true
     /*conforms*/ public var windowId: UInt32?
     /*conforms*/ public var workspaceName: WorkspaceName?
 
-    public init(rawArgs: [String], direction: CardinalDirection) {
+    public init(rawArgs: [String], cardinalOrDfsDirection: CardinalOrDfsDirection) {
         self.rawArgs = .init(rawArgs)
-        self.direction = direction
+        self.cardinalOrDfsDirection = cardinalOrDfsDirection
     }
 
     public init(rawArgs: [String], windowId: UInt32) {
@@ -54,12 +54,24 @@ public enum FocusCmdTarget {
     case direction(CardinalDirection)
     case windowId(UInt32)
     case dfsIndex(UInt32)
+    case dfsRelative(DfsNextPrev)
+
+    var isDfsRelative: Bool {
+        if case .dfsRelative = self {
+            return true
+        } else {
+            return false
+        }
+    }
 }
 
 extension FocusCmdArgs {
     public var target: FocusCmdTarget {
-        if let direction {
-            return .direction(direction)
+        if let cardinalOrDfsDirection {
+            return switch cardinalOrDfsDirection {
+                case .direction(let dir): .direction(dir)
+                case .dfsRelative(let nextPrev): .dfsRelative(nextPrev)
+            }
         }
         if let windowId {
             return .windowId(windowId)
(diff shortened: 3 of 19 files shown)


rickyz:
I added an implementation of swap as well (which depends on part of the implementation of `focus (dfs-next|dfs-prev)`, so it was most convenient to include in this PR as well. Happy to split that out if preferred though!

timobenn:
Somewhat unfortunate/amusing timing, I had been poking at this same feature (and learning Swift), though I don't consider mine ready for a PR yet: https://github.com/nikitabobko/AeroSpace/compare/main...timobenn:AeroSpace:focus-dfs-next-prev-248

A couple of things to note:

* neither your PR nor my branch (yet) support floating windows; they're not included in the list of leaf windows. I was going to see today if simply appending the list of floating windows to the list of leaf windows (unless `--ignore-floating` is set would feel natural in use.
* my branch supports both `workspace` and `all-monitors-outer-frame` boundary options -- feel free to steal that

nikitabobko:
> neither your PR nor my branch (yet) support floating windows

@timobenn wdym? This PR supports focusing floating windows via `dfs-next|dfs-prev`

timobenn:
@nikitabobko when I'd tested mine last week, I came to the conclusion that `target.workspace.rootTilingContainer.allLeafWindowsRecursive` does not include floating windows, so any index operations on that use just that list would skip floating windows (thus I added [a note](https://github.com/nikitabobko/AeroSpace/commit/66efe21a2bf01f0ddd0e9bdf325073322f133119)). It's possible I misinterpreted my testing, but it looks like this PR would run into the same issue I was, where `focus dfs-next/dfs-prev` would just skip floating windows. At least, I don't see how it would include them.

Sources/Common/cmdArgs/impl/FocusCmdArgs.swift
```diff
@@ -53,16 +53,46 @@ public struct FocusCmdArgs: CmdArgs {
...
+
+func parseFocusTargetArg(_ arg: String, _ nextArgs: inout [String]) -> Parsed<FocusTargetArg> {
+    return switch arg {
+        case "left": .success(.direction(.left))
+        case "down": .success(.direction(.down))
+        case "up": .success(.direction(.up))
+        case "right": .success(.direction(.right))
+        case "dfs-next": .success(.dfsRelative(.next))
+        case "dfs-prev": .success(.dfsRelative(.prev))
+        default: .failure("Can't parse '\(arg)\'. Possible values: left|down|up|right|dfs-next|dfs-prev")
```
nikitabobko, line 70: I don't like the hardcoded `left|down|up|right|dfs-next|dfs-prev`. I'd prefer to make use of `CaseIterable` of every nested `case` and combine the result. Or maybe it's even possible to write a generic functions which trverses `CaseIterable`

nikitabobko: The same for parsing the argument itself. I don't like the hardcoded `case "left"`, `case "down"`, `case "dfs-next"`

docs/aerospace-focus.adoc
```diff
@@ -32,6 +35,15 @@ This behavior can be disabled with `--ignore-floating` flag.
...
 https://github.com/nikitabobko/AeroSpace/issues/5
 
+*1. (left|down|up|right) arguments*
+
+{manpurpose}
+
+*2. (dfs-next|dfs-prev) arguments*
+
+Set focus to the window before or after the current window in the depth-first order (top-to-bottom and left-to-right) of windows in the current workspace tree.
+In this mode, `--boundaries` must be `workspace` (the default) and `--boundaries-action` can be set to one of `(stop|fail|wrap-around-the-workspace)`.
```
nikitabobko, line 45: I know that docs for some of other commands written in this style, but please prefer the style with `ARGUMENTS` section like in `aerospace-focus-monitor`

Other commands that use this style of documentation (e.g. `workspace`) will be migrated to `ARGUMENTS`-section-style in the future

docs/aerospace-focus.adoc
```diff
@@ -12,6 +12,9 @@ include::util/man-attributes.adoc[]
 aerospace focus [-h|--help] [--ignore-floating]
                 [--boundaries <boundary>] [--boundaries-action <action>]
                 (left|down|up|right)
+aerospace focus [-h|--help] [--ignore-floating]
```
nikitabobko, line 15: `--ignore-floating` is also supported here

grammar/commands-bnf-grammar.txt
```diff
@@ -111,8 +112,10 @@ aerospace -h;
 <move_node_to_monitor1_flag> ::= --window-id <window_id>|--focus-follows-window|--fail-if-noop|--wrap-around;
 <move_node_to_monitor2_flag> ::= --window-id <window_id>|--focus-follows-window|--fail-if-noop;
 
-<focus_flag> ::= --boundaries <boundary>|--boundaries-action <boundaries_action>|--ignore-floating;
-<boundaries_action> ::= stop|fail|wrap-around-the-workspace|wrap-around-all-monitors;
+<focus_dir_flag> ::= --boundaries <boundary>|--boundaries-action <dir_boundaries_action>|--ignore-floating;
+<focus_dfs_relative_flag> ::= --boundaries-action <dfs_relative_boundaries_action>|--ignore-floating;
+<dir_boundaries_action> ::= stop|fail|wrap-around-the-workspace|wrap-around-all-monitors;
+<dfs_relative_boundaries_action> ::= stop|fail|wrap-around-the-workspace;
```
nikitabobko, line 120: I'd prefer the shell completion to suggest yet unsupported `wrap-around-all-monitors` rather than complicating the grammar

grammar/commands-bnf-grammar.txt
```diff
@@ -20,9 +20,10 @@ aerospace -h;
 
     | flatten-workspace-tree [--workspace <workspace>]
 
-    | focus [<focus_flag>]... (left|down|up|right) [<focus_flag>]...
+    | focus [<focus_dir_flag>]... (left|down|up|right) [<focus_dir_flag>]...
```
nikitabobko, line 23: IMO "dir" = "direction" is not obvious shortening

nikitabobko (changes requested):
I didn't read all of the code yet, here the first portion of comments

Overall - looks good, thanks

I don't care if the commits are in one big PR or two separate PRs. I review commit by commit anyway.

If your second commit depends on the first one, I understand why you put them into a single PR. Otherwise, you might yourself prefer to split it into two different PRs, because the code review will take longer with several commits in one PR

nikitabobko:
@timobenn since `focus` command is special in that regard, it has a small "hack" inside of it https://github.com/nikitabobko/AeroSpace/blob/fb3a863dd201ef69953d02eb989ed3dd0fed1946/Sources/AppBundle/command/impl/FocusCommand.swift#L9-L15

timobenn:
@nikitabobko I must have been running into a weird issue with window state when first launching the debug build on top of a running instance; when I test today I'm able to get both my branch and this PR to acknowledge floating windows correctly, as long as I make sure to fully reset the layout after launch first, before floating any windows.

timobenn:
If you all have interest in having dfs-next/prev fully function with `all-monitors-outer-frame` & `wrap-around-all-monitors`, perhaps I can put up a smaller PR after this is merged.

Sources/Common/cmdArgs/impl/FocusCmdArgs.swift
rickyz: Renamed FocusTargetArg to CardinalOrDfsDirection and implemented the CaseIterable and RawRepresentable protocols for it so that parseEnum and unionLiteral can be used.

docs/aerospace-focus.adoc
rickyz: Thanks for the example, switched to `ARGUMENTS`-section-style.

grammar/commands-bnf-grammar.txt
rickyz: Done - I had left it out because I wasn't sure that `wrap-around-all-monitors` made sense for dfs-next/dfs-prev. Within a workspace, the ordering of DFS indices are well defined, but it wasn't clear to me that there was a clear correct next monitor to wrap around to.

grammar/commands-bnf-grammar.txt
rickyz: Renamed to focus_direction_flag

Sources/AppBundle/command/impl/SwapCommand.swift
```diff
@@ -0,0 +1,51 @@
+import AppKit
+import Common
+
+struct SwapCommand: Command {
+    let args: SwapCmdArgs
+
+    func run(_ env: CmdEnv, _ io: CmdIo) -> Bool {
+        guard let target = args.resolveTargetOrReportError(env, io), let currentWindow = target.windowOrNil else {
```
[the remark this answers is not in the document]
rickyz: Done (though now that I see that target could be overridden by `resolveTargetOrReportError`, is it even correct to emit "noWindowIsFocused" in the second guard?

docs/aerospace-swap.adoc
```diff
@@ -0,0 +1,45 @@
...
+:manname: aerospace-swap
+// tag::purpose[]
+:manpurpose: Swaps the focused window with the nearest window in the given direction.
+// end::purpose[]
+
+// =========================================================== Synopsis
+== Synopsis
+[verse]
+// tag::synopsis[]
+aerospace swap [-h|--help] [--move-focus-to-target] [--wrap-around]
```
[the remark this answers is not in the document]
rickyz: Oops, fixed

rickyz:
Thanks for the review!

@timobenn Oops, sorry for the unintended duplication of work - I took your nicer naming for the focus argument, but didn't pull in the monitor wrapping code - feel free to follow up with that if folks are able to agree on what that behavior would be!

fwiw, I didn't attempt to implement `wrap-around-all-monitors` because I wasn't sure that there was a well-defined next monitor to wrap to with dfs-next/dfs-prev. With the cardinal directions, it makes sense to wrap to the opposite direction edge, but couldn't think of something analogous for monitors (especially since the prev focused window could be adjacent to multiple monitors).

timobenn:
@rickyz 

> @timobenn Oops, sorry for the unintended duplication of work - I took your nicer naming for the focus argument, but didn't pull in the monitor wrapping code - feel free to follow up with that if folks are able to agree on what that behavior would be!

No worries! Risk I took by working so slowly without broadcasting I was doing it.

For wrapping around all monitors I was using this logic:

* when it's time to pick a different monitor, use the same logic as is used by `focus-monitor [--wrap-around] (next|prev)`, which goes to the "next" or "prev" monitor based on the same order they are shown in the menu bar
* if `dfs-next`, go to first DFS window it the new monitor, else the last DFS window

For the limited amount of manual testing I'd done so far, that logic was feeling natural.

rickyz:
Ah, I see - I wasn't aware that montiors already had some fixed ordering to them.

Re the earlier discussion about floating windows - it looks like the current trick for dealing with floating windows inserts them into the window tree at a location depending on its location on the screen. As a result, they are assigned an unintuitive DFS position when using dfs-next and dfs-prev - I think I would actually prefer we explicitly sorted floating windows after all of the windows in the tree instead - will update the implementation to do that if there are no objections.

timobenn:
I do agree that floating windows being mixed among the leaves the way they currently are feels a bit weird in the context of dfs-next/prev. I'm not sure how much better grouping them together would feel (though I expect it would be at least a little), or how much complexity would be added by handling floating windows  differently based on the focus target.

Sources/AppBundle/command/impl/SwapCommand.swift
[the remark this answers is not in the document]
nikitabobko: > is it even correct to emit "noWindowIsFocused" in the second guard?

Probably, it'd be better to replace all "No window is focused" messages with smth like "Workspace X is empty"

grammar/commands-bnf-grammar.txt
nikitabobko: > I wasn't sure that `wrap-around-all-monitors` made sense for `dfs-next`/`dfs-prev`

I'm not sure myself. It requires more thinking. The way monitors are currently represented in AeroSpace, `dfs-next`/`dfs-prev` rather doesn't make sense for monitors

nikitabobko (approved):
Thanks, merged. Sorry, that it took me that long to get back to the PR.

I also must say that your last few PRs were one of the highest quality PRs among all the PRs I received.

I'd be happy to see you contribution in the future, shall you decide to submit more PRs, given how slow I might be

Sources/AppBundle/command/impl/FocusCommand.swift
```diff
@@ -36,6 +36,24 @@ struct FocusCommand: Command {
...
+            case .dfsRelative(let nextPrev):
+                let windows = target.workspace.rootTilingContainer.allLeafWindowsRecursive
+                guard let currentIndex = windows.firstIndex(where: { $0 == target.windowOrNil }) else {
+                    return false
+                }
+                var targetIndex = switch nextPrev {
+                    case .dfsNext: currentIndex + 1
+                    case .dfsPrev: currentIndex - 1
+                }
+                if targetIndex < 0 || targetIndex >= windows.count {
```
nikitabobko, line 48: Small tip:
```
if !(0 ..< windows.count).contains(targetIndex) {
```

rickyz:
> Thanks, merged. Sorry, that it took me that long to get back to the PR.
> 
> I also must say that your last few PRs were one of the highest quality PRs among all the PRs I received.
> 
> I'd be happy to see you contribution in the future, shall you decide to submit more PRs, given how slow I might be

Thanks for the kind words, and for making Aerospace! Will keep trying to send patches for issues I encounter, but I'm quickly running out of major annoyances thanks to all the optimizations and bugfixes you've been making!

rcorre:
Fantasitc! I've now got this, which seems to mirror the behavior I'm used to from most other tiling WMs (dwm/river/hyprland/etc.), and works whether you're vsplit/hsplit/maximized. Thanks to everyone who contributed!

```toml
cmd-j = 'focus dfs-prev --boundaries-action wrap-around-the-workspace'
cmd-k = 'focus dfs-next --boundaries-action wrap-around-the-workspace'
```

Issue #1231, opened 2025-03-23
Binding `reload-config` in the `main` binding mode is unreliable


nikitabobko:
To reproduce, run AeroSpace with the following config snippet:
```
[mode.main.binding]
    # dialog command is to indicate whether the keybinding was triggered.
    alt-e = ['reload-config', '''exec-and-forget osascript -e 'display dialog "Test"' ''']
```

Pressing alt-e once triggers the keybind as expected. However, pressing alt-e a second time will no longer trigger the key binding.

I suspect the issue is that the [HotKey library](https://github.com/soffes/HotKey/) ties hotkey registration/deregistration to the lifetime of the HotKey object: https://github.com/soffes/HotKey/blob/a3cf605d7a96f6ff50e04fcb6dea6e2613cfcbe4/Sources/HotKey/HotKey.swift#L30-L50

Hotkey handlers are called with the `HotKey` object still in scope:
https://github.com/soffes/HotKey/blob/a3cf605d7a96f6ff50e04fcb6dea6e2613cfcbe4/Sources/HotKey/HotKeysController.swift#L127-L133

When a `ReloadConfigCommand` triggered by a hotkey resets and reinstalls hotkeys, the `HotKey` for the reload is still alive. I suspect this causes attempts to register the same hotkey in the reloaded config to fail silently at:
https://github.com/soffes/HotKey/blob/a3cf605d7a96f6ff50e04fcb6dea6e2613cfcbe4/Sources/HotKey/HotKeysController.swift#L69-L72

It's pretty easy to work around this in practice by binding `exec-and-forget aerospace reload-config` instead of `reload-config`, but it would be nice if this sharp edge didn't exist.

Perhaps a fix could be to change `ReloadConfigCommand` to dispatch a separate callback to the main thread so that the config reload no longer runs from inside a hotkey handler. Alternately, `resetHotKeys` could unregister existing hotkeys by setting `isPaused = true` on every hotkey before clearing `hotkeys`.

btw, thank you so much for building AeroSpace - I have long wished for a macOS window manager that implements its own workspaces instead of building on the not-great Spaces model/APIs!

_Originally posted by @rickyz in https://github.com/nikitabobko/AeroSpace/discussions/1230_

nikitabobko:
Fixed in 0.18.0

PR #1248, opened 2025-03-30
Explicitly unregister hotkeys when resetting hotkeys

rickyz:
We cannot always rely on destruction of the HotKey object to trigger
unregistration because we might be running inside a hotkey handler that
is keeping its HotKey object alive.

_fixes https://github.com/nikitabobko/AeroSpace/issues/1231


nikitabobko:
Thanks! I've checked that it fixes the issue. Merged

Issue #1216, opened 2025-03-16
Windows may randomly jump to the focused workspace


nikitabobko:
There are two cases when it's possible:

**1.** When you unlock the screen. It's because the current algorithm of saving state before the screen is locked is incorrect. To fix it, we should make `TreeNode` immutable https://github.com/nikitabobko/AeroSpace/issues/1215, which will make it a breath to take snapshots of the tree on its every modification before windows start "dying out" when the screen is locked.

**2.** During high CPU usage / when apps are slow at responding AX requests. When AeroSpace checks for dead windows, it doesn't distinguish between long responses and `nil` response. Though it's a hypothesis, which should be investigated

Both of the issue are caused by https://github.com/nikitabobko/AeroSpace/issues/445 (since 0.16.0-Beta)

Action points: 

**(1)** Investigate 21336ad382539b35fdc94b4fbd55408e10b101f8

**(2)** Test case

<details>

```
commit a6b44c561e48191bf5ad2454ab4e414df2c4c733
Author:     Nikita Bobko <[git@bobko.xyz](mailto:git@bobko.xyz)>
AuthorDate: Thu Dec 5 22:24:21 2024 +0100
Commit:     Nikita Bobko <[git@bobko.xyz](mailto:git@bobko.xyz)>
CommitDate: Sun Feb 16 18:56:55 2025 +0100

    TODO. Improve reliability of windows position restore on screen unlock. FUCK THIS DOESN'T WORK

    test case:
        Given: window1 and window2, window1 is in the closedWindowCache cache, window2 isn't

        Steps:
        1. lock the screen
        2. window1 dies -> cache is not updated
        3. window2 dies -> cache is updated, but it won't include window1, since window1 is already dead :(
        4. unlock the screen
        5. Only window2 is restored

    The commit makes sure that the "given" is never the case

diff --git Sources/AppBundle/tree/MacWindow.swift Sources/AppBundle/tree/MacWindow.swift
index f2c2e685..27e68a97 100644
--- Sources/AppBundle/tree/MacWindow.swift
+++ Sources/AppBundle/tree/MacWindow.swift
@@ -38,6 +38,7 @@ final class MacWindow: Window, CustomStringConvertible {
                 allWindowsMap[id] = window
                 debugWindowsIfRecording(window)
                 if !restoreClosedWindowsCacheIfNeeded(newlyDetectedWindow: window) {
+                    if !isClosedWindowsCacheUseful() { resetClosedWindowsCache() }
[...]

Issue #1215, opened 2025-03-16
Big meta issue: rewrite core `TreeNode` mutable double-linked tree data structure to immutable single-linked persistent tree


nikitabobko:
`TreeNode` is a core data structure of AeroSpace and unfortunately it's mutable.

The proposal is to rewrite it to immutable single-linked persistent tree.

Motivation:
1. Stability. The current logic for binding two `TreeNode` is fragile and sometimes leads to hard to debug crashes: https://github.com/nikitabobko/AeroSpace/issues/706
2. Performance. Once we make AeroSpace multi-threaded to handle hanging apps (https://github.com/nikitabobko/AeroSpace/issues/131), instead of "drawing" the layout on main thread and sending requests to app-threads, it makes sense to pass the tree data structure to each app-thread so that it could draw its part of the tree, which effectively makes the drawing multi-threaded by number of apps. Multi-threading drawing will help to reduce threads context switching for cases when apps have several visible windows. Though the tree has to be traversed multiple times in multi-threaded drawing approach. Precise performance measurements needs to be done.
3. It makes it possible to take snapshots of the tree on its every modification before windows start "dying out" when the screen is locked. It should help to fix this issue: 
4. Features like `aerospace undo` become possible (though, it may be too far fetched, who would need such a feature?)

nikitabobko:
Note to myself: after this issue is fixed, please review all the usages of `die` and `dieT` functions to make sure that they are not abused

zhuoqun-chen:
Hi @nikitabobko, I'm wondering if #726 also related to this meta-issue, I sometimes encounter the same situation when I have many chrome windows and vscode windows.

nikitabobko:
@zhuoqun-chen no, they are not related. I've answered in #726

Davincible:
Thanks a lot for your hard work on Aerospace guys. Just moved over to a mac machine from Hyprland and some of these improvements will improve QOL a lot, looking forward to the release! (for me Ghostty tabs as windows is the biggest pain point)

Davincible:
https://github.com/nikitabobko/AeroSpace/compare/main...DanielLemky:AeroSpace:main

I just found this fork, seems to address Ghostty issues. Would this be a reasonable (temporary /permanent idk) fix to merge in?

Not sure how this ties into your master refactor plan, but the sound of ghostty working properly is music to my ears

zzJinux:
[...]

PR #1249, opened 2025-03-30
Coalesce idempotent refreshSession calls triggered by event handlers

rickyz:
On operations like workspace changes, AeroSpace receives a stampede of
accessibility events, each of which triggers a serial call to
`refreshSession()` on the main thread. Since these `refreshSession()` calls
are idempotent, we can coalesce these calls so that
multiple accessibility events can be handled by a single
`refreshSession()`.

With this change, operations like workspace switch go from
O(num_source_windows + num_dest_windows) to O(1) `refreshSession()`
calls (with similar improvements for operations like closing windows or
resizing windows).

### DIFF
--- a/Sources/AppBundle/layout/refresh.swift
+++ b/Sources/AppBundle/layout/refresh.swift
@@ -47,6 +47,25 @@ func refreshAndLayout(_ event: RefreshSessionEvent, screenIsDefinitelyUnlocked: 
     refreshSession(event, screenIsDefinitelyUnlocked: screenIsDefinitelyUnlocked, startup: startup, body: {})
 }
 
+@MainActor private var havePendingRefresh = false
+@MainActor private var pendingRefreshScreenIsDefinitelyUnlocked = false
+
+@MainActor
+func scheduleRefreshAndLayout(_ event: RefreshSessionEvent, screenIsDefinitelyUnlocked: Bool = false) {
+    if havePendingRefresh {
+        if screenIsDefinitelyUnlocked {
+            pendingRefreshScreenIsDefinitelyUnlocked = true
+        }
+        return
+    }
+    havePendingRefresh = true
+    pendingRefreshScreenIsDefinitelyUnlocked = screenIsDefinitelyUnlocked
+    DispatchQueue.main.async { @MainActor in
+        havePendingRefresh = false
+        refreshSession(event, screenIsDefinitelyUnlocked: pendingRefreshScreenIsDefinitelyUnlocked, startup: false, body: {})
+    }
+}
+
 @MainActor
 func refreshModel() {
     gc()
@@ -81,7 +100,7 @@ func refreshObs(_ obs: AXObserver, ax: AXUIElement, notif: CFString, data: Unsaf
     check(Thread.isMainThread)
     let notif = notif as String
     MainActor.assumeIsolated {
-        refreshAndLayout(.ax(notif), screenIsDefinitelyUnlocked: false)
+        scheduleRefreshAndLayout(.ax(notif), screenIsDefinitelyUnlocked: false)
     }
 }
 
--- a/Sources/AppBundle/GlobalObserver.swift
+++ b/Sources/AppBundle/GlobalObserver.swift
@@ -11,7 +11,7 @@ class GlobalObserver {
         }
         let notifName = notification.name.rawValue
         MainActor.assumeIsolated {
-            refreshAndLayout(.globalObserver(notifName), screenIsDefinitelyUnlocked: false)
+            scheduleRefreshAndLayout(.globalObserver(notifName), screenIsDefinitelyUnlocked: false)
         }
     }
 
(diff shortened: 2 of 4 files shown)


Sources/AppBundle/layout/refresh.swift
```diff
@@ -47,6 +47,25 @@ func refreshAndLayout(_ event: RefreshSessionEvent, screenIsDefinitelyUnlocked:
...
+@MainActor private var havePendingRefresh = false
+@MainActor private var pendingRefreshScreenIsDefinitelyUnlocked = false
+
+@MainActor
+func scheduleRefreshAndLayout(_ event: RefreshSessionEvent, screenIsDefinitelyUnlocked: Bool = false) {
+    if havePendingRefresh {
+        if screenIsDefinitelyUnlocked {
+            pendingRefreshScreenIsDefinitelyUnlocked = true
+        }
+        return
```
nikitabobko, line 59: This `if` proves that it's not 100% idempotent :/

`screenIsDefinitelyUnlocked` is definitely a hack anyway that I hope to get rid of in scope https://github.com/nikitabobko/AeroSpace/issues/1215 + https://github.com/nikitabobko/AeroSpace/issues/1216

Sources/AppBundle/layout/refresh.swift
```diff
@@ -47,6 +47,25 @@ func refreshAndLayout(_ event: RefreshSessionEvent, screenIsDefinitelyUnlocked:
...
+    }
+    havePendingRefresh = true
+    pendingRefreshScreenIsDefinitelyUnlocked = screenIsDefinitelyUnlocked
```
nikitabobko, line 62: It feels more natural to me to coalesce the latest `screenIsDefinitelyUnlocked` state into `pendingRefreshScreenIsDefinitelyUnlocked` in whatever case:

```suggestion
    pendingRefreshScreenIsDefinitelyUnlocked = screenIsDefinitelyUnlocked
    if havePendingRefresh { return }
    havePendingRefresh = true
```

This way we will run the next refresh session with the latest known `screenIsDefinitelyUnlocked`

wdyt?

rickyz: My logic for preferring true here was that the comments at https://github.com/nikitabobko/AeroSpace/blob/ae0e5770e5791a45773629fa19214e43ac7c94a2/Sources/AppBundle/tree/frozen/closedWindowsCache.swift#L111-L124 suggest that resetting the closed window cache on relayout is critical for correctness. Because of that, I was worried that losing a reset via https://github.com/nikitabobko/AeroSpace/blob/ae0e5770e5791a45773629fa19214e43ac7c94a2/Sources/AppBundle/layout/refresh.swift#L12 could potentially trigger that buggy sequence.

nikitabobko: Thanks, makes sense.

Merged with the following cleanup patch:
```diff
diff --git a/Sources/AppBundle/initAppBundle.swift b/Sources/AppBundle/initAppBundle.swift
index 38751b79..c64d0fc8 100644
--- a/Sources/AppBundle/initAppBundle.swift
+++ b/Sources/AppBundle/initAppBundle.swift
@@ -23,7 +23,7 @@ import Foundation
     AXUIElementSetMessagingTimeout(AXUIElementCreateSystemWide(), 1.0)
     startUnixSocketServer()
     GlobalObserver.initObserver()
-    refreshAndLayout(.startup1, screenIsDefinitelyUnlocked: false, startup: true)
+    refreshSession(.startup1, screenIsDefinitelyUnlocked: false, startup: true, body: {})
     refreshSession(.startup2, screenIsDefinitelyUnlocked: false) {
         if serverArgs.startedAtLogin {
             _ = config.afterLoginCommand.runCmdSeq(.defaultEnv, .emptyStdin)
diff --git a/Sources/AppBundle/layout/refresh.swift b/Sources/AppBundle/layout/refresh.swift
index 8ac24370..91cafb31 100644
--- a/Sources/AppBundle/layout/refresh.swift
+++ b/Sources/AppBundle/layout/refresh.swift
@@ -42,27 +42,16 @@ func refreshSession<T>(_ event: RefreshSessionEvent, screenIsDefinitelyUnlocked:
     return result
 }

-@MainActor
-func refreshAndLayout(_ event: RefreshSessionEvent, screenIsDefinitelyUnlocked: Bool, startup: Bool = false) {
-    refreshSession(event, screenIsDefinitelyUnlocked: screenIsDefinitelyUnlocked, startup: startup, body: {})
-}
-
 @MainActor private var havePendingRefresh = false
-@MainActor private var pendingRefreshScreenIsDefinitelyUnlocked = false

 @MainActor
 func scheduleRefreshAndLayout(_ event: RefreshSessionEvent, screenIsDefinitelyUnlocked: Bool = false) {
-    if havePendingRefresh {
-        if screenIsDefinitelyUnlocked {
-            pendingRefreshScreenIsDefinitelyUnlocked = true
-        }
-        return
-    }
+    if screenIsDefinitelyUnlocked { resetClosedWindowsCache() }
+    if havePendingRefresh { return }
     havePendingRefresh = true
-    pendingRefreshScreenIsDefinitelyUnlocked = screenIsDefinitelyUnlocked
     DispatchQueue.main.async { @MainActor in
         havePendingRefresh = false
-        refreshSession(event, screenIsDefinitelyUnlocked: pendingRefreshScreenIsDefinitelyUnlocked, startup: false, body: {})
+        refreshSession(event, screenIsDefinitelyUnlocked: false, startup: false, body: {})
     }
 }

```

PR #1259, opened 2025-04-01
Fix bug causing nonexistent windows to remain in workspace trees.

rickyz:
Previously, when registering window observers fails, the
`window.garbageCollect()` call would be a no-op because the window is
not in `allWindowsMap`. This would lead to such windows being left bound
to the workspace tree, since binding happens inside the window
initializer.


nikitabobko:
Good catch! Merged https://github.com/nikitabobko/AeroSpace/commit/f1f81520fcd9c111535380b071612a32725dfe1e

PR #1344, opened 2025-04-25
adds note about KeePassXC in common pitfalls

audishos:
re. getting stuck in secure input mode, disabling `option` key binding

### DIFF
--- a/docs/guide.adoc
+++ b/docs/guide.adoc
@@ -710,3 +710,5 @@ You can also use `on-window-detected` to force tile or force float all windows o
 == Common pitfall: keyboard keys handling
 
 If you can't make AeroSpace handle some keys in your config, please make sure that you don't have keys conflict with other software that might listen to global keys (e.g. skhd, Karabiner-Elements, Raycast)
+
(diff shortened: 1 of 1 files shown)


docs/guide.adoc
```diff
@@ -710,3 +710,5 @@ You can also use `on-window-detected` to force tile or force float all windows o
 == Common pitfall: keyboard keys handling
 
 If you can't make AeroSpace handle some keys in your config, please make sure that you don't have keys conflict with other software that might listen to global keys (e.g. skhd, Karabiner-Elements, Raycast)
+
+If you have KeePassXC installed and up-to-date (v2.7.10 at the time of writing), you may notice that bindings which use the `option` key will stop working after some time. This is a known issue where KeePassXC gets stuck in secure input mode (see: https://github.com/keepassxreboot/keepassxc/issues/11906[#11906]). The recommendation for now is to downgrade to v2.7.9.
```
nikitabobko, line 714: It's not only about `KeePassXC`. First and foremost, it's a macOS problem (aka "security feature") which is called "secure input" and should be phrased this way. When any "security input" field is focused, global `opt` won't work (e.g. Terminal.app has it in its top menubar `Terminal -> Secure Keyboard Entry`)

The KeePassXC case is interesting that it's possible to break macOS "secure input" in a way that even if the application runs only in background, it can also break global `opt` key

gianlucatruda: I'm also having the same issue with Ferdium on macOS 14.8

PR #1388, opened 2025-05-07
feat: Add TOML Configuration Schema

nealxm:
continuation of discussion #1378

this is a first attempt at a config schema. most of the descriptions are pulled from current docs/default config example. of course if you feel anything is misleading or incorrect I would be happy to revise. currently, taplo is the leading lsp for toml files so this schema is designed to work with it. taplo's behavior is not always consistent when providing autocompletion/documentation on hover (I will be investigating this of course) but I still think this will be very helpful for new users trying to make aerospace their own.

to test this schema add the following directive at the top of your configuration file
`#:schema https://raw.githubusercontent.com/nealxm/AeroSpace/refs/heads/feat-config-schema/resources/config-schema.json`

thanks so much for your time and consideration, I hope you enjoy this addition.

### DIFF
--- a/grammar/config-schema.json
+++ b/grammar/config-schema.json
@@ -1,0 +1,232 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "properties": {
+    "start-at-login": {
+      "description": "start aerospace at login to macos user session.",
+      "type": "boolean",
+      "default": false
+    },
+    "after-login-command": {
+      "description": "run aerospace command at login, only used if start-at-login is true. https://nikitabobko.github.io/AeroSpace/guide#commands",
+      "oneOf": [
+        { "type": "string" },
+        { "type": "array", "items": { "type": "string" } }
+      ]
+    },
+    "after-startup-command": {
+      "description": "run aerospace command when aerospace starts, this is after login commands. https://nikitabobko.github.io/AeroSpace/guide#commands",
+      "oneOf": [
+        { "type": "string" },
+        { "type": "array", "items": { "type": "string" } }
+      ]
+    },
+    "enable-normalization-flatten-containers": {
+      "description": "combine container with parent if it only contains one node. https://nikitabobko.github.io/AeroSpace/guide#normalization",
+      "type": "boolean",
+      "default": true
+    },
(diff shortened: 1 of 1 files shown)


nealxm:
my apologies for the force push, i did not realize the resources folder was reserved and did not permit new files. I hope including the new file in `./grammar` is okay.

here is the updated directive for you to test with:
`#:schema https://raw.githubusercontent.com/nealxm/AeroSpace/refs/heads/feat-config-schema/grammar/config-schema.json`

nikitabobko:
Thanks, the PR needs more love to make it production ready:

**1.** The schema should be auto-generated from https://github.com/nikitabobko/AeroSpace/blob/main/Sources/AppBundle/config/parseConfig.swift#L90 ; Otherwise, the maintenance cost is too high and I won't bother myself updating the schema when I introduce new config options.

You can add schema generation to `./run-tests.sh`. This way the CI will check that we don't forget to update the schema when we add new config options.

You might need to change the definitions of config parsing facilities to make them more declarative. Right now they are very imperative, and they should stay that way, but I'm ok to change them to make them more "declarative-aware". 

**2.** The default config and i3 configs should be changed to include your comment. Or actually, I prefer the starship approach with a real key `"$schema" = ` + a brief comment on top of `"$schema" = ` which explains how to use it (links to docs, which IDEs and LSP support that, etc.)

**3.** The schema should be included in the `AeroSpace.app` bundle distribution, which means that you need to put it to `resources/` and update `AeroSpace.xcodeproj/project.pbxproj` as suggested by the CI. The `schema` in the TOML config then should point to the local file `"$schema" = "/Applications/AeroSpace.app/Contents/Resources/config-schema.json"`

**4.** Integrate schema validation to `./run-tests.sh`

A fair warning: I don't mentor people and I don't help people with the implementation. I only give the pointers and expect people to finish the work on their own. And even so, the PR might be rejected if the quality is not up to par. Otherwise, "the contribution" costs more than it gives. Thank you for your understanding.

I understand that I'm asking a lot from contributors. It's ok if you are not ready to take it on.

I'm closing the PR for now, since there is yet a lot of work to be done. Please open a new PR when you have something more production ready.

PR #1396, opened 2025-05-08
Added 'layout' inter variable to "list-windows" output. 

Fukislim:
This makes it so programs like JankyBorders can change the border color based on "floating" or "tiling" layout.
E.g. the border is red if if it is floating and yellow if it is tiling, by checking the value on change of focus.


nikitabobko:
Low quality PR

PR #1421, opened 2025-05-16
Add `--boundaries` and `--boundaries-action` options for move command

### LINKED ISSUE #183: [Feature Request] Return status of an aerospace command after executing it

gabsrp2002:
## Summary

This PR closes #183

It implements the logic of boundaries in move command exactly as it was described in the task:

- `--boundaries <boundary>` supports the options
  - `workspace` (default): Use the workspace boundaries
  - `all-monitors-outer-frame`: Use the outer frame of all monitors as boundaries
  
- `--boundaries-action <action>` supports options:
  - `create-implicit-container` (default): The command creates an implicit container (as it was by default)
  - `stop`: The command doesn't do anything when there's a boundary in the direction
  - `fail`: The command fails when there's a boundary in the direction
  
## Test

Added tests for the new boundaries actions in the move command.

## Feature demonstration

There's a monitor in the left of my screen.

https://github.com/user-attachments/assets/3a39da7e-9c5a-4850-87f1-e0eb3abfeb50


Sources/AppBundle/command/impl/MoveCommand.swift
```diff
@@ -24,9 +24,31 @@ struct MoveCommand: Command {
...
+                    if args.boundariesAction == .stop {
+                        return true
+                    }
+
+                    if args.boundariesAction == .fail {
+                        return false
+                    }
+                }
+
+                return moveOut(io, window: currentWindow, direction: direction)
```
nikitabobko, line 51: Please take a look at `hitWorkspaceBoundaries` and `hitAllMonitorsOuterFrameBoundaries` functions and make the implementation for the `move` command to follow the same style

Sources/AppBundle/command/impl/MoveCommand.swift
```diff
@@ -24,9 +24,31 @@ struct MoveCommand: Command {
                             currentWindow.bind(to: parent, adaptiveWeight: prevBinding.adaptiveWeight, index: indexOfSiblingTarget)
                     }
                     return true
-                } else {
-                    return moveOut(io, window: currentWindow, direction: direction)
                 }
+
+                if hasWorkspaceBoundaryInDirection(window: currentWindow, direction: direction) {
+                    if args.boundaries == .allMonitorsUnionFrame,
```
nikitabobko, line 30: All enums that can be potentially extended in the future must be switched over exhaustively via `switch-case` construction. `==` for such enums is not ok

Sources/AppBundle/command/impl/MoveCommand.swift
```diff
@@ -24,9 +24,31 @@ struct MoveCommand: Command {
...
+                       let moveNodeToMonitorArgs = parseMoveNodeToMonitorCmdArgs([
+                           "--fail-if-noop",
+                           "--focus-follows-window",
+                           "--window-id", "\(currentWindow.windowId)",
+                           direction.rawValue,
+                       ]).unwrap().0,
```
nikitabobko, line 36: A couple of comments:
1. `--fail-if-noop` doesn't make sense for `move-node-to-monitor <direction>` https://github.com/nikitabobko/AeroSpace/issues/1423
2. Please replace args parsing with manual `MoveNodeToMonitorCmdArgs` object construction

nikitabobko (changes requested):

Sources/AppBundle/command/impl/MoveCommand.swift
nikitabobko: For clarification: "`hitWorkspaceBoundaries` and `hitAllMonitorsOuterFrameBoundaries` style" is not a minor incremental addition to the current patch, but it will significantly change the implementation

Sources/AppBundle/command/impl/MoveCommand.swift
```diff
@@ -37,17 +59,47 @@ struct MoveCommand: Command {
...
+    let innerMostChild =
+        window.parents.first(where: {
+            return switch $0.parent?.cases {
+                case .tilingContainer(let parent): parent.orientation == direction.orientation
+                // Stop searching
+                case .workspace, .macosMinimizedWindowsContainer, nil, .macosFullscreenWindowsContainer,
+                     .macosHiddenAppsWindowsContainer, .macosPopupWindowsContainer:
+                    true
+            }
+        }) as! TilingContainer
```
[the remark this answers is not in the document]
gabsrp2002: Ok, I removed it in the latest commits.

Sources/AppBundle/command/impl/MoveCommand.swift
```diff
@@ -67,7 +119,8 @@ private let moveOutMacosUnconventionalWindow = "moving macOS fullscreen, minimiz
 
             bindTo = parent.rootTilingContainer
             bindToIndex = direction.insertionOffset
-        case .macosMinimizedWindowsContainer, .macosFullscreenWindowsContainer, .macosHiddenAppsWindowsContainer:
+        case .macosMinimizedWindowsContainer, .macosFullscreenWindowsContainer,
+             .macosHiddenAppsWindowsContainer:
```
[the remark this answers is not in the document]
gabsrp2002: Ok, I removed it in the latest commits.

Sources/AppBundleTests/command/MoveCommandTest.swift
```diff
@@ -1,7 +1,8 @@
-@testable import AppBundle
 import Common
 import XCTest
 
+@testable import AppBundle
```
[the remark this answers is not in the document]
gabsrp2002: Ok, I removed it in the latest commits.

Sources/AppBundle/command/impl/MoveCommand.swift
```diff
@@ -37,17 +59,47 @@ struct MoveCommand: Command {
...
+    switch direction {
+        case .left, .up:
+            return window.ownIndex == 0
+        case .right, .down:
+            return window.ownIndex == rootTilingContainer.children.count - 1
+    }
+}
+
+private let moveOutMacosUnconventionalWindow =
+    "moving macOS fullscreen, minimized windows and windows of hidden apps isn't yet supported. This behavior is subject to change"
```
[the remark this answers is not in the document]
gabsrp2002: Ok, I removed it in the latest commits.

Sources/AppBundle/command/impl/MoveCommand.swift
gabsrp2002: Sure, I added it to the latest commit.

Sources/AppBundle/command/impl/MoveCommand.swift
gabsrp2002: Makes sense to me. I changed it in the latest commit.

Sources/AppBundle/command/impl/MoveCommand.swift
gabsrp2002: I applied this in the latest commit.

Sources/AppBundle/command/impl/MoveCommand.swift
```diff
@@ -37,6 +41,83 @@ struct MoveCommand: Command {
...
+) -> Bool {
+    switch args.boundaries {
+        case .workspace:
+            return switch args.boundariesAction {
+                case .stop: true
+                case .fail: false
+                case .createImplicitContainer: moveOut(io, window: window, direction: direction)
+            }
+        case .allMonitorsUnionFrame:
+            return hitAllMonitorsOuterFrameBoundaries(window, io, args, direction, env)
```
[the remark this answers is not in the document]
nikitabobko: And as another side-effect, the body of `hitAllMonitorsOuterFrameBoundaries` function is incorrect either.

The body written in a way where you are trying to move the window by direction to another monitor whilst `boundariesAction` is explicitly set to `createImplicitContainer` by user

nikitabobko (changes requested):

Sources/AppBundle/command/impl/MoveCommand.swift
```diff
@@ -37,6 +41,83 @@ struct MoveCommand: Command {
...
+) -> Bool {
+    switch args.boundariesAction {
+        case .stop:
+            return true
+        case .fail:
+            return false
+        case .createImplicitContainer:
+            let moveNodeToWorkspaceArgs = MoveNodeToWorkspaceCmdArgs(
+                rawArgs: [],
+                windowId: window.windowId,
```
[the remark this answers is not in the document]
gabsrp2002: Ok, I'll remove it and also simplify the logic to use only `MoveNodeToMonitorCmdArgs`.

Sources/AppBundle/command/impl/MoveCommand.swift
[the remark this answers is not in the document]
gabsrp2002: Yep, I forgot to push the commit fixing that. Thanks for flagging it out. I'll push it.

Sources/AppBundle/command/impl/MoveCommand.swift
```diff
@@ -37,6 +41,83 @@ struct MoveCommand: Command {
...
+
+    guard let windowParent = window.parent else {
+        return false // Shouldn't happen
+    }
+
+    if windowParent != rootTilingContainer {
+        return false
+    }
+
+    if rootTilingContainer.orientation != direction.orientation {
```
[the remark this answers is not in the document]
gabsrp2002: Because if there's a layout like:

```
.v_tiles
--- window 1
--- window 2 (focused)
--- window 3
```

and I run `move left` then there's a boundary. Thinking about it better, I think it should've returned true. I'm accidentally creating a implicit container in this case always. I'll fix it.

Sources/AppBundle/command/impl/MoveCommand.swift
```diff
@@ -37,6 +41,79 @@ struct MoveCommand: Command {
...
+        case .allMonitorsUnionFrame:
+            let moveNodeToMonitorArgs = MoveNodeToMonitorCmdArgs(
+                rawArgs: [],
+                target: .directional(direction),
+                focusFollowsWindow: true
+            )
+
+            if MoveNodeToMonitorCommand(args: moveNodeToMonitorArgs).run(env, io) {
+                return true
+            }
```
nikitabobko, line 94: Please don't rely on `MoveNodeToMonitorCommand`'s implicit behavior that it will fail if there are no monitors in the appropriate direction.

`MoveNodeToMonitorCommand` could fail for other reasons.

Please explicitly check if there is a monitor in specified direction similar to how `focus` command does that via `findRelativeMonitor`

Sources/AppBundle/command/impl/MoveCommand.swift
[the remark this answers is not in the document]
nikitabobko: Thanks, I misread your code, I thought you were comparing `window.parent.orientation` and `rootTilingContainer.orientation`

Now when I understood your intend, I disagree with the proposed behavior. I think these two situations should behave in the same way:

```
monitor1                  monitor2
  h_tiles
    window1
    window2 (focused)
// move --boundaries all-monitors-outer-frame right
```

```
monitor1                  monitor2
  v_tiles
    window1
    window2 (focused)
// move --boundaries all-monitors-outer-frame right
```

(monitor2 is to the right of monitor1)

At least because if window2 is the only child of its parent, the situations are visually indistinguishable

nikitabobko (changes requested):

Sources/AppBundle/command/impl/MoveCommand.swift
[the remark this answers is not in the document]
gabsrp2002: Yes, I agree. I changed it to return true.

Sources/AppBundle/command/impl/MoveCommand.swift
gabsrp2002: Ok. I'll change it.

gabsrp2002: It's fixed now. I also made it fail if aerospace can't find the monitors in the given direction for some reason (same logic as the focus command).

Sources/AppBundle/command/impl/MoveCommand.swift
```diff
@@ -37,6 +41,82 @@ struct MoveCommand: Command {
...
+        return false // Shouldn't happen
+    }
+
+    if windowParent != rootTilingContainer {
+        return false
+    }
+
+    if rootTilingContainer.orientation != direction.orientation {
+        return true
+    }
```
nikitabobko, line 61: Please add a test that would cover this case. If I comment this `if` out, the test should fail

Sources/AppBundle/command/impl/MoveCommand.swift
```diff
@@ -37,6 +41,82 @@ struct MoveCommand: Command {
...
+        case .workspace:
+            return switch args.boundariesAction {
+                case .stop: true
+                case .fail: false
+                case .createImplicitContainer: moveOut(io, window: window, direction: direction)
+            }
+        case .allMonitorsUnionFrame:
+            guard let (monitors, index) = window.nodeMonitor?.findRelativeMonitor(inDirection: direction) else {
+                return io.err("Can't find monitors in direction \(direction)")
+            }
```
nikitabobko, line 88: So when `boundaries` == `allMonitorsUnionFrame` and there is no monitor in the given direction, you always fail. It doesn't look right

Consider `move --boundaries all-monitors-outer-frame --boundaries-action create-implicit-container right`

Please make sure to check for all combination of flags before submitting the review

nikitabobko (changes requested):

Sources/AppBundle/command/impl/MoveCommand.swift
```diff
@@ -37,6 +41,82 @@ struct MoveCommand: Command {
...
+        return false // Shouldn't happen
+    }
+
+    guard let windowParent = window.parent else {
+        return false // Shouldn't happen
+    }
+
+    if windowParent != rootTilingContainer {
+        return false
+    }
```
nikitabobko, line 57: Please add a test that would cover this case. If I comment this `if` out, the test should fail

Sources/AppBundle/command/impl/MoveCommand.swift
```diff
@@ -37,6 +41,82 @@ struct MoveCommand: Command {
...
+    if rootTilingContainer.orientation != direction.orientation {
+        return true
+    }
+
+    switch direction {
+        case .left, .up:
+            return window.ownIndex == 0
+        case .right, .down:
+            return window.ownIndex == rootTilingContainer.children.count - 1
+    }
```
nikitabobko, line 68: Please add a test that would cover this case. If I replace this `switch` with constant `return true`/`return false`, the test should fail

Sources/AppBundle/command/impl/MoveCommand.swift
gabsrp2002: For what I understood about the method implementation, this only returns nil when the the current monitor is not found:
```swift
    func findRelativeMonitor(inDirection direction: CardinalDirection) -> (monitorsInDirection: [Monitor], index: Int)? {
        let currentMonitor = self
        let monitors = sortedMonitors.filter {
            currentMonitor.rect.topLeftCorner == $0.rect.topLeftCorner ||
                $0.relation(to: currentMonitor) == direction.orientation
        }
        guard let index = monitors.firstIndex(where: { $0.rect.topLeftCorner == currentMonitor.rect.topLeftCorner }) else { return nil }
        return (monitors, index + direction.focusOffset)
    }
```

I could change the error message, but I think it should return a fail. This is actually a fail, isn't it? This is the same that happens in the focus command if the boundary is `all-monitors-outer-frame`, regardless of the boundary action:

```swift
@MainActor private func hitWorkspaceBoundaries(
    _ target: LiveFocus,
    _ io: CmdIo,
    _ args: FocusCmdArgs,
    _ direction: CardinalDirection
) -> Bool {
    switch args.boundaries {
        case .workspace:
            return switch args.boundariesAction {
                case .stop: true
                case .fail: false
                case .wrapAroundTheWorkspace: wrapAroundTheWorkspace(target, io, direction)
                case .wrapAroundAllMonitors: dieT("Must be discarded by args parser")
            }
        case .allMonitorsUnionFrame:
            let currentMonitor = target.workspace.workspaceMonitor
            guard let (monitors, index) = currentMonitor.findRelativeMonitor(inDirection: direction) else {
                return io.err("Can't find monitor in direction \(direction)")
            }

            if let targetMonitor = monitors.getOrNil(atIndex: index) {
                return targetMonitor.activeWorkspace.focusWorkspace()
            } else {
                guard let wrapped = monitors.get(wrappingIndex: index) else { return false }
                return hitAllMonitorsOuterFrameBoundaries(target, io, args, direction, wrapped)
            }
    }
}
```

Did I misunderstand  something? I tested the exactly combination of flags you sent me and it worked fine.

nikitabobko: Ah, thanks for the elaboration, you're right. The message `Can't find monitors in direction \(direction)"` definitely tricked me, but I understand that you copied it from other `findRelativeMonitor` usages.

I think the message should be changed to smth `io.err("Should never happen. Can't find current monitor")` in all of the usages.

If you want to feel free to fix it; otherwise, I will do it myself later anyway

gabsrp2002: Sure, I'll do it

Sources/AppBundle/command/impl/MoveCommand.swift
gabsrp2002: Ok

Sources/AppBundle/command/impl/MoveCommand.swift
gabsrp2002: Ok

Sources/AppBundle/command/impl/MoveCommand.swift
gabsrp2002: Ok

Sources/AppBundle/command/impl/MoveCommand.swift
gabsrp2002: While trying to implement those tests I've noticed that a situation like this one can occur:

```
h_tiles
  window1
  v_tiles
    window2 (focused)
    window3
```

Then a boundary should be detected when trying to `move right`. I've updated the boundary detection logic to include this case. I'll add tests later, but I've uploaded the new logic, which I think is better.

rwblokzijl:
Cant wait for this feature. Especially since there is no workaround available. @nikitabobko any chance you could review this?

nikitabobko:
Merged with the following patch on top:

```diff
diff --git a/Sources/AppBundle/command/impl/MoveCommand.swift b/Sources/AppBundle/command/impl/MoveCommand.swift
index 1ce97e80..1aea9b2d 100644
--- a/Sources/AppBundle/command/impl/MoveCommand.swift
+++ b/Sources/AppBundle/command/impl/MoveCommand.swift
@@ -26,11 +26,7 @@ struct MoveCommand: Command {
                     }
                 }
 
-                if hasWorkspaceBoundaryInDirection(node: currentWindow, direction: direction) {
-                    return hitWorkspaceBoundaries(currentWindow, io, args, direction, env)
-                }
-
-                return moveOut(io, window: currentWindow, direction: direction)
+                return moveOut(window: currentWindow, direction: direction, io, args, env)
             case .workspace: // floating window
                 return io.err("moving floating windows isn't yet supported") // todo
             case .macosMinimizedWindowsContainer, .macosFullscreenWindowsContainer, .macosHiddenAppsWindowsContainer:
@@ -41,65 +37,9 @@ struct MoveCommand: Command {
     }
 }
 
-protocol TilingTreeMember {
-    var nodeWorkspace: Workspace? { get }
-    var parent: NonLeafTreeNodeObject? { get }
-    var index: Int { get }
-}
-
-extension TilingContainer: TilingTreeMember {
-    var index: Int { ownIndex ?? -1 }
-}
-
-extension Window: TilingTreeMember {
-    var index: Int { ownIndex }
-}
-
-@MainActor private func hasWorkspaceBoundaryInDirection(
-    node: TilingTreeMember, direction: CardinalDirection,
-) -> Bool {
-    // Asserts that the node is part of the tiling tree
-    guard let rootTilingContainer = node.nodeWorkspace?.rootTilingContainer else {
-        return false
-    }
-
-    // Make node mutable
-    var node = node
-
-    while node as? TilingContainer != rootTilingContainer {
-        // Asserts that the node is part of the tiling tree
-        guard let nodeParent = node.parent as? TilingContainer else {
-            return false
-        }
-
-        // If it's on the boundary of its parent, check for the parent
-        if nodeParent.orientation != direction.orientation {
-            node = nodeParent
-            continue
-        }
-
-        // If it's not on the boundary of its parent, it's not on the boundary of the workspace
-        switch direction {
-            case .left, .up:
-                if node.index != 0 {
-                    return false
-                }
-            case .right, .down:
-                if node.index != rootTilingContainer.children.count - 1 {
-                    return false
-                }
-        }
-
-        // If it's on the boundary of its parent, check for the parent
-        node = nodeParent
-    }
-
-    // If we reached the root tiling container, it's on the boundary of the workspace
-    return true
-}
-
 @MainActor private func hitWorkspaceBoundaries(
     _ window: Window,
+    _ workspace: Workspace,
     _ io: CmdIo,
     _ args: MoveCmdArgs,
     _ direction: CardinalDirection,
@@ -107,10 +47,12 @@ extension Window: TilingTreeMember {
 ) -> Bool {
     switch args.boundaries {
         case .workspace:
-            return switch args.boundariesAction {
-                case .stop: true
-                case .fail: false
-                case .createImplicitContainer: moveOut(io, window: window, direction: direction)
+            switch args.boundariesAction {
+                case .stop: return true
+                case .fail: return false
+                case .createImplicitContainer:
+                    createImplicitContainerAndMoveWindow(window, workspace, direction)
+                    return true
             }
         case .allMonitorsUnionFrame:
             guard let (monitors, index) = window.nodeMonitor?.findRelativeMonitor(inDirection: direction) else {
@@ -118,7 +60,7 @@ extension Window: TilingTreeMember {
             }
 
             guard monitors.getOrNil(atIndex: index) != nil else {
-                return hitAllMonitorsOuterFrameBoundaries(window, io, args, direction)
+                return hitAllMonitorsOuterFrameBoundaries(window, workspace, io, args, direction)
             }
 
             let moveNodeToMonitorArgs = MoveNodeToMonitorCmdArgs(
@@ -133,23 +75,29 @@ extension Window: TilingTreeMember {
 
 @MainActor private func hitAllMonitorsOuterFrameBoundaries(
     _ window: Window,
+    _ workspace: Workspace,
     _ io: CmdIo,
     _ args: MoveCmdArgs,
     _ direction: CardinalDirection
 ) -> Bool {
     switch args.boundariesAction {
-        case .stop:
-            return true
-        case .fail:
-            return false
+        case .stop: return true
+        case .fail: return false
         case .createImplicitContainer:
-            return moveOut(io, window: window, direction: direction)
+            createImplicitContainerAndMoveWindow(window, workspace, direction)
+            return true
     }
 }
 
 private let moveOutMacosUnconventionalWindow = "moving macOS fullscreen, minimized windows and windows of hidden apps isn't yet supported. This behavior is subject to change"
 
-@MainActor private func moveOut(_ io: CmdIo, window: Window, direction: CardinalDirection) -> Bool {
+@MainActor private func moveOut(
+    window: Window,
+    direction: CardinalDirection,
+    _ io: CmdIo,
+    _ args: MoveCmdArgs,
+    _ env: CmdEnv,
+) -> Bool {
     let innerMostChild = window.parents.first(where: {
         return switch $0.parent?.cases {
             case .tilingContainer(let parent): parent.orientation == direction.orientation
@@ -159,25 +107,15 @@ private let moveOutMacosUnconventionalWindow = "moving macOS fullscreen, minimiz
         }
     }) as? TilingContainer
     guard let innerMostChild else { return false }
-    let bindTo: TilingContainer
-    let bindToIndex: Int
     guard let parent = innerMostChild.parent else { return false }
     switch parent.nodeCases {
         case .tilingContainer(let parent):
             check(parent.orientation == direction.orientation)
-            bindTo = parent
             guard let ownIndex = innerMostChild.ownIndex else { return false }
-            bindToIndex = ownIndex + direction.insertionOffset
-        case .workspace(let parent): // create implicit container
-            let prevRoot = parent.rootTilingContainer
-            prevRoot.unbindFromParent()
-            // Force tiles layout
-            _ = TilingContainer(parent: parent, adaptiveWeight: WEIGHT_AUTO, direction.orientation, .tiles, index: 0)
-            check(prevRoot != parent.rootTilingContainer)
-            prevRoot.bind(to: parent.rootTilingContainer, adaptiveWeight: WEIGHT_AUTO, index: 0)
-
-            bindTo = parent.rootTilingContainer
-            bindToIndex = direction.insertionOffset
+            window.bind(to: parent, adaptiveWeight: WEIGHT_AUTO, index: ownIndex + direction.insertionOffset)
+            return true
+        case .workspace(let parent):
+            return hitWorkspaceBoundaries(window, parent, io, args, direction, env)
         case .macosMinimizedWindowsContainer, .macosFullscreenWindowsContainer, .macosHiddenAppsWindowsContainer:
             return io.err(moveOutMacosUnconventionalWindow)
         case .macosPopupWindowsContainer:
@@ -185,13 +123,21 @@ private let moveOutMacosUnconventionalWindow = "moving macOS fullscreen, minimiz
         case .window:
             die("Window can't contain children nodes")
     }
+}
 
-    window.bind(
-        to: bindTo,
-        adaptiveWeight: WEIGHT_AUTO,
-        index: bindToIndex
-    )
-    return true
+@MainActor private func createImplicitContainerAndMoveWindow(
+    _ window: Window,
+    _ workspace: Workspace,
+    _ direction: CardinalDirection
+) {
+    let prevRoot = workspace.rootTilingContainer
+    prevRoot.unbindFromParent()
+    // Force tiles layout
+    _ = TilingContainer(parent: workspace, adaptiveWeight: WEIGHT_AUTO, direction.orientation, .tiles, index: 0)
+    check(prevRoot != workspace.rootTilingContainer)
+    prevRoot.bind(to: workspace.rootTilingContainer, adaptiveWeight: WEIGHT_AUTO, index: 0)
+
+    window.bind(to: workspace.rootTilingContainer, adaptiveWeight: WEIGHT_AUTO, index: direction.insertionOffset)
 }
 
 @MainActor private func deepMoveIn(window: Window, into container: TilingContainer, moveDirection: CardinalDirection) -> Bool {
diff --git a/Sources/AppBundleTests/command/MoveCommandTest.swift b/Sources/AppBundleTests/command/MoveCommandTest.swift
index 5404e556..1d4b7cf1 100644
--- a/Sources/AppBundleTests/command/MoveCommandTest.swift
+++ b/Sources/AppBundleTests/command/MoveCommandTest.swift
@@ -193,7 +193,7 @@ final class MoveCommandTest: XCTestCase {
         assertEquals(
             workspace.layoutDescription,
             .workspace([
-                .h_tiles([.window(1), .v_tiles([.window(2), .window(3)])]),
+                .h_tiles([.window(1), .v_tiles([.window(3)]), .window(2)]),
             ])
         )
         assertEquals(result.exitCode, 0)
diff --git a/docs/aerospace-move.adoc b/docs/aerospace-move.adoc
index dd37483e..02edde88 100644
--- a/docs/aerospace-move.adoc
+++ b/docs/aerospace-move.adoc
@@ -35,7 +35,7 @@ Defines move boundaries. +
 The default is: `workspace`
 
 --boundaries-action <boundary-action>::
-Defines move boundaries action. +
+Defines the behavior when requested to move across the `<boundary>`. +
 `<boundary-action>` possible values: `(stop|fail|create-implicit-container)`. +
 The default is: `create-implicit-container`
 
@@ -127,9 +127,10 @@ v_tiles
 +
 `v_tiles` is an implicitly created container.
 +
-*Remark*: If `--boundaries` is set to `all-monitors-outer-frame` and there is a monitor in the `up` direction, the implicit container isn't be created.
+*Remark*: If `--boundaries` is set to `all-monitors-outer-frame` and there is a monitor in the `up` direction,
+the implicit container isn't be created.
 +
-Instead, `window 2` would be sent to the monitor.
+Instead, `window 2` would be moved to the monitor above the current.
 
 // end::body[]

```

Issue #397, opened 2024-08-12
Allow percentages in `resize` command


nikitabobko:
For example
- `resize width 25%` should make window to occupy 1/4 of the monitor width
- `resize width 50%` should make window to occupy half of the monitor
- `resize width +10%` should increase window's width by 10% of the monitor's width

canscaw28:
+1

nikitabobko:
And in `accordion-padding` setting

PR #1440, opened 2025-05-26
Implement sizing via percentages

ernesto-jimenez:
This tackles issue https://github.com/nikitabobko/AeroSpace/issues/397


CheariX:
Sounds super useful. How does it behave if you switch a workspace to a different monitor with a different resolution? Does the layout stay consistent, i.e. does it use the same percentage levels on the different screen? I think Yabai did this and it was super useful.

PR #1269, opened 2025-04-08
feat(config): Add window count and workspace pattern support for dyna…

xand3r40r93:
…mic gaps

Enhance DynamicConfigValue to support window count and workspace pattern matching in per-monitor configurations. This allows for more granular control over gaps based on the number of windows and workspace names.

Key changes:
- Add windows and workspace parameters for PerMonitorValue
- Implement window count aware gap configurations
- Add workspace pattern matching support
- Update documentation with new configuration examples
- Add comprehensive test coverage

PR #1512, opened 2025-06-28
Add max width settings for single-window workspaces and fullscreen

mybuddymichael:
## The problem

When using an ultrawide monitor, a single window in a workspace can often be too wide.

## The solution

This commit introduces two new settings and an additional flag to the fullscreen command that make single windows behave better on wider monitors. The changes are fully backwards compatible.

New settings:

1. `single-window-max-width-percent`: Sets maximum width percentage (1-100) for windows which are the only window in a workspace, and also limits the width of the workspace when in fullscreen mode
2. `single-window-exclude-app-ids`: An array of app bundle IDs to exclude from the width limit

New flag to the `fullscreen` command:

- `--no-max-width`: Ignores single-window width constraints when in fullscreen mode

## Additional notes

With these settings and the flag, the user can separately bind fullscreen in both the normal (constrained) mode and in the unconstrained mode, allowing quick switching between the two. (This is how I use it and it feels very natural.)

This addresses #60.

### DIFF
--- a/Sources/AppBundle/command/impl/FullscreenCommand.swift
+++ b/Sources/AppBundle/command/impl/FullscreenCommand.swift
@@ -21,6 +21,7 @@ struct FullscreenCommand: Command {
         }
         window.isFullscreen = newState
         window.noOuterGapsInFullscreen = args.noOuterGaps
+        window.noMaxWidthInFullscreen = args.noMaxWidth
 
         // Focus on its own workspace
         window.markAsMostRecentChild()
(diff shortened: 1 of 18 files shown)


jjhidalgar:
Related: https://github.com/nikitabobko/AeroSpace/pull/1269

nikitabobko:
Thanks, but I reject the PR.

1. The feature introduces quite new concepts that I want to think about myself. I don't trust third-party contributors to do the thinking and implementation of this specific feature.
2. The implementation in this PR is far from what I imagine.

mybuddymichael:
No problem. Thanks for creating this project!

Issue #1495, opened 2025-06-24
Create a proper config error reporting dialog


nikitabobko:
Right now, we use AppleScript which doesn't look reliable, especially slash escaping part:

```
let args = [
    "-e",
    """
    display dialog "\(message.replacing("\"", with: "\\\""))" with title "\(title)"
    """,
]
Result { try? Process.run(URL(filePath: "/usr/bin/osascript"), arguments: args) }.getOrDie()
```

<img width="466" alt="Image" src="https://github.com/user-attachments/assets/ce455573-6a13-4efb-91fc-3c7af118e2b1" />

And the AppleScript window has max-width restriction which is not nice

Ghostty for inspiration:

<img width="910" alt="Image" src="https://github.com/user-attachments/assets/2861e406-80e9-4acc-b1f4-1f16ceafb57e" />

mobile-ar:
Will something like this work for you? It's fully SwiftUI

<img width="611" height="407" alt="Image" src="https://github.com/user-attachments/assets/a8841ee5-4d33-472f-8309-4eef4d37aa89" />

It will be triggered like this:

```swift
Task { @MainActor in
    MessageModel.shared.message = Message(title: "...", description: "...", body: "...")
}
```

I can submit a PR, but still it's a WIP. There is this `func showMessageInGui` in `common` that's being used from `func dieT(...`, so I'm still trying to figure out how to use this from common and from AppBundle.
Also, need to figure out how to show this window on top of all the workspace windows as the current AppleScript one.

nikitabobko:
@mobile-ar I assume that the error text is selectable? Looks good, thank you, I'd be happy to receive a PR for this

> I can submit a PR, but still it's a WIP. There is this func showMessageInGui in common that's being used from func dieT(..., so I'm still trying to figure out how to use this from common and from AppBundle.

You don't need to. You can leave `showMessageInGui` + Console.app reporting for fatal errors, and introduce this new SwiftUI window only for config errors

>  Also, need to figure out how to show this window on top of all the workspace windows as the current AppleScript one.

To make it appear on all workspaces, you can do this:

```diff
diff --git a/Sources/AppBundle/tree/MacApp.swift b/Sources/AppBundle/tree/MacApp.swift
index 382ceeff..5798198a 100644
--- a/Sources/AppBundle/tree/MacApp.swift
+++ b/Sources/AppBundle/tree/MacApp.swift
[...]

PR #1545, opened 2025-07-13
Add a customizable MesssageView (initially for config error reporting)

mobile-ar:
This PR is meant to help solve issue #1495  - As required the text is selectable, both description and body.

Notes:
- Refactored `MenuBar` to moved `"Open config in [editor]"` into a `var openConfigButton: some View { ...` to be able to be reused inside the MessageView to show that same option in this new window for easy access.
- `Message` has a `type`, for now it's a 1 case enum that uses `config` as default, it's meant to support later customizations if other type of messages need to be displayed (to customize for example the extra button action or even the image that wants to be displayed).
- Used the `#Preview` macro inside `MessageView.swift` to use live previews in Xcode to iterate through the UI faster (this can be removed it won't affect functionality, but it's nice to have when working with UI).

### DIFF
--- a/Sources/AppBundle/command/impl/ReloadConfigCommand.swift
+++ b/Sources/AppBundle/command/impl/ReloadConfigCommand.swift
@@ -33,16 +33,15 @@ struct ReloadConfigCommand: Command {
                 configUrl = url
                 activateMode(activeMode)
                 syncStartAtLogin()
+                MessageModel.shared.message = nil
             }
             return true
         case .failure(let msg):
             stdout.append(msg)
             if !args.noGui {
-                showMessageInGui(
-                    filenameIfConsoleApp: nil,
-                    title: "AeroSpace Config Error",
-                    message: msg,
-                )
+                Task { @MainActor in
+                    MessageModel.shared.message = Message(description: "AeroSpace Config Error", body: msg)
+                }
             }
             return false
     }
--- a/Sources/AppBundle/ui/MessageView.swift
+++ b/Sources/AppBundle/ui/MessageView.swift
@@ -7,6 +7,7 @@ public func getMessageWindow(messageModel: MessageModel) -> some Scene {
     SwiftUI.Window(messageModel.message?.title ?? aeroSpaceAppName, id: messageWindowId) {
         MessageView(model: messageModel)
             .onAppear {
+                NSApp.setActivationPolicy(.accessory)
                 NSApplication.shared.windows.forEach {
                     if let identifier = $0.identifier?.rawValue, identifier == messageWindowId {
                         $0.level = .floating
@@ -23,6 +24,7 @@ public let messageWindowId = "\(aeroSpaceAppName).messageView"
 public struct MessageView: View {
     @StateObject private var model: MessageModel
     @Environment(\.dismiss) private var dismiss
+    @FocusState var focus: Bool
 
     public init(model: MessageModel) {
         self._model = .init(wrappedValue: model)
@@ -36,6 +38,7 @@ public struct MessageView: View {
                     .font(.system(size: 48))
                 Text("\(model.message?.description ?? "")")
                     .padding(.horizontal)
+                    .focusable()
             }
             .padding()
             ScrollView {
@@ -43,6 +46,11 @@ public struct MessageView: View {
                     HStack {
                         TextEditor(text: .constant(model.message?.body ?? ""))
                             .font(.system(size: 12).monospaced())
+                            .focused($focus)
+                            .onExitCommand {
+                                // Escape to remove focus from the TextEditor, and then you can hit Return to trigger the Close, or use 'CMD + ,' for open config and 'CMD + R' for reload config (same as the menu shortcuts.
+                                focus = false
+                            }
                         Spacer()
                     }
                     Spacer()
@@ -78,6 +86,9 @@ public struct MessageView: View {
         .onDisappear {
             // If user closes the screen with the macOS native close (x) button and then the error is still the same, this window will not appear again
             model.message = nil
+        }
+        .onAppear {
+            focus = true
         }
     }
 }
--- a/Sources/AppBundle/ui/MessageView.swift
+++ b/Sources/AppBundle/ui/MessageView.swift
@@ -1,0 +1,112 @@
+import Common
+import SwiftUI
+
+@MainActor
+public func getMessageWindow(messageModel: MessageModel) -> some Scene {
+    // Using SwiftUI.Window because another class in AeroSpace is already called Window
+    SwiftUI.Window(messageModel.message?.title ?? aeroSpaceAppName, id: messageWindowId) {
+        MessageView(model: messageModel)
+            .onAppear {
+                NSApplication.shared.windows.forEach {
+                    if let identifier = $0.identifier?.rawValue, identifier == messageWindowId {
+                        $0.level = .floating
+                    }
+                }
+            }
+    }
+    .windowResizability(.contentMinSize)
+    //.windowLevel(.floating) //This might be the SwiftUI way of doing window level instead of the onAppear block above, but it's only available from macOS 15.0
+}
+
+public let messageWindowId = "\(aeroSpaceAppName).messageView"
+
+public struct MessageView: View {
+    @StateObject private var model: MessageModel
+    @Environment(\.dismiss) private var dismiss
+
+    public init(model: MessageModel) {
+        self._model = .init(wrappedValue: model)
+    }
+
+    public var body: some View {
+        VStack(alignment: .leading) {
+            HStack(alignment: .center) {
+                Image(systemName: "exclamationmark.triangle.fill")
+                    .foregroundColor(.yellow)
+                    .font(.system(size: 48))
+                Text("\(model.message?.description ?? "")")
+                    .padding(.horizontal)
+            }
+            .padding()
+            ScrollView {
+                VStack(alignment: .leading) {
+                    HStack {
+                        TextEditor(text: .constant(model.message?.body ?? ""))
+                            .font(.system(size: 12).monospaced())
+                            .textSelection(.enabled)
+                        Spacer()
+                    }
+                    Spacer()
+                }
+                .padding()
+            }
+            .background(Color(.controlBackgroundColor))
+            .clipShape(RoundedRectangle(cornerRadius: 8, style: .continuous))
+            .padding(.horizontal)
+            HStack {
+                Spacer()
+                if let type = model.message?.type {
+                    switch type {
+                        case .config:
+                            openConfigButton
+                    }
+                }
+                Button("Continue") {
+                    model.message = nil
+                    self.dismiss()
+                }
+                .keyboardShortcut(.defaultAction)
+            }
+            .padding()
+        }
+        .textSelection(.enabled)
+        .frame(minWidth: 480, maxWidth: 960, minHeight: 200)
+        .onChange(of: model.message) { message in
+            if message == nil {
+                self.dismiss()
+            }
+        }
+    }
+}
+
+public class MessageModel: ObservableObject {
+    @MainActor public static let shared = MessageModel()
+    @Published public var message: Message? = nil
+
+    private init() {}
+}
+
+public enum MessageType {
+    case config
+}
+
+public struct Message: Hashable, Equatable {
+    public let type: MessageType
+    public let title: String
+    public let description: String
+    public let body: String
+
+    init(type: MessageType = .config, title: String = aeroSpaceAppName, description: String, body: String) {
+        self.type = type
+        self.title = title
+        self.description = description
+        self.body = body
+    }
+}
+
+#Preview {
+    MessageView(model: MessageModel.shared)
+        .onAppear {
+            MessageModel.shared.message = Message(type: .config, description: "Description", body: "Body")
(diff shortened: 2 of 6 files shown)


Sources/AppBundle/ui/MessageView.swift
```diff
@@ -0,0 +1,112 @@
...
+            .padding(.horizontal)
+            HStack {
+                Spacer()
+                if let type = model.message?.type {
+                    switch type {
+                        case .config:
+                            openConfigButton
+                    }
+                }
+                Button("Continue") {
```
nikitabobko: I'd prefer to call the button "Close" because that's exactly what it does.

Sources/AppBundle/ui/MessageView.swift
```diff
@@ -0,0 +1,112 @@
...
+                    model.message = nil
+                    self.dismiss()
```
nikitabobko: Why do you need an explicit `.dismiss`? Won't `onChange` do the `.dismiss` for you?

Sources/AppBundle/ui/MessageView.swift
```diff
@@ -0,0 +1,112 @@
...
+            }
+            .background(Color(.controlBackgroundColor))
+            .clipShape(RoundedRectangle(cornerRadius: 8, style: .continuous))
+            .padding(.horizontal)
+            HStack {
+                Spacer()
+                if let type = model.message?.type {
+                    switch type {
+                        case .config:
+                            openConfigButton
```
nikitabobko, line 69: Love the open config button here. Let's also put a reload config button here

Sources/AppBundle/ui/MessageView.swift
```diff
@@ -0,0 +1,112 @@
...
+                    }
+                }
+                Button("Continue") {
+                    model.message = nil
+                    self.dismiss()
+                }
+                .keyboardShortcut(.defaultAction)
```
nikitabobko, line 75: Is this button supposed to trigger on enter key? For some reason, it doesn't do that on my machine. It highlights the button as blue, but the enter key does nothing

Sources/AppBundle/command/impl/ReloadConfigCommand.swift
```diff
@@ -33,16 +33,15 @@ struct ReloadConfigCommand: Command {
                 configUrl = url
                 activateMode(activeMode)
                 syncStartAtLogin()
+                MessageModel.shared.message = nil
             }
             return true
         case .failure(let msg):
             stdout.append(msg)
             if !args.noGui {
-                showMessageInGui(
```
nikitabobko, line 41: `showMessageInGui` can be now cleaned up. We can make `filenameIfConsoleApp` non-Optional + delete dead code inside

Sources/AppBundle/ui/MessageView.swift
```diff
@@ -0,0 +1,112 @@
...
+                Text("\(model.message?.description ?? "")")
+                    .padding(.horizontal)
+            }
+            .padding()
+            ScrollView {
+                VStack(alignment: .leading) {
+                    HStack {
+                        TextEditor(text: .constant(model.message?.body ?? ""))
+                            .font(.system(size: 12).monospaced())
+                            .textSelection(.enabled)
```
nikitabobko: Keyboard arrows for some reason don't work in TextEditor

nikitabobko: Do we need an explicit `.textSelection(.enabled)`? If I drop it, the text is still selectable

nikitabobko (changes requested):
Thank you, mostly looks good. I've left a couple of comments

Sources/AppBundle/command/impl/ReloadConfigCommand.swift
mobile-ar: Done!

Sources/AppBundle/ui/MessageView.swift
mobile-ar: Done

Sources/AppBundle/ui/MessageView.swift
mobile-ar: You are right! Fixed!

Sources/AppBundle/ui/MessageView.swift
mobile-ar: I'm glad to hear that. Added 'reload config' button now.

Sources/AppBundle/ui/MessageView.swift
mobile-ar: True that. Removed. It was unnecessary.

Sources/AppBundle/ui/MessageView.swift
mobile-ar: This is something I'm still trying to figure out. In the 'preview' text is selectable using keyboard and actions are triggered like the close button on enter key press. But the MessageView window is for some reason not being identified as an active window.

For example: if you open any text editor and then you trigger the reload config to show an error, even you can see the cursor in the messageView body, you can see that the cursor is also active in the background editor, and the keyboard interaction goes always to the editor or app behind the MessageView for some reason).

I think it might be related to the app being set up as `Application is agent (UIElement) = YES` in the target configuration. 
This option here: 
image

I will try to research a little bit more to try finding a solution for this. If you have any ideas of what this could be related to, let me know.

Sources/AppBundle/ui/MessageView.swift
mobile-ar: This is the same as `.keyboardShortcut(.defaultAction)` issue I replied above.

Sources/AppBundle/ui/MessageView.swift
mobile-ar: Ok, I think I understand now the issue, and I have a fix for this. 
I need to set `NSApp.setActivationPolicy(.accessory)` so the Message window is able to be in focus and accept keyboard input.

Then in `MacApp.swift` I need to skip registering AeroSpace Message widow by adding this line:
``` swift
if nsApp.bundleIdentifier == aeroSpaceAppId { return nil }
```
But the issue I'm getting is that when running AeroSpace from terminal (using `run-debug.sh`) is that `nsApp.bundleIdentifier` is `nil`, windows get registered and app crashes. If I run this from Xcode using the `AeroSpace.xcodeproj` it works correctly, because the app gets the bundleId from the info.plist setup in the .xcodeproj.
So I found this as a way to avoid this issue on debug :
``` swift
#if DEBUG
    if let bundleURL = nsApp.bundleURL, bundleURL.absoluteString.contains("AeroSpaceApp") { return nil }
#endif
```
@nikitabobko Are you ok with this or do you recommend other way to do this? You can find this changes in the [last commit](https://github.com/nikitabobko/AeroSpace/pull/1545/commits/22e20e97cd6823e587b6eda3feb9eb71c4827207). I will update after your recommendations if needed.

nikitabobko: Thank you for taking a look. That works for me.

I'd only replace the `bundleURL`/`bundleId` comparison with PID comparison as it doesn't depend on debug/release

nikitabobko:
Thanks, merged with the following patch on top (minor code style changes + pid comparison mentioned above):

```diff
diff --git a/Sources/AppBundle/tree/MacApp.swift b/Sources/AppBundle/tree/MacApp.swift
index 511c5607..bb2c717b 100644
--- a/Sources/AppBundle/tree/MacApp.swift
+++ b/Sources/AppBundle/tree/MacApp.swift
@@ -42,11 +42,9 @@ final class MacApp: AbstractApp {
         // Don't perceive any of the lock screen windows as real windows
         // Otherwise, false positive ax notifications might trigger that lead to gcWindows
         if nsApp.bundleIdentifier == lockScreenAppBundleId { return nil }
-        if nsApp.bundleIdentifier == aeroSpaceAppId { return nil }
-        #if DEBUG
-            if let bundleURL = nsApp.bundleURL, bundleURL.absoluteString.contains("AeroSpaceApp") { return nil }
-        #endif
         let pid = nsApp.processIdentifier
+        // AX requests crash if you send them to yourself
+        if pid == myPid { return nil }
 
         while true {
             if let existing = allAppsMap[pid] { return existing }
diff --git a/Sources/AppBundle/ui/MenuBar.swift b/Sources/AppBundle/ui/MenuBar.swift
index c373ec8f..82dfee32 100644
--- a/Sources/AppBundle/ui/MenuBar.swift
+++ b/Sources/AppBundle/ui/MenuBar.swift
@@ -94,8 +94,6 @@ var reloadConfigButton: some View {
                 try await runSession(.menuBarButton, token) { _ = reloadConfig() }
             }
         }.keyboardShortcut("R", modifiers: .command)
-    } else {
-        Text("").hidden()
     }
 }
 
diff --git a/Sources/AppBundle/ui/MessageView.swift b/Sources/AppBundle/ui/MessageView.swift
index cd5eff41..132c5195 100644
--- a/Sources/AppBundle/ui/MessageView.swift
+++ b/Sources/AppBundle/ui/MessageView.swift
@@ -7,9 +7,10 @@ public func getMessageWindow(messageModel: MessageModel) -> some Scene {
     SwiftUI.Window(messageModel.message?.title ?? aeroSpaceAppName, id: messageWindowId) {
         MessageView(model: messageModel)
             .onAppear {
+                // Set activation policy; otherwise, AeroSpace windows won't be able to receive focus and accept keyborad input
                 NSApp.setActivationPolicy(.accessory)
                 NSApplication.shared.windows.forEach {
-                    if let identifier = $0.identifier?.rawValue, identifier == messageWindowId {
+                    if $0.identifier?.rawValue == messageWindowId {
                         $0.level = .floating
                     }
                 }
@@ -48,7 +49,8 @@ public struct MessageView: View {
                             .font(.system(size: 12).monospaced())
                             .focused($focus)
                             .onExitCommand {
-                                // Escape to remove focus from the TextEditor, and then you can hit Return to trigger the Close, or use 'CMD + ,' for open config and 'CMD + R' for reload config (same as the menu shortcuts.
+                                // Escape to remove focus from the TextEditor, and then you can hit Return to trigger the Close,
+                                // or use 'CMD + ,' for open config and 'CMD + R' for reload config (same as the menu shortcuts)
                                 focus = false
                             }
                         Spacer()
diff --git a/Sources/AppBundle/util/appBundleUtil.swift b/Sources/AppBundle/util/appBundleUtil.swift
index b1a08d8b..5820bd99 100644
--- a/Sources/AppBundle/util/appBundleUtil.swift
+++ b/Sources/AppBundle/util/appBundleUtil.swift
@@ -5,6 +5,7 @@ import os
 
 let signposter = OSSignposter(subsystem: aeroSpaceAppId, category: .pointsOfInterest)
 
+let myPid = NSRunningApplication.current.processIdentifier
 let lockScreenAppBundleId = "com.apple.loginwindow"
 let AEROSPACE_WINDOW_ID = "AEROSPACE_WINDOW_ID" // env var
 let AEROSPACE_WORKSPACE = "AEROSPACE_WORKSPACE" // env var
diff --git a/Sources/Common/util/showMessageInGui.swift b/Sources/Common/util/showMessageInGui.swift
index 7ec141ac..91da785f 100644
--- a/Sources/Common/util/showMessageInGui.swift
+++ b/Sources/Common/util/showMessageInGui.swift
@@ -1,5 +1,6 @@
 import Foundation
 
+// todo refactor. showMessageInGui in common code looks weird
 public func showMessageInGui(filenameIfConsoleApp: String, title: String, message: String) {
     let titleAndMessage = "##### \(title) #####\n\n" + message
     if isCli {

```

Sources/AppBundle/ui/MessageView.swift
```diff
@@ -0,0 +1,126 @@
...
+            }
+            .padding()
+            ScrollView {
+                VStack(alignment: .leading) {
+                    HStack {
+                        TextEditor(text: .constant(model.message?.body ?? ""))
+                            .font(.system(size: 12).monospaced())
+                            .focused($focus)
+                            .onExitCommand {
+                                // Escape to remove focus from the TextEditor, and then you can hit Return to trigger the Close, or use 'CMD + ,' for open config and 'CMD + R' for reload config (same as the menu shortcuts.
```
nikitabobko, line 51: It's cool that cmd+r and cmd+, now work in this window. Though I'd like the buttons in the window to show the hotkeys alongside (similar to how it's done in the menu bar)

Feel free to submit a follow up PR

mobile-ar: Cool. I will take a look, not sure if possible, given that menu bar display that by itself. Will take a look on how this might work for normal buttons.

nikitabobko: > not sure if possible, given that menu bar display that by itself

I think it's fine to write a manual logic for this

mobile-ar: @nikitabobko Which option do you prefer:

1 - Simple inline text (the shortcut with the button text might get a little lost in this case)
simple-inline

2 - GroupBox (a little more complex UI logic, but less cluttered buttons)
group-box

I like option 2 a little bit more. But up to you which one you prefer. Or if you have any other suggestions.

Let me know and I can submit a PR for the option you prefer.

nikitabobko: Feel free to choose yourself

PR #1564, opened 2025-07-17
Store the state (windows, workspaces, layout) on change and restore on startup

grothkopp:
Hello,
it's me again ;)
During development it became extremely annoying to restore my "layouts" manually everytime I had to restart Aerospace after a new compile.
So I vibe-coded* this PR which stores the state in a dotfile and restores it on startup.
It's very helpful to me so I wanted to share it. Again I don't know if this fits in with the overall architecture, but maybe it can become usable (and mergeable) with some changes?

* I know how to code, but I have no experience with swift.


grothkopp:
I've just seen that some tests are failing. I think probably because in Sources/AppBundle/tree/frozen/closedWindowsCache.swift in restoreTreeRecursive I modified a loop to handle missing window IDs?
Anyway before I work on that I'll wait for your feedback if this PR is worth putting more time in.

EmilioAK:
Any updates on this? Would be an extremely useful feature for me as well

michalliu:
https://github.com/nikitabobko/AeroSpace/pull/2046 here is a standalone tool for the same purpose

PR #1571, opened 2025-07-19
Adjustments to the config error window

nikitabobko:
- Hide minimize button
- Enter to close the window even when TextEditor is focused

### DIFF
--- a/Sources/AppBundle/ui/MessageView.swift
+++ b/Sources/AppBundle/ui/MessageView.swift
@@ -47,14 +47,17 @@ public struct MessageView: View {
             ScrollView {
                 VStack(alignment: .leading) {
                     HStack {
-                        TextEditor(text: .constant(model.message?.body ?? ""))
+                        let cancelOnEnterBinding: Binding<String> = Binding(
+                            get: { model.message?.body ?? "" },
+                            set: { newText in
+                                if let prev = model.message?.body.count(where: \.isNewline), newText.count(where: \.isNewline) > prev {
+                                    model.message = nil
+                                }
+                            },
+                        )
+                        TextEditor(text: cancelOnEnterBinding)
                             .font(.system(size: 12).monospaced())
                             .focused($focus)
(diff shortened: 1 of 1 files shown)


Sources/AppBundle/ui/MessageView.swift
```diff
@@ -45,14 +47,17 @@ public struct MessageView: View {
             ScrollView {
                 VStack(alignment: .leading) {
                     HStack {
-                        TextEditor(text: .constant(model.message?.body ?? ""))
+                        let cancelOnEnterBinding: Binding<String> = Binding(
+                            get: { model.message?.body ?? "" },
+                            set: { newText in
+                                if newText.count(where: { $0 == "\n" }) != model.message?.body.count(where: { $0 == "\n" }) {
```
nikitabobko, line 53: Looks ugly, but it's the best I found

nikitabobko:
cc @mobile-ar would be happy if you could leave your review

Sources/AppBundle/ui/MessageView.swift
```diff
@@ -45,14 +47,17 @@ public struct MessageView: View {
             ScrollView {
                 VStack(alignment: .leading) {
                     HStack {
-                        TextEditor(text: .constant(model.message?.body ?? ""))
+                        let cancelOnEnterBinding: Binding<String> = Binding(
+                            get: { model.message?.body ?? "" },
+                            set: { newText in
+                                if let prev = model.message?.body.count(where: \.isNewline), newText.count(where: \.isNewline) > prev {
```
nikitabobko, line 53: Looks ugly, but it's the best I found

mobile-ar: Your option is better than my .onExitCommand. It's a little more complex in code but for sure provides better usability for the final user.
It might be the only option because all .keyPress option are only available in macOS 14.0 or newer.

Issue #1424, opened 2025-05-16
Add `fullscreen --hide-background-windows` option


nikitabobko:
### _

- [x] I read https://github.com/nikitabobko/AeroSpace/issues/947

viky-01:
Hi!  I’d like to work on this issue. It looks interesting and I'd love to contribute. Please let me know if it's okay to proceed or if anyone else is already working on it.

MikaelCarpenter:
I've got a branch up with an implementation for this. I asked some [questions](https://github.com/nikitabobko/AeroSpace/discussions/1416#discussioncomment-13744874) in the discussion thread, but maybe here was more appropriate.

cc: @nikitabobko

PR #1572, opened 2025-07-20
Add --hide-others to fullscreen command and move hideInCorner into layoutRecursive

MikaelCarpenter:
The refactor ended up being larger than I had hoped. Very open to edits for style, and of course any issues I may be overlooking.

**For the actual issue (#1424)**
- Added the `--hide-others` option to `fullscreen`
  - Updated the docs, grammar, and completions
  - Added tests for parsing and the hideOthers option
- Added a new property to `Window`, `shouldHideOthersWhileFullscreen`
- Updates `FullscreenCommand` to set that new property's value
- Updated `layoutRecursive` to call `hideInCorner` on the other windows in the activeWorkspace when there is a fullscreen window with the new property set to `true`.

**Moved the `hideInCorner` functionality into `layoutRecursive`, according to the `todo` on it**
- Moved the function out of `MacWindow` and into `layoutRecursive` as an extension on `Window`
- The logic for it had two references to things specific to `MacWindow`, that I had to generalize to `Window`:
  - take the logic for determining the pixel offset and move it onto `Window` and allow `MacWindow` to have it's override for macOS and Zoom handling
  - similarly, for`isHiddenInCorner` I added methods to set it and clear it but let `MacWindow` preserve it's definition of `prevUnhiddenProportionalPositionInsideWorkspaceRect`. 

**Move the calls to `hideInCorner` and `unhideFromCorner` in `refresh` into `layoutRecursive`, so it can be the sole responsible owner of window positioning and visibility**
- One of the things `refresh` does is calculate the optimal hide corner. I don't have strong opinions on where that logic should live so I left it and instead updated `Workspace` to have a property on it to indicate which corner it's windows should hide in
- It also tried to hide all windows in inactive workspaces, so I just moved that logic into layoutRecursive where we're calling `hideInCorner` from now. Now instead of calling `hideInCorner` I can just call `layoutWorkspace`
- `layoutRecursive` was also already handling unhide logic via it's original `else` branch (now the `.visible` branch), so for `refresh` I was able to replace all calls to `unhideLogic` with `layoutWorkspace` as well. This allowed me to remove that function entirely.
- NOTE: there was mention of a specific ordering to reduce flicker. I tried testing manually to see if I noticed any flickering after this refactor and I don't, but I could be missing something! Please let me know if so

**Cleaning up with window layout logic, because at this point it was handling three window layout operations**
- I abstracted it out into `WindowLayoutMode` to cover the three operations
- Then created `determineWindowLayoutMode` to encapsulate the logic for determining which operation to do

### DIFF
--- a/Sources/AppBundle/layout/layoutRecursive.swift
+++ b/Sources/AppBundle/layout/layoutRecursive.swift
@@ -26,25 +26,25 @@ extension TreeNode {
                     window.lastAppliedLayoutVirtualRect = nil
                     try await window.layoutFloatingWindow(context)
                 }
+                workspace.optimalHideCorner = nil
             case .window(let window):
-                if window.windowId != currentlyManipulatedWithMouseWindowId {
-                    lastAppliedLayoutVirtualRect = virtual
-                    let mostRecentWindow = context.workspace.rootTilingContainer.mostRecentWindowRecursive
-                    let mostRecentWindowIsFullscreen = mostRecentWindow?.isFullscreen ?? false
-                    let othersShouldBeHidden = mostRecentWindow?.shouldHideOthersWhileFullscreen ?? false
+                guard window.windowId != currentlyManipulatedWithMouseWindowId else { return }
 
-                    if window == mostRecentWindow && mostRecentWindowIsFullscreen {
+                lastAppliedLayoutVirtualRect = virtual
+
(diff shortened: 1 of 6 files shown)


JoshuaFurman:
Any reason this was closed before getting merged in? Curious as this is my only pain point with aerospace atm

Issue #1493, opened 2025-06-24
Add an indicator for `volume` command


nikitabobko:
Right now, `volume` just changes the volume without any visual feedback

Here is Raycast indicator for inspiration:

<img width="352" alt="Image" src="https://github.com/user-attachments/assets/90032606-6f53-4f2f-84c0-9e87d3eee1cc" />

mobile-ar:
Hi, it's me again 😄 I have built something like this for the volume visual feedback. I was inspired by the volume indicator from iOS/iPadOS. It will be displayed for 2 seconds when you stop interacting with the volume value will go away (same as the native indicator).
It will be presented in the center right of the active monitor (I feel that position is less intrusive than the native macOS center bottom of the screen.

<img width="98" height="242" alt="Image" src="https://github.com/user-attachments/assets/dc108f25-5abf-4ec3-ab73-5d262de614f9" />

This is how it looks. Currently I'm just displaying this in the main screen. I will like to display this in the active monitor instead, I'm still trying to figure this out.

<img width="1800" height="1169" alt="Image" src="https://github.com/user-attachments/assets/8ca2ed53-2ca6-443f-a152-879a70d8a584" />

This is the position and size in my main MacBook screen (retina 1800x1169).

Let me know your comments on this design, or if you prefer this to be placed in some other place or this control to have a different shape. (Maybe this can get a config later for position on screen in case this is in the middle of some user workflow in that part of the screen).

If you want to check this out I will be happy to create a PR.

nikitabobko:
@mobile-ar you are welcome to proceed

nikitabobko:
Released in 0.20.0-Beta

qaisjp:
This is really neat, thanks!

PR #1607, opened 2025-08-03
Add volume indicator

mobile-ar:
As mentioned in #1493 I added a volume indicator using NSPanel, given that I could not find any equivalent for SwiftUI.

Current implementation will display the volume always in the mainMonitor. I try using the "activeMonitor", it worked for 2 displays side by side, but failed to display correctly in a 3rd monitor in this arrangement:

image

### DIFF
--- a/Sources/AppBundle/ui/VolumeView.swift
+++ b/Sources/AppBundle/ui/VolumeView.swift
@@ -1,0 +1,98 @@
+import AppKit
+import SwiftUI
+
+public class VolumePanel: NSPanel {
+    @MainActor public static var shared: VolumePanel = VolumePanel()
+    private var timer: Timer?
+    private var panelFrame = NSRect(x: 0, y: 0, width: 50, height: 206)
+
+    private init() {
+        super.init(
+            contentRect: panelFrame,
+            styleMask: [.nonactivatingPanel, .borderless, .hudWindow, .utilityWindow],
+            backing: .buffered,
+            defer: false,
+        )
+        self.level = .floating
+        self.collectionBehavior = [.canJoinAllSpaces, .fullScreenAuxiliary]
+        self.isReleasedWhenClosed = false
+        self.hidesOnDeactivate = false
+        self.isMovableByWindowBackground = false
+        self.alphaValue = 1
+        self.hasShadow = true
+        self.backgroundColor = .clear
+    }
+
+    public func update(with volume: Float) {
+        timer?.invalidate()
+        self.contentView?.subviews.removeAll()
+        let hostingView = NSHostingView(rootView: VolumeView(volume: volume))
+        if let contentViewBounds = self.contentView?.bounds {
+            hostingView.frame = contentViewBounds
+        }
+        hostingView.autoresizingMask = [.width, .height]
+        self.contentView?.addSubview(hostingView)
+        panelFrame.origin.x = mainMonitor.width - panelFrame.size.width - 20
+        panelFrame.origin.y = (mainMonitor.height - panelFrame.size.height) / 2
+        self.setFrame(panelFrame, display: true)
+        self.orderFrontRegardless()
+        startTimer()
+    }
+
+    func startTimer() {
+        timer = .scheduledTimer(withTimeInterval: 2, repeats: false) { _ in
+            Task { @MainActor [weak self] in
+                self?.close()
+            }
+        }
+    }
+}
+
+struct VolumeView: View {
+    @State var volume: Float? = nil
+
+    @Environment(\.colorScheme) var colorScheme: ColorScheme
+    private var barColor: Color {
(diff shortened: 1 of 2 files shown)


Sources/AppBundle/ui/VolumeView.swift
```diff
@@ -0,0 +1,98 @@
...
+        self.backgroundColor = .clear
+    }
+
+    public func update(with volume: Float) {
+        timer?.invalidate()
+        self.contentView?.subviews.removeAll()
+        let hostingView = NSHostingView(rootView: VolumeView(volume: volume))
+        if let contentViewBounds = self.contentView?.bounds {
+            hostingView.frame = contentViewBounds
+        }
```
nikitabobko, line 32: Can you please explain what are these 3 lines for? When I comment them out, the indicator is weirdly clipped, and I cannot explain it

image

Sources/AppBundle/ui/VolumeView.swift
```diff
@@ -0,0 +1,98 @@
...
+        self.contentView?.addSubview(hostingView)
+        panelFrame.origin.x = mainMonitor.width - panelFrame.size.width - 20
+        panelFrame.origin.y = (mainMonitor.height - panelFrame.size.height) / 2
+        self.setFrame(panelFrame, display: true)
+        self.orderFrontRegardless()
+        startTimer()
+    }
+
+    func startTimer() {
+        timer = .scheduledTimer(withTimeInterval: 2, repeats: false) { _ in
```
nikitabobko, line 43:
```suggestion
        timer = .scheduledTimer(withTimeInterval: 2 /* seconds */, repeats: false) { _ in
```

Merged with this patch

Sources/AppBundle/ui/VolumeView.swift
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    public func update(with volume: Float) {
+        timer?.invalidate()
+        self.contentView?.subviews.removeAll()
+        let hostingView = NSHostingView(rootView: VolumeView(volume: volume))
+        if let contentViewBounds = self.contentView?.bounds {
+            hostingView.frame = contentViewBounds
+        }
+        hostingView.autoresizingMask = [.width, .height]
```
nikitabobko, line 33: What is this line for? When I comment it out, I see no difference. It has to do something with resizing but even if commented out, the window is non-resizable

nikitabobko:
Thank you, merged. I have a couple of questions in the review, I'd appreciate if you could answer them.

Given you active involvment, I've granted you the write access to the repo and [the permission to create issues](https://github.com/nikitabobko/AeroSpace/commit/27140a65497a32e757d6574d7ff98267e30e1523) (the bot won't close issues if they are created by you).

If you are sure about your change, feel free to push directly to the `main` (for example, this PR would be fine to push directly). If you are not sure, you are welcome to create a PR as usual. Use your common sense to decide when to and when not to push directly. I will still review every commit in the `main`, so if I have comments, I will ping you in the respective commit. Don't be afraid to make mistakes in the `main`, the baseline is to keep the CI green.

One thing that I want to explicitly mention: please write more elaborate commit messages. The baseline is to mention an appropriate github issue (if there is an issue), so that the commit contains all necessary information, and could be properly reviewed. For example, you wrote a perfect PR description here, and IMO it should better go to the commit message (I've amended your commit, and copy-pasted your PR description to the commit).

Also feel free to write to my git author email if you have any questions in private. I guess your git author email is valid, and I can use it as well, right?

nikitabobko:
`aerospace volume mute-toggle` doesn't work as expected, can you please fix it in further commits directly in the `main` branch?

Sources/AppBundle/ui/VolumeView.swift
mobile-ar: This was for getting the bounds of the contentView and setting it as the frame of the hosting view of the SwiftUI view.
I think this was part of my first attempt to set the panel to accommodate the size of the SwiftUI view automatically. This could be replaced with 1 line 
```swift 
hostingView.frame = NSRect(x: 0, y: 0, width: panelFrame.width, height: panelFrame.height)
```
I can add this fix in another commit.

Sources/AppBundle/ui/VolumeView.swift
mobile-ar: Same as this https://github.com/nikitabobko/AeroSpace/pull/1607#discussion_r2281076551 this 4 lines can be replaced with that 1 line to set the frame.
I was setting the autoresizingMask to make the panel autoresize when added to the view hierarchy. But that won't work that way.

mobile-ar:
Yes, I will push a commit with the mute-toggle fix.

Thanks for the extra access granted, I like working in this app, and using it also. 

Yes, feel free to contact me in my git author email, I will do the same if I have any questions.

Issue #206, opened 2024-04-24
Built-in method for resizing or moving the window under the cursor


frantic1048:
Currently, AeroSpace does not offer a built-in method for resizing or moving the window under the cursor. I have attempted the following methods:

1. `defaults write -g NSWindowShouldDragOnGesture YES`: This does not work on macOS 14.4.1 (Sonoma). Additionally, this method does not allow customization of the default modifier key (<kbd>ctrl</kbd> + <kbd>cmd</kbd>).

2. Running Yabai with only mouse features (Yabai has built-in support for resizing and moving windows with modifier+mouse). I understand this might seem a bit ridiculous. It did not work: the windows can be moved or resized by Yabai, but it seems that AeroSpace forces them to remain in their original positions.

   ```sh
      # set mouse interaction modifier key
      yabai -m config mouse_modifier cmd

      # set modifier + left-click drag to move window
      yabai -m config mouse_action1 move

      # set modifier + right-click drag to resize window
      yabai -m config mouse_action2 resize

      # disable Yabai window management
      yabai -m rule --add app=".*" manage=off
      ```
3. Trying other tools just for resizing or moving windows, such as [easy-move-resize](https://github.com/dmarcotte/easy-move-resize), it works when AeroSpace is not running. However, when AeroSpace is active, it behaves similarly to the Yabai case.

Will it be possible to add this feature to AeroSpace? 

BTW, as a lost child from i3wm/swaywm to macOS, AeroSpace gives me the most similar experience to i3wm/swaywm. Thanks for your amazing work! :kissing_heart:

nikitabobko:
I agree that it'd be nice to have it

For extensibility purposes, I've been thinking about implementing the "grab and move" and "grab and resize" as separate commands

Brainstorming:
```
[mode.main.binding]
alt-leftMouse = 'grab-and-move --release-on keyup'
alt-rightMouse = 'grab-and-resize --release-on keyup'
```

It'd allow cool things like:
```
[mode.main.binding]
alt-g = 'grab-and-move --release-on alt-g' # Press alt-g once to "grab" window. Press alt-g the second time to release it
```

thiagomeireless:
Are there are news about this feature, maybe even a 'beta' branch? Trying to migrate to Aerospace and really missing this 😅

tymscar:
As an update, `NSWindowShouldDragOnGesture` works just fine on Sequoia.

[...]

Issue #371, opened 2024-08-11
Mouse Events


samholmes:
Events like mouse move or mouse click would be useful to exit modes. My current main mode only has one binding to enter an alt mode with all my single key bindings to navigate windows etc. The space bar is the exit binding from this alt mode. Being able to setup mouse events to also exit this alt mode would be very handy and useful to not accidentally forget that I'm in the alt mode.

nikitabobko:
Something like this could be supported in future versions

```toml
[mode.service.binding]
  leftMouseUp = 'mode main'
  leftMouseDown = 'mode main'
  rightMouseUp = 'mode main'
  rightMouseDown = 'mode main'
```

Things to consider: feature interaction with modifiers

```toml
[mode.main.binding]
  alt-leftMouseDown = 'grab-window on'
  leftMouseUp = 'grab-window off'
```

Hypothetical synopsis:

```
grab-window on --resize
grab-window on --swap [--join-with-modifier=<modifier>]
grab-window on --join-with [--swap-modifier=<modifier>]
grab-window off
```

freisenhauer:
```
[mode.main.binding]
alt-leftMouseDown = 'grab-window on'
alt-leftMouseUp = 'grab-window off'
```

especially this would help me a LOT. I've started a discussions a few days ago which kind of addresses this topic. 
https://github.com/nikitabobko/AeroSpace/discussions/656

freisenhauer:
hey, any updates on this? I would still love to have this feature :(
If I would be able to program swift, I could offer help, but I've never done anything with sift.

PR #1618, opened 2025-08-07
Add command right click functionality for window resizing like in i3wm

tymscar:
screen-recording-pr

This PR adds the right click drag feature from i3wm.
The user should be able to command right-click and drag on any window, and it will resize in the correct direction. If you drag to the left from the left side, you make it bigger, for example.

I have gotten it quite smooth, and as far as I can tell, it actually circumvents the issue that arises when you manually resize a window over another one, and they switch spaces. I can't replicate that with this method. This method runs at the refresh rate of your fastest monitor, or if that fails, I default it to 60fps.

I have also made it so when you hold Command, it doesn't pass through the right-click to the window under it. That was annoying because while dragging, I would by mistake click on Back or something.

Fixes issue https://github.com/nikitabobko/AeroSpace/issues/206


tymscar:
Hello!

Sorry for bothering @nikitabobko, I just wanted to know if theres anything else needed from me for this PR?

Sources/AppBundle/GlobalObserver.swift
```diff
@@ -75,5 +78,55 @@ class GlobalObserver {
...
         }
+
+        let mask = (
+            (1 << CGEventType.rightMouseDown.rawValue) |
+                (1 << CGEventType.rightMouseDragged.rawValue) |
+                (1 << CGEventType.rightMouseUp.rawValue) |
+                (1 << CGEventType.mouseMoved.rawValue) |
+                (1 << CGEventType.flagsChanged.rawValue),
+        )
+        if cmdRightTap == nil,
```
nikitabobko: `GlobalObserver.initObserver` is called only once at startup. Why do you need `cmdRightTap == nil` check?

Sources/AppBundle/mouse/resizeCmdRightDrag.swift
```diff
@@ -0,0 +1,120 @@
...
+    }
+}
+
+@MainActor
+private func preferredFrameDuration() -> Duration {
+    let maxFps = NSScreen.screens.map { $0.maximumFramesPerSecond }.max() ?? 60
+    let fps = max(maxFps, 1)
+    let nanosPerFrame = 1_000_000_000 / fps
+    return .nanoseconds(nanosPerFrame)
+}
```
nikitabobko, line 136: Please take a look at how throttling is implemented in `func movedObs` and do the same.

I think there is no need for `Task.sleep`, and there is no need for complicated frames math

Sources/AppBundle/mouse/resizeCmdRightDrag.swift
```diff
@@ -0,0 +1,120 @@
...
+        .compactMap { node -> TreeNode? in
+            let p = node.parent as? TilingContainer
+            return (p?.orientation == orientation && p?.layout == .tiles) ? node : nil
+        }
+        .forEach { $0.setWeight(orientation, $0.getWeightBeforeResize(orientation) + diff) }
+
+    let sibling = parent.children[neighborIndex]
+    sibling.setWeight(orientation, sibling.getWeightBeforeResize(orientation) - diff)
+
+    currentlyManipulatedWithMouseWindowId = window.windowId
```
nikitabobko: I read this code diagonally, so maybe I'm missing something, but my intuition says that the existing `func resizeWithMouse` can be reused

Sources/AppBundle/GlobalObserver.swift
```diff
@@ -75,5 +78,55 @@ class GlobalObserver {
...
+           let tap = CGEvent.tapCreate(
+               tap: .cgSessionEventTap,
+               place: .headInsertEventTap,
+               options: .defaultTap,
+               eventsOfInterest: CGEventMask(mask),
+               callback: { _, type, event, _ in
+                   if !TrayMenuModel.shared.isEnabled { return Unmanaged.passUnretained(event) }
+                   let flags = event.flags
+                   let isCmd = flags.contains(.maskCommand)
```
nikitabobko: The modifier must be configurable in the config

Smth like:
```toml
mouse-resize-modifier = 'alt' # Possible values: cmd, alt, ctrl, shift
```

AeroSpace's default MOD key is `alt`, so that should be the default for resizing as well

Sources/AppBundle/mouse/resizeCmdRightDrag.swift
```diff
@@ -0,0 +1,120 @@
+import AppKit
+import Common
+
+
+private enum ResizeEdge { case left, right, up, down }
```
nikitabobko: Can't `CardinalDirection` just be reused?

Sources/AppBundle/mouse/resizeCmdRightDrag.swift
nikitabobko: You might need to refactor `func resizeWithMouse` to make it more generic to handle both cases, but I don't want these obviously related logics being implemented two times in two different ways

nikitabobko: Your custom logic doesn't handle more complicated cases:

https://github.com/user-attachments/assets/a61a78bb-e322-44ba-8c77-0c4d5b97f82b

nikitabobko (changes requested):
Thanks for the PR. I do want to implement this feature at some point.

I was thinking of more generic mouse events where users could assign command for whatever mouse + keyboard events (as described here: https://github.com/nikitabobko/AeroSpace/issues/371), but we can start small and implement the mouse dragging/moving with the modifier the way you propose it in this PR, and the way it's done in i3

I have left my first iteration review comments. The most major problem is logic duplication. Please share the logic with existing `func resizeWithMouse`. Once/If you fix the comments, we can discuss the PR further

Sources/AppBundle/GlobalObserver.swift
tymscar: Sorry about that. I usually try to code more defensively. I didn't look to deep into how we reload the config, so I was thinking the whole thing might get reinitialised.

Sources/AppBundle/GlobalObserver.swift
tymscar: That makes sense. I personally always prefer command for this sort of things. Thanks for bringing this up!

Sources/AppBundle/mouse/resizeCmdRightDrag.swift
tymscar: Oh yeah, that's so much cleaner!

tymscar:
Thanks so much for the review @nikitabobko !

So I spent quite a while trying all sorts of things out and I think I got something that reduces the duplication as much as I can.

You might know this better, but as far as I can tell, I think the way we resize with the mouse normally and how right-click drag should work is totally different.

In the case of the mouse resize, that's just normal macOS resizing, and then the accessibility framework updates us on what happened so we can sync up the weights. Whereas for the right-click drag one, that's fully custom, which means we have to handle the actual resizing logic. So what I ended up doing was I factored out the weights calculation and cleaned up the logic a bit, and now both methods use that, but I am not married to the idea. I am all ears if you think there's something cleaner.

Thanks for giving me the opportunity to learn more!

tymscar:
Hey @nikitabobko!

I have been using this fork now for four months daily, and I find it super helpful. So I spent some time fixing merge conflicts, rebased onto main, as well as fixed the issue with resizing diagonals you have told me about using that GIF.

jthomaschewski:
@tymscar Thanks a lot for the PR! I’ve been running your branch for a week with no major issues.

Just one thing: Would it be possible to add support for resizing floating windows as well?
Currently it appears to work only for windows within split containers

tymscar:
Thank you @jthomaschewski ! 
 
 I don't really use floating windows, and I will be busy starting from tomorrow until middle of December, but I might take a look at it after that!

tymscar:
Hey @jthomaschewski I got the floating window resize in. Sorry it took this long, life just got in the way. I think it's way more useful than I first thought. Thanks for suggesting it.

Also, @nikitabobko I have now fixed the diagonal resize you have mentioned and I think the logic is also cleaner. Is there anything else you think needs to be done for this to get merged? Thank you so much!

jthomaschewski:
Thanks a lot @tymscar, just tried it out. Works great for me so far!

Out of scope but maybe related: it would be great to be able to hold the same key and move windows with `<mouse-resize-modifier> + left-mouse-btn` (similar to i3/sway). Theoretically, there is a native feature for that via `NSWindowShouldDragOnGesture`, but it doesn't seem to work with AeroSpace (also the native feature doesn't allow customization of the key)

tymscar:
Hey @nikitabobko, gentle ping on this. Still happy to make any changes if needed.

SilvrDuck:
First of all thanks a lot @nikitabobko and @tymscar :) Aerospace with this feature makes me happy to use my mac again!

I've encoutered a kind of minor bug that you will encounter only if you have random floating windows behind your tiling.

Here it is with semi-transparent terminals to help illustrate:

aerospace_drag

  ### Steps to reproduce
  1. Open two windows tiled side by side in a workspace.
  2. Get a third **floating** window and position it behind the first two.
  3. **Cmd + right-click + drag** on the tiled windows, **clicking above the background floating window**.
  
  **Expected:** tiled windows are resized
  **Outcome:** background window is resized/moved
  
### Environment
  | | |
  |---|---|
  | Build | tymscar's AeroSpace fork, **v0.21.1** |
  | Commit | `f365929e60aba7d84dbd27bf28169a3f9c1236c8` |
  | Feature | i3-style Cmd+right-click drag-resize (upstream PR #1618) |
  | Install | release zip → `~/Applications/AeroSpace.app` + `~/.local/bin/aerospace` (not Homebrew) |
  | macOS | 26.5 (build 25F71) |
  | Arch | arm64 (Apple Silicon) |
  | Displays | Replicated on built in monitor as well as on an external one |

PR #1646, opened 2025-08-20
Add notion app id to goodies

z1645444:


nikitabobko:
Thanks, sometimes people submit PRs to add apps that I don't want to endorse. Since I don't want to think about such things, I reframed the list in the `main` to include Apple-only applications.

z1645444:
Totally understand.

Issue #1486, opened 2025-06-23
Add an indication when secure input is on


nikitabobko:
Quite often users get confused that AeroSpace global hotkeys stop working because of macOS Secure Input feature that disallows intercepting global hotkeys

- https://github.com/nikitabobko/AeroSpace/issues?q=is%3Aissue%20state%3Aclosed%20secure%20input
- https://github.com/nikitabobko/AeroSpace/discussions?discussions_q=is%3Aopen+secure+input

It turns out it's possible to know if secure input is on globally:

```swift
import Carbon

print(IsSecureEventInputEnabled())
```

The suggestion is to use the mentioned API to warn users that the secure input is on.

## UI

The exact UI is under the question

Options off the top of my head:

- Menu bar changes from workspace indication to exclamation mark
- Half-transparent overlay/popup

Contexts.app UI for inspiration

<img width="389" alt="Image" src="https://github.com/user-attachments/assets/89efa6f6-fdd8-4cae-88e9-c9f4ee222cd9" />

## Open questions

- Is `CGEvent.tapCreate` API also affected by Secure Input? https://github.com/nikitabobko/AeroSpace/issues/1012

zenangst:
@nikitabobko can confirm that `CGEvent.tapCreate` is also impacted by Secure Input, regardless of how it's configured (`CGEventTapLocation` and `CGEventTapPlacement`).

mobile-ar:
I implemented this using a new MenuBarExtra with .window style and a custom icon generated using SF Symbols option to create custom icon (merged a keyboard icon with the slash icon. I think this reflects that keyboard shortcuts are disabled).

It looks like this

<img width="338" height="333" alt="Image" src="https://github.com/user-attachments/assets/249f28e6-edae-4224-a196-82ce703ccea5" />

It will display along side AeroSpace workspaces menubar. I implemented this way, because I could not find a way to have 1 menubar with 2 different styles (.window and .menu).

I will submit a PR for this soon, I'm having some issues with the icon working on Xcode build but not working on `./run-debug.sh` builds.

@nikitabobko let me know if you prefer some other text to display in this popup to better explain the functionality/issue.

nikitabobko:
I have 3 points about this issue:
1. Do we want to show it as a menubar? The problem with the menubar approach is that if there are too many of them, then they could get behind the notch. What I was thinking of is something like NSPanel as you did with the volume indicator, but I am open to the discussion
[...]

PR #1653, opened 2025-08-25
Add an indication when secure input is on (Issue #1486)

mobile-ar:
Adding SecureInputPanel to display a new NSPanel to inform the user why the shortcuts are not working while Secure Input is enabled in the focused app. - (Issue: https://github.com/nikitabobko/AeroSpace/issues/1486)

Ended up using the same icon that Ghostty uses, that's a default system icon. The panel will be displayed at the bottom right of the screen

The panel has 2 states, by default is minimized 
image

On click it expand to show more information about Secure Input
image

### DIFF
--- a/Sources/AppBundle/layout/refresh.swift
+++ b/Sources/AppBundle/layout/refresh.swift
@@ -38,6 +38,7 @@ func runRefreshSessionBlocking(
             gcMonitors()
 
             updateTrayText()
+            SecureInputPanel.shared.refresh()
             try await normalizeLayoutReason()
             if shouldLayoutWorkspaces { try await layoutWorkspaces() }
         }
@@ -68,6 +69,7 @@ func runSession<T>(
             let focusAfter = focus.windowOrNil
 
             updateTrayText()
+            SecureInputPanel.shared.refresh()
             try await layoutWorkspaces()
             if focusBefore != focusAfter {
                 focusAfter?.nativeFocus() // syncFocusToMacOs
--- a/Sources/AppBundle/ui/SecureInputView.swift
+++ b/Sources/AppBundle/ui/SecureInputView.swift
@@ -5,25 +5,12 @@ import SwiftUI
 private let iconSize = CGSize(width: 50, height: 50)
 private let textSize = CGSize(width: 440, height: 100)
 
-public class SecureInputPanel: NSPanel {
+public class SecureInputPanel: NSPanelHud {
     @MainActor public static var shared: SecureInputPanel = SecureInputPanel()
     private var hostingView = NSHostingView(rootView: SecureInputView())
 
-    private init() {
-        super.init(
-            contentRect: .zero,
-            styleMask: [.nonactivatingPanel, .borderless, .hudWindow, .utilityWindow],
-            backing: .buffered,
-            defer: false,
-        )
-        self.level = .floating
-        self.collectionBehavior = [.canJoinAllSpaces, .fullScreenAuxiliary]
-        self.isReleasedWhenClosed = false
-        self.hidesOnDeactivate = false
-        self.isMovableByWindowBackground = false
-        self.alphaValue = 1
-        self.hasShadow = true
-        self.backgroundColor = .clear
+    override private init() {
+        super.init()
     }
 
     public func refresh() {
--- a/Sources/AppBundle/ui/SecureInputView.swift
+++ b/Sources/AppBundle/ui/SecureInputView.swift
@@ -3,7 +3,7 @@ import Carbon
 import SwiftUI
 
 private let iconSize = CGSize(width: 50, height: 50)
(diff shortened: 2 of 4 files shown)


Sources/AppBundle/ui/SecureInputView.swift
```diff
@@ -0,0 +1,93 @@
...
+            Rectangle()
+                .fill(Color.gray.opacity(isMinimized ? 0.8 : 1.0))
+            if isMinimized {
+                Image(systemName: "lock.shield.fill")
+                    .resizable()
+                    .aspectRatio(contentMode: .fit)
+                    .padding(6)
+            } else {
+                Text("**Secure Input** is active. **Secure Input** is a macOS security feature that prevents applications from reading keyboard events.")
+                    .font(.system(.title3, design: .monospaced))
```
nikitabobko: Why is it monospaced?

Sources/AppBundle/ui/SecureInputView.swift
```diff
@@ -0,0 +1,93 @@
...
+        ZStack(alignment: .center) {
+            Rectangle()
+                .fill(Color.gray.opacity(isMinimized ? 0.8 : 1.0))
+            if isMinimized {
+                Image(systemName: "lock.shield.fill")
+                    .resizable()
+                    .aspectRatio(contentMode: .fit)
+                    .padding(6)
+            } else {
+                Text("**Secure Input** is active. **Secure Input** is a macOS security feature that prevents applications from reading keyboard events.")
```
nikitabobko:
```suggestion
                Text("AeroSpace cannot respond to keyboard shortcuts while **Secure Input** is active. **Secure Input** is a macOS security feature that prevents applications from reading keyboard events.")
```

We should make it clear that the message comes from AeroSpace

Sources/AppBundle/ui/SecureInputView.swift
```diff
@@ -0,0 +1,93 @@
...
+        )
+        self.level = .floating
+        self.collectionBehavior = [.canJoinAllSpaces, .fullScreenAuxiliary]
+        self.isReleasedWhenClosed = false
+        self.hidesOnDeactivate = false
+        self.isMovableByWindowBackground = false
+        self.alphaValue = 1
+        self.hasShadow = true
+        self.backgroundColor = .clear
+    }
```
nikitabobko: Can we deduplicate the code by extracting the common super class for `SecureInputPanel` and `VolumePanel`?

Sources/AppBundle/ui/SecureInputView.swift
```diff
@@ -0,0 +1,93 @@
...
+
+    public func refresh() {
```
nikitabobko, line 17: This function is called many times from the refresh session. I think it makes sense to avoid doing the unnecessary work when:
1. the panel is already shown and secure input is enabled
2. and when the panel is hidden and secure input is disabled.

You need to store some kind of "isVisible" boolean flag in the panel. Or maybe the panel already have the appropriate flag

Sources/AppBundle/layout/refresh.swift
```diff
@@ -68,6 +69,7 @@ func runSession<T>(
             let focusAfter = focus.windowOrNil
 
             updateTrayText()
+            SecureInputPanel.shared.refresh()
```
nikitabobko, line 72: Question. What happens in the following test case:
1. Trigger the secure input and the popup from AeroSpace
2. Disable AeroSpace
3. Unfocus the app that triggered secure input

Does the popup from AeroSpace go away?

nikitabobko (commented):
Thank you, the code looks mostly fine to me, but I have a few questions.

It's also possible to continue the review directly in the `main` if you want to.

Sources/AppBundle/ui/SecureInputView.swift
mobile-ar: I went with monospaced to be easier to read and consistent with the default menu style that it's monospaced. I can make it default system font if you prefer.

Sources/AppBundle/layout/refresh.swift
mobile-ar: It wasn't going away. Now I fixed this to check if the app is disabled it will be removed.

Sources/AppBundle/ui/SecureInputView.swift
mobile-ar: I changed to normal system font

mobile-ar:
> Thank you, the code looks mostly fine to me, but I have a few questions.
> 
> It's also possible to continue the review directly in the `main` if you want to.

I just rebased this branch with main, I have fixed previous comments, let me know if you are ok with me merging this PR

Blackjacx:
WOW this feature is super cool! Thanks so much. I really never knew why Aerospace did not respond. Now figuring this out is super easy 🙏

Issue #1012, opened 2025-01-18
Investigate a possibility to use `CGEvent.tapCreate` for global hotkeys


nikitabobko:
Right now, we use Carbon API. It's a deprecated framework, but since a looot of apps use, I don't think that Apple is gonna drop it in foreseeable future

The alternative API I recently found is 
https://developer.apple.com/documentation/coregraphics/cgevent/1454426-tapcreate
https://forums.developer.apple.com/forums/thread/735223

Copying the response from apple forum in case it gets deleted:
```
I was recently asked a very similar question in a DTS incident. Pasted in below are the relevant snippets from my response.

Share and Enjoy
—
Quinn “The Eskimo!” @ Developer Technical Support @ Apple
let myEmail = "eskimo" + "1" + "@" + "apple.com"

There are two parts to creating an app like this:

  - Ensuring your app runs in the background (A)

  - Monitoring key presses (B)

I’ll cover each in turn.

With regards point A, running in the background, the approach I recommend is to split the core of your app off into a separate app that you then nest within your main app. Set the [LSUIElement property](https://developer.apple.com/documentation/bundleresources/information_property_list/lsuielement) on the nested app, so it doesn’t show up in the Dock or have its own menu bar.

In that nested app, use a [menu bar status item](https://developer.apple.com/documentation/appkit/menus_cursors_and_the_dock#2871629) to present your UI.

Note If you’re using SwiftUI, there’s now a [SwiftUI equivalent](https://developer.apple.com/documentation/swiftui/scenes#creating-a-menu-bar-extra).

In your main app, have a UI to control whether the nested app runs as a login item, that is, runs while the user is logged in. Use the [SMAppService.loginItem(identifier:) method](https://developer.apple.com/documentation/servicemanagement/smappservice/3945411-loginitem) to configure this.

If you need to support older system, use the [SMLoginItemSetEnabled function](https://developer.apple.com/documentation/servicemanagement/1501557-smloginitemsetenabled) instead..

IMPORTANT App Review has specific constraints about the use of login items. See clause 2.4.5(iii) of the [App Store Review Guidelines](https://developer.apple.com/app-store/review/).

With regards B, monitoring key presses, there are a variety of APIs to monitor keyboard events, including:

  - The [CGEventTap subsystme](https://developer.apple.com/documentation/coregraphics/cgevent/1454426-tapcreate)

[...]

PR #1691, opened 2025-09-12
Automatic key-mapping preset based on currently selected keyboard layout

NMAC427:
By default, all key bindings in the AeroSpace config get interpreted based on a qwerty layout. When switching keyboard layouts, this means that AeroSpace's keyboard shortcuts don't adjust based on the new location of the key as defined by the keyboard layout.

This usually isn't a big problem, as `key-mapping.key-notation-to-key-code` can be used to adjust the layout. However, when working with a custom keyboard that is remapped on the firmware level (i.e. anything running ZMK), in conjuncion with a keyboard that is remapped in software (i.e. builtin MacBook keyboard), this results in significant issues.

For example, let's consider someone using the Dvorak layout with the `dvorak` preset configured in AeroSpace. Let's assume they have some command bound to `alt-o`. Due to the `key-notation-to-key-code` mapping, AeroSpace starts listening for events concerning the key code 1 (which on qwerty corresponds to 's').

When they press `alt-o` on the builtin MacBook keyboard, they press the physical `s` key, resulting in the key code 1.
However, when they press `alt-o` on their other keyboard that is remapped to Dvorak in the keyboard firmware, the keyboard sends key code 31 (which on qwerty corresponds to 'o') instead. Thus AeroSpace won't run the command.

## PR checklist

- [x] Each commit must explain what/why/how in its description. https://cbea.ms/git-commit/

### DIFF
--- a/Sources/AppBundle/config/KeycodeToString.swift
+++ b/Sources/AppBundle/config/KeycodeToString.swift
@@ -1,0 +1,113 @@
+import Carbon
+import Foundation
+
+// This file is a Swift translation of the Objective-C code provided by the user.
+// It translates a keycode to a string, taking the current keyboard layout into account.
+func keycodeToString(_ keyCode: UInt16) -> String? {
+    // Handle special keys first.
+    switch Int(keyCode) {
+        case kVK_F1: return "f1"
+        case kVK_F2: return "f2"
+        case kVK_F3: return "f3"
+        case kVK_F4: return "f4"
+        case kVK_F5: return "f5"
+        case kVK_F6: return "f6"
+        case kVK_F7: return "f7"
+        case kVK_F8: return "f8"
+        case kVK_F9: return "f9"
+        case kVK_F10: return "f10"
+        case kVK_F11: return "f11"
+        case kVK_F12: return "f12"
+        case kVK_F13: return "f13"
+        case kVK_F14: return "f14"
+        case kVK_F15: return "f15"
+        case kVK_F16: return "f16"
+        case kVK_F17: return "f17"
+        case kVK_F18: return "f18"
+        case kVK_F19: return "f19"
+        case kVK_Space: return "space"
+        case kVK_Escape: return "esc"
+        case kVK_Delete: return "backspace"
+        case kVK_ForwardDelete: return "forwardDelete"
+        case kVK_LeftArrow: return "left"
+        case kVK_RightArrow: return "right"
+        case kVK_UpArrow: return "up"
+        case kVK_DownArrow: return "down"
+        case kVK_Help: return "help"
+        case kVK_Home: return "home"
+        case kVK_End: return "end"
+        case kVK_PageUp: return "pageUp"
+        case kVK_PageDown: return "pageDown"
+        case kVK_Tab: return "tab"
+        case kVK_Return: return "enter"
+
+        // Keypad
+        case kVK_ANSI_Keypad0: return "keypad0"
+        case kVK_ANSI_Keypad1: return "keypad1"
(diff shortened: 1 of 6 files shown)


Sources/AppBundle/config/KeycodeToString.swift
```diff
@@ -0,0 +1,113 @@
+import Carbon
+import Foundation
+
+// This file is a Swift translation of the Objective-C code provided by the user.
```
nikitabobko, line 4: What is the original obj-c code? Where is it taken from?

Sources/AppBundle/config/keysMap.swift
```diff
@@ -164,6 +165,20 @@ let keyNotationToKeyCode: [String: Key] = [
     "right": .rightArrow,
 ]
 
+private func automaticMap() -> [String: Key] {
```
nikitabobko, line 168: Do I get it correctly that the dictionary you return here depends on the current layout/input-source?

If so, then it's not expected that we fix input-source mapping at the moment of config reload

Sources/AppBundle/config/KeycodeToString.swift
```diff
@@ -0,0 +1,113 @@
...
+        keyCode,
+        UInt16(kUCKeyActionDisplay),
+        0, // No modifiers
+        UInt32(LMGetKbdType()),
+        UInt32(kUCKeyTranslateNoDeadKeysMask),
+        &deadKeyState,
+        chars.count,
+        &actualLength,
+        &chars,
+    )
```
nikitabobko, line 93: I don't mind to have a feature to handle keyboard events based on their keysym/key-notation (rather than keycodes). It's the logic you are trying to achieve here, right?

But I'd prefer explore `CGEvent.tapCreate` API first - #1012

It looks like `CGEvent.tapCreate` can handle keysyms-use-case better without the necessity to reload the config.

Sources/AppBundle/initAppBundle.swift
```diff
@@ -27,6 +28,19 @@ import Foundation
...
+    DistributedNotificationCenter.default.addObserver(
+        forName: .init(kTISNotifySelectedKeyboardInputSourceChanged as String),
+        object: nil,
+        queue: .main,
+    ) { _ in
+        Task { @MainActor in
+            guard config.keyMapping.preset == .automatic else {
+                return
+            }
+            _ = reloadConfig()
```
nikitabobko, line 41: Sporadic config updates when users switch their input source is a no-go

Sources/AppBundle/config/KeycodeToString.swift
nikitabobko: > But I'd prefer explore `CGEvent.tapCreate` API first - [#1012](https://github.com/nikitabobko/AeroSpace/issues/1012)

If you feel adventurous, you can try to give this issue a shot

NMAC427:
@nikitabobko I just hacked together a simple implementation that uses `CGEvent.tapCreate` instead of the Carbon API. It shouldn't be too difficult to get to a clean implementation. However, before I get started, I would like to get your opinion:
Do you foresee any reason why we should keep around any of the keymap functionality? I would propose that we combine the `CGEvent` API with `UCKeyTranslate` to translate keycodes to characters based on the current keyboard layout, and completely removing the keymap functionality. The only downside that I can see is that it would be a breaking config change.

NMAC427:
Should I replace this PR with the one using CGEvent? https://github.com/nikitabobko/AeroSpace/compare/main...NMAC427:AeroSpace:cgevent

NMAC427:
This would also fix #1472

NMAC427:
Closed in favour of #1714

Issue #1687, opened 2025-09-10
Add `%{all}` interpolation variable


nikitabobko:
### Sanity check

- [x] I read https://github.com/nikitabobko/AeroSpace/issues/947

### Body

The proposal:
```shell
$ aerospace list-windows --format '%{all}' --json
[
  {
    "app-bundle-id" : "org.alacritty",
    "app-bundle-path" : "/Applications/Alacritty.app",
    "app-exec-path" : "/Applications/Alacritty.app/Contents/MacOS/alacritty",
    "app-name" : "Alacritty",
    "app-pid" : 15052,
    ...
```

Motivation: Exploring all possible variables in interactive way

Requirements: `%{all}` requires `--json` flag

BenjaminBini:
Hi,

Was my implementation bad?

Or should I add doc and tests to it?

PR #1708, opened 2025-09-19
Implement #1687 - Add %{all} format variables to lists queries

BenjaminBini:
Add %{all} format variable support to commands listing Aerospace objects :
* `list-windows`
* `list-workspaces`
* `list-apps`
* `list-monitors`

Includes relevant tests and doc.

More details in commit description.


BenjaminBini:
Rebased on main

Issue #1161, opened 2025-03-02
Add `%{window-parent-container-layout}` and `%{workspace-root-container-layout}` variables


nikitabobko:
`%{window-parent-container-layout}` (and `%{window-layout}` alias) should be added to:
- `list-windows`

`%{workspace-root-container-layout}` should be added to:
- `list-workspaces`
- `list-windows`

Example:
```
$ aerospace list-windows --focused --format '%{window-layout}'
h_tiles
```

nikitabobko:
Alternatives: `%{window-parent-container-layout}`/``%{window-parent-layout}``

Pros:
- It's probably more clear. But man, come on, what else does `%{window-layout}` mean?

Cons:
- More verbose

nikitabobko:
I made my mind. The names should be: `%{window-parent-container-layout}` (with the `%{window-layout}` alias) and `%{workspace-root-container-layout}`

alewis001:
Hi. 

Based on [this discussion](https://github.com/nikitabobko/AeroSpace/discussions/1686) I've started to have a go at this change. ([the branch is here](https://github.com/alewis001/AeroSpace/tree/1161-layout-variables))

What should the possible values for `%{window-parent-container-layout}` be? 

I currently have:
* `h_tiles`
* `v_tiles`
* `h_accordion`
* `v_accordion`
* `floating`

alewis001:
I think I now have both the window's parent and workspace layout output working [here](https://github.com/nikitabobko/AeroSpace/compare/main...alewis001:AeroSpace:1161-layout-variables).

@nikitabobko Before I embark on the docs and tests, I'd like to make sure I'm on the right track code wise. Would it be possible for you to have a quick look my code and confirm whether or not I'm at least mostly in the right direction? If I am, I'll plough on with the docs and tests, otherwise I'll sort out the code based on your feedback. 

If that's ok, how would you prefer to do that? Do you want me to open a PR knowing that the docs, tests, etc. are missing at the moment or, would you prefer to eyeball my branch and I follow up with a PR later on once I've done docs, tests, etc?

nikitabobko:
Implemented in 0.20.0-Beta

PR #1709, opened 2025-09-19
#1161 Window and workspace layout output variables

alewis001:
Adds the following format output variables:

list-windows:
* %{window-parent-container-layout} (alias %{window-layout})
* %{workspace-root-container-layout}

list-workspaces:
* %{workspace-root-container-layout}

## PR checklist

- [x] Each commit must explain what/why/how in its description. https://cbea.ms/git-commit/
- [ ] The GitHub Actions CI must pass (you can fix failures after submitting a PR).

### DIFF
--- a/Sources/AppBundle/command/format.swift
+++ b/Sources/AppBundle/command/format.swift
@@ -1,5 +1,14 @@
 import Common
 
+private func toLayoutString(tc: TilingContainer) -> String {
+    switch (tc.layout, tc.orientation) {
+        case (.tiles, .h): return LayoutCmdArgs.LayoutDescription.h_tiles.rawValue
+        case (.tiles, .v): return LayoutCmdArgs.LayoutDescription.v_tiles.rawValue
+        case (.accordion, .h): return LayoutCmdArgs.LayoutDescription.h_accordion.rawValue
+        case (.accordion, .v): return LayoutCmdArgs.LayoutDescription.v_accordion.rawValue
+    }
+}
+
 enum AeroObj {
     case window(window: Window, title: String)
     case workspace(Workspace)
@@ -124,7 +133,7 @@ extension String {
                     case .windowLayout,
                          .windowParentContainerLayout:
                         switch w.parent?.nodeCases {
-                            case .tilingContainer(let tc): .success(.string(self.toLayoutString(tc: tc)))
+                            case .tilingContainer(let tc): .success(.string(toLayoutString(tc: tc)))
                             default: .success(.string("floating"))
                         }
                 }
@@ -133,7 +142,7 @@ extension String {
                     case .workspaceName: .success(.string(w.name))
                     case .workspaceVisible: .success(.bool(w.isVisible))
                     case .workspaceFocused: .success(.bool(focus.workspace == w))
-                    case .workspaceRootContainerLayout: .success(.string(self.toLayoutString(tc: w.rootTilingContainer)))
+                    case .workspaceRootContainerLayout: .success(.string(toLayoutString(tc: w.rootTilingContainer)))
                 }
             case (.monitor(let m), .monitor(let f)):
                 return switch f {
@@ -163,14 +172,5 @@ extension String {
             ?? FormatVar.WorkspaceFormatVar(rawValue: self).flatMap(FormatVar.workspace)
             ?? FormatVar.AppFormatVar(rawValue: self).flatMap(FormatVar.app)
             ?? FormatVar.MonitorFormatVar(rawValue: self).flatMap(FormatVar.monitor)
-    }
-
-    private func toLayoutString(tc: TilingContainer) -> String {
-        switch (tc.layout, tc.orientation) {
-            case (.tiles, .h): return "h_tiles"
-            case (.tiles, .v): return "v_tiles"
-            case (.accordion, .h): return "h_accordion"
-            case (.accordion, .v): return "v_accordion"
-        }
     }
 }
--- a/Sources/AppBundle/command/format.swift
+++ b/Sources/AppBundle/command/format.swift
@@ -121,12 +121,19 @@ extension String {
                     case .windowId: .success(.uint32(w.windowId))
                     case .windowIsFullscreen: .success(.bool(w.isFullscreen))
                     case .windowTitle: .success(.string(title))
+                    case .windowLayout,
+                         .windowParentContainerLayout:
+                        switch w.parent?.nodeCases {
+                            case .tilingContainer(let tc): .success(.string(self.toLayoutString(tc: tc)))
+                            default: .success(.string("floating"))
+                        }
                 }
             case (.workspace(let w), .workspace(let f)):
                 return switch f {
                     case .workspaceName: .success(.string(w.name))
                     case .workspaceVisible: .success(.bool(w.isVisible))
                     case .workspaceFocused: .success(.bool(focus.workspace == w))
+                    case .workspaceRootContainerLayout: .success(.string(self.toLayoutString(tc: w.rootTilingContainer)))
                 }
             case (.monitor(let m), .monitor(let f)):
                 return switch f {
@@ -156,5 +163,14 @@ extension String {
             ?? FormatVar.WorkspaceFormatVar(rawValue: self).flatMap(FormatVar.workspace)
             ?? FormatVar.AppFormatVar(rawValue: self).flatMap(FormatVar.app)
             ?? FormatVar.MonitorFormatVar(rawValue: self).flatMap(FormatVar.monitor)
+    }
+
+    private func toLayoutString(tc: TilingContainer) -> String {
+        switch (tc.layout, tc.orientation) {
+            case (.tiles, .h): return "h_tiles"
+            case (.tiles, .v): return "v_tiles"
+            case (.accordion, .h): return "h_accordion"
+            case (.accordion, .v): return "v_accordion"
+        }
     }
 }
--- a/Sources/AppBundle/command/format.swift
+++ b/Sources/AppBundle/command/format.swift
@@ -9,6 +9,21 @@ private func toLayoutString(tc: TilingContainer) -> String {
     }
 }
 
+@MainActor
+private func toLayoutResult(w: Window) -> Result<Primitive, String> {
+    guard let parent = w.parent else { die("Window must have a parent") }
+    return switch getChildParentRelation(child: w, parent: parent) {
+        case .tiling(let tc): .success(.string(toLayoutString(tc: tc)))
+        case .rootTilingContainer: .success(.string(toLayoutString(tc: w.nodeWorkspace!.rootTilingContainer)))
+        case .floatingWindow: .success(.string("floating"))
+        case .macosNativeFullscreenWindow: .success(.string("macos_native_fullscreen"))
+        case .macosNativeHiddenAppWindow: .success(.string("macos_native_window_of_hidden_app"))
+        case .macosNativeMinimizedWindow: .success(.string("macos_native_minimized"))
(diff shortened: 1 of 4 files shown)


docs/aerospace-list-workspaces.adoc
```diff
@@ -68,6 +68,7 @@ The following variables can be used inside `<output-format>`:
 %{workspace}:: String. Name of the belonging workspace
 %{workspace-is-focused}:: Boolean. True if the workspace has focus
 %{workspace-is-visible}:: Boolean. True if the workspace is visible. A workspace can be visible but not focused in a multi-monitor setup
+%{workspace-root-container-layout}:: String. The layout (`v_tiles`, `h_tiles`, `v_accordion`, `h_accordion`) of the workspace's window container
```
nikitabobko:
```suggestion
%{workspace-root-container-layout}:: String. The layout (`v_tiles`, `h_tiles`, `v_accordion`, `h_accordion`) of the workspace's root container
```

Sources/AppBundle/command/format.swift
```diff
@@ -157,4 +164,13 @@ extension String {
             ?? FormatVar.AppFormatVar(rawValue: self).flatMap(FormatVar.app)
             ?? FormatVar.MonitorFormatVar(rawValue: self).flatMap(FormatVar.monitor)
     }
+
+    private func toLayoutString(tc: TilingContainer) -> String {
```
nikitabobko: Please don't nest this function inside `extension String`, put the function on the top-level. You don't use `self`

Sources/AppBundle/command/format.swift
```diff
@@ -121,12 +121,19 @@ extension String {
                     case .windowId: .success(.uint32(w.windowId))
                     case .windowIsFullscreen: .success(.bool(w.isFullscreen))
                     case .windowTitle: .success(.string(title))
+                    case .windowLayout,
+                         .windowParentContainerLayout:
+                        switch w.parent?.nodeCases {
+                            case .tilingContainer(let tc): .success(.string(self.toLayoutString(tc: tc)))
+                            default: .success(.string("floating"))
+                        }
```
nikitabobko: Please match exhaustively. Don't use `default:`

You are missing other cases when the window is macos_hidden, macos_fullscreen, etc.

I suggest using the following values for these windows:
- `macos_native_fullscreen`
- `macos_native_minimized`
- `macos_native_window_of_hidden_app`
- `NULL-WINDOW-LAYOUT` for `MacosPopupWindowsContainer`. (`MacosPopupWindowsContainer` is a /dev/null container for all weird accessibility objects that are not windows and they noramlly shouldn't appear in `list-windows`)

Sources/AppBundle/command/format.swift
```diff
@@ -157,4 +164,13 @@ extension String {
...
             ?? FormatVar.MonitorFormatVar(rawValue: self).flatMap(FormatVar.monitor)
     }
+
+    private func toLayoutString(tc: TilingContainer) -> String {
+        switch (tc.layout, tc.orientation) {
+            case (.tiles, .h): return "h_tiles"
+            case (.tiles, .v): return "v_tiles"
+            case (.accordion, .h): return "h_accordion"
+            case (.accordion, .v): return "v_accordion"
+        }
```
nikitabobko:
```suggestion
        switch (tc.layout, tc.orientation) {
            case (.tiles, .h): return LayoutCmdArgs.LayoutDescription.h_tiles.rawValue
            case (.tiles, .v): return LayoutCmdArgs.LayoutDescription.v_tiles.rawValue
            case (.accordion, .h): return LayoutCmdArgs.LayoutDescription.h_accordion.rawValue
            case (.accordion, .v): return LayoutCmdArgs.LayoutDescription.v_accordion.rawValue
        }
```

nikitabobko (changes requested):

Sources/AppBundle/command/format.swift
alewis001: Is having the function in `format.swift` ok or is there a better place for it to go?

Sources/AppBundle/command/format.swift
alewis001: Thanks. I'd wondered if there was  a way to do this instead of having the strings.

Sources/AppBundle/command/format.swift
alewis001: I think I may not have a full understanding of the parent child relationship in all cases.

I've added the options for the macos_native and null window layout as you've suggested.

That leaves `window`, `workspace` and `none`. 

Can a Window have another Window as its parent? If so, what should the value be in that case? 

Is `none` an error scenario that should use `.failure` to indicate the parent layout can't be determined or, can this happen and what should the value be?

Where I've attempted to map all cases, I don't see a case where a window is considered `floating`. If my understanding is correct, when a Window is floating its parent is the workspace. In which case the workspace has a tiling layout. I'll do a bit more digging in case I'm missing something.

As I say, I might not have a the right understanding and hence may be asking stupid questions :)

nikitabobko: Yeah, sorry, you better use `getChildParentRelation` function for that, the function describes all possible relations

Anticipating your question about `shimContainerRelation` - use `.failure` for that. `shimContainerRelation` is not possible in your case

> Can a Window have another Window as its parent? 

No, it's not possible. `getChildParentRelation` will cover this case for you.

> Is `none` an error scenario that should use `.failure` to indicate the parent layout can't be determined or, can this happen and what should the value be?

Yes, you can use `.failure` for this case as it should not happen.

And it's better to use `nil` instead of `none`. It's a small Swift weirdness.

> If my understanding is correct, when a Window is floating its parent is the workspace

Correct, but you better use `getChildParentRelation` function for that

Sources/AppBundle/command/format.swift
nikitabobko: `format.swift` is ok

alewis001: Ok, thanks

docs/aerospace-list-workspaces.adoc
alewis001: I've made the suggested change

Sources/AppBundle/command/format.swift
alewis001: I've made the suggested change.

Sources/AppBundle/command/format.swift
alewis001: I've made the suggested change.

Sources/AppBundle/command/format.swift
alewis001: I think/hope I've made the necessary changes. I'm not convinced by my implementation of the `.rootTilingContainer` case by using `!` so I certainly won't be surprised if you want me to take a different route. 

I'm also not sure whether my failure message for the shim case is accurate.

Hopefully the rest is ok but I'm happy to follow whatever changes you want me to make.

nikitabobko:
Merged manually with the following patch on top:

```diff
diff --git Sources/AppBundle/command/format.swift Sources/AppBundle/command/format.swift
index 6e613117..7d31ab7d 100644
--- Sources/AppBundle/command/format.swift
+++ Sources/AppBundle/command/format.swift
@@ -1,29 +1,5 @@
 import Common
 
-private func toLayoutString(tc: TilingContainer) -> String {
-    switch (tc.layout, tc.orientation) {
-        case (.tiles, .h): return LayoutCmdArgs.LayoutDescription.h_tiles.rawValue
-        case (.tiles, .v): return LayoutCmdArgs.LayoutDescription.v_tiles.rawValue
-        case (.accordion, .h): return LayoutCmdArgs.LayoutDescription.h_accordion.rawValue
-        case (.accordion, .v): return LayoutCmdArgs.LayoutDescription.v_accordion.rawValue
-    }
-}
-
-@MainActor
-private func toLayoutResult(w: Window) -> Result<Primitive, String> {
-    guard let parent = w.parent else { die("Window must have a parent") }
-    return switch getChildParentRelation(child: w, parent: parent) {
-        case .tiling(let tc): .success(.string(toLayoutString(tc: tc)))
-        case .rootTilingContainer: .success(.string(toLayoutString(tc: w.nodeWorkspace!.rootTilingContainer)))
-        case .floatingWindow: .success(.string("floating"))
-        case .macosNativeFullscreenWindow: .success(.string("macos_native_fullscreen"))
-        case .macosNativeHiddenAppWindow: .success(.string("macos_native_window_of_hidden_app"))
-        case .macosNativeMinimizedWindow: .success(.string("macos_native_minimized"))
-        case .macosPopupWindow: .success(.string("NULL-WINDOW-LAYOUT"))
-        case .shimContainerRelation: .failure("Window cannot have a shim container relation")
-    }
-}
-
 enum AeroObj {
     case window(window: Window, title: String)
     case workspace(Workspace)
@@ -184,3 +160,27 @@ extension String {
             ?? FormatVar.MonitorFormatVar(rawValue: self).flatMap(FormatVar.monitor)
     }
 }
+
+private func toLayoutString(tc: TilingContainer) -> String {
+    switch (tc.layout, tc.orientation) {
+        case (.tiles, .h): return LayoutCmdArgs.LayoutDescription.h_tiles.rawValue
+        case (.tiles, .v): return LayoutCmdArgs.LayoutDescription.v_tiles.rawValue
+        case (.accordion, .h): return LayoutCmdArgs.LayoutDescription.h_accordion.rawValue
+        case (.accordion, .v): return LayoutCmdArgs.LayoutDescription.v_accordion.rawValue
+    }
+}
+
+private func toLayoutResult(w: Window) -> Result<Primitive, String> {
+    guard let parent = w.parent else { return .failure("NULL-PARENT") }
+    return switch getChildParentRelation(child: w, parent: parent) {
+        case .tiling(let tc): .success(.string(toLayoutString(tc: tc)))
+        case .floatingWindow: .success(.string(LayoutCmdArgs.LayoutDescription.floating.rawValue))
+        case .macosNativeFullscreenWindow: .success(.string("macos_native_fullscreen"))
+        case .macosNativeHiddenAppWindow: .success(.string("macos_native_window_of_hidden_app"))
+        case .macosNativeMinimizedWindow: .success(.string("macos_native_minimized"))
+        case .macosPopupWindow: .success(.string("NULL-WINDOW-LAYOUT"))
+
+        case .rootTilingContainer: .failure("Not possible")
+        case .shimContainerRelation: .failure("Window cannot have a shim container relation")
+    }
+}
diff --git docs/aerospace-list-windows.adoc docs/aerospace-list-windows.adoc
index 99c00337..7dd4a4f4 100644
--- docs/aerospace-list-windows.adoc
+++ docs/aerospace-list-windows.adoc
@@ -79,7 +79,8 @@ The following variables can be used inside `<output-format>`:
 %{window-id}:: Number. Window unique ID
 %{window-title}:: String. Window title
 %{window-is-fullscreen}:: Boolean. Is window in fullscreen by `aerospace fullscreen` command
-%{window-parent-container-layout} (alias %{window-layout}):: String. The layout (`v_tiles`, `h_tiles`, `v_accordion`, `h_accordion`, `floating`) of the window's parent container.
+%{window-layout}:: String. Alias for `%{window-parent-container-layout}`
+%{window-parent-container-layout}:: String. The layout (`v_tiles`, `h_tiles`, `v_accordion`, `h_accordion`, `floating`) of the window's parent container.
 
 %{app-bundle-id}:: String. Application unique identifier. https://developer.apple.com/documentation/appstoreconnectapi/bundle_ids[Bundle ID]
 %{app-name}:: String. Application name
```

alewis001:
Thanks @nikitabobko! Sorry I missed using `LayoutCmdArgs.LayoutDescription.floating.rawValue` instead of `floating`. I'm happier as well that the `rootTilingContainer` case turned out to be a failure scenario in the end, rather than what I did 😄

Issue #28, opened 2023-12-02
Alt shortcuts blocking accented letters


lknop:
AeroSpace seems to intercept all shortcuts with Option key. However, in some international layouts Option is used to write accented letters (e.g. Ō, Ł). It would be great if there was an option to differentiate between left and right Option key, and match bindings only with one of them, passing the unmatched ones to the OS for default handling.

Note: I am using 0.5 beta.

nikitabobko:
The same issue in Amethyst https://github.com/ianyh/Amethyst/issues/1091

wojciech-kulik:
As a workaround, you can use [skhd](https://github.com/koekeishiya/skhd), a single line in the config:
```
lalt - l : aerospace focus right
```

GraemeARobinson:
@wojciech-kulik, neat solution. Would be good if this was included in aerospace in future though but good to have a workaround

GraemeARobinson:
Unfortunately this workaround doesn't work when secure input is enabled, which is sometimes enforced by running apps 😢

webskydavid:
I also miss this feature. The only solution for me is to disable the AeroSpace when typing special letters.

webskydavid:
Someone wanted to add this keys but it seems it does not work. 
<img width="741" alt="image" src="https://github.com/nikitabobko/AeroSpace/assets/4489798/03be3a13-cb7b-4750-978d-c372d123a706">

lknop:
Due to this issue I stopped using Aerospace (and Mac altogether) for some time. However, I need to swtich to MacOS temporarily and having used i3 for quite some time I decided to figure out the solution. I saw the skhd workaround, but since I am using karabiner with [goku](https://github.com/yqrashawn/GokuRakuJoudo) for my keyboard shortcuts, I will paste my version of the workaround.

`karabiner.edn`
```
:main [
     {:des "Change left_option to 3 modifiers combination, f17 when used alone"
         :rules [[:left_option :!COleft_control nil {:alone [:f17]}]]}
  ]
```

this remaps left option to alt-cmd-ctrl, which can then be used in Aerospace shortcuts, leaving right option as the default modifier for accented letters. I also had to modify other shortcuts from alt to alt-cmd-ctrl but the sample below should be enough to get started.

`.aerospace.toml`
```
alt-cmd-ctrl-1 = 'workspace 1'
alt-cmd-ctrl-2 = 'workspace 2'
alt-cmd-ctrl-3 = 'workspace 3'
[...]

PR #1714, opened 2025-09-21
Transition to `CGEvent.tapCreate` for hotkey bindings (#1012)

NMAC427:
Replace the Carbon based HotKey library with `CGEvent.tapCreate` as layed out in #1012.
As a consequence, we get direct access to all key-down events, and can choose to handle them based on the current modifier state and key code. Therefore, in addition to #1012, this change also addresses all of the following issues / feature requests:

- #28
- #1011
- #1472

The primary user facing changes are as follows:
- Added new modifiers `lcmd`, `rcmd`, `lalt`, `ralt`, `lctrl`, `rctrl`, `lshift`, `rshift`, and `fn`.
- Add `key-mapping.match-key-event-by` to choose between matching events based on key codes or key symbols.

## PR checklist


NMAC427:
@nikitabobko Any chance you could take another look at this? Based on the reactions, it seems like this would be quite a popular feature. I've personally been running custom build of this PR for the last 2 1/2 months, regularly switching keyboard layouts, without any issues.

joao-paulo-santos:
I have to use a mac at work and this would be a game changer for my aerospace experience, getting it a step closer to my hyprland customization at home

nikitabobko:
Previous PR: https://github.com/nikitabobko/AeroSpace/pull/1691

---

1. Please avoid merge commits. Use only rebase and force push.
2. The commits must be atomic. The first commit must be the migration from Carbon API to `CGEvent.tapCreate` and nothing else. Right now, you do the migration, change semantics, add new features, and do breaking changes all together in a single commit.
3. Each commit must have its respective git description.
4. Please also take a look at https://github.com/nikitabobko/AeroSpace/blob/main/.github/pull_request_template.md

Answering questions from your [previous PR](https://github.com/nikitabobko/AeroSpace/pull/1691)

> However, before I get started, I would like to get your opinion: Do you foresee any reason why we should keep around any of the keymap functionality?

Yes, we should keep the keymap functionality. Languages and keyboard layouts are wild beasts full of corner cases and exceptions. I am sure that there are people who prefer the current behavior of matching by keycodes (I am kinda such person as I type in qwerty and cyrillic).

The current default of matching by key codes should stay at least for compatibility reasons.

> This usually isn't a big problem, as `key-mapping.key-notation-to-key-code` can be used to adjust the layout. However, when working with a custom keyboard that is remapped on the firmware level (i.e. anything running ZMK), in conjuncion with a keyboard that is remapped in software (i.e. builtin MacBook keyboard), this results in significant issues.

I find this use case unconvincing. You have to switch input sources when you connect your external keyboard anyway. If it's the only use case that we have on the table, I am hesitant to optimize for it. If somebody has other use cases, they are welcome to describe them.

I like the current algorithm of matching by key codes + manual remapping via `key-mapping.key-notation-to-key-code`, because it covers 99% of use cases, and it offloads the responsibility of wild west layouts onto the user.

I imagine a config option that would allow to match by key symbols instead of key codes:
```toml
[key-mapping]
  # Possible values: (key-codes|key-symbols)
  match-key-event-by = 'key-codes' # The current default
```

but I am still not convinced in its necessity.

Rasalas:
I didn't test this "patch" (and I might not understand the problems), but that AltGr and Alt are the same key stopped me from using this at first. 

I have it running and I cant write the EURO sign or Backslash right now, because it registers as if I'm tapping the ALT key only. 

I can probably change the config (I think I did for @ but I think I installed another app to replace it like descibed in a StackOverflow answer) but it's genuinely annoying.
If this can fix it, I'd be really grateful if @NMAC427 could follow your standards. 

This and Aerospace not splitting the current Part to half (it only seems to make vertical splits(L/R)) are probably the only real problems with Aerospace.   
... and macOS registering/publishing apps positions(I assume) wrong I assume ("crashing" apps break it; cant "reload" a space)

gdlg:
Same issue as @Rasalas. I use a Bépo layout which heavily relies on AltGr for many characters including ones needed for programming. In general, there are many common international layouts which use AltGr: see https://en.wikipedia.org/wiki/AltGr_key for a list. So not being able distinguish between AltGr and Alt is a non-starter for me.

> I like the current algorithm of matching by key codes + manual remapping via key-mapping.key-notation-to-key-code, because it covers 99% of use cases, and it offloads the responsibility of wild west layouts onto the user.

As far as I understand, `key-mapping.key-notation-to-key-code` doesn’t help for this use case because it’s not used for modifiers. Since those layouts are used by default by many countries, I wouldn’t call them wild west and it’s significantly more than just 1% of the use cases.

kfiz:
Have to chime in here. Really like aerospace because it solves window management on a mac for me. I'm using some kind of workaround currently that is really cumbersome where I have to manually press some key combination to disable aerospace from taking over `Alt` and `AltGr` just to be able to type an @ sign or  brackets `{ [` on my MacBook Pro which as you can imagine happens a lot. Would really appreciate to merge/implement this if it solves this problem.

NMAC427:
@nikitabobko Thanks for the feedback. I reworked the PR into three atomic commits, while maintaining the default behaviour of matching based on key codes, but also allowing to switch to symbol based matching using `key-mapping.match-key-event-by` as you proposed.

```toml
[key-mapping]
    # Possible values: (key-code|key-symbol)
    match-key-event-by = 'key-symbol'  # default is 'key-code'
```

---

> I find this use case unconvincing. You have to switch input sources when you connect your external keyboard anyway. 

I use a tool called [autokbisw](https://github.com/ohueter/autokbisw) to automatically switch input sources for my internal and external keyboard.

> If it's the only use case that we have on the table, I am hesitant to optimize for it. If somebody has other use cases, they are welcome to describe them.
> I like the current algorithm of matching by key codes + manual remapping via `key-mapping.key-notation-to-key-code`, because it covers 99% of use cases, and it offloads the responsibility of wild west layouts onto the user.

I understand your concern, but given that the audience of AeroSpace probably consists mostly of enthusiasts, I wouldn't be surprised if a decent fraction use custom keyboards + keyboard layouts for improved ergonomics. This feature has at least been requested before (#1472), and is personally my main motivation for this PR. However, if you want, I can also move the `key-mapping.match-key-event-by` part into a separate PR.

micschr0:
I have the same conflict with my German QWERTZ keymap - can't type @, {, [ without workarounds.
The workarounds felt too hacky or super cumbersome, so I stopped using AeroSpace for now.

I would really appreciate seeing this merged if it enables distinguishing left vs right alt/cmd/ctrl keys for hotkey bindings and would bring me back!

kfiz:
@nikitabobko I've been looking to get builds of this as suggested  here https://github.com/nikitabobko/AeroSpace/blob/main/dev-docs/development.md at https://github.com/nikitabobko/AeroSpace/actions/runs/20724817788 . Are there supposed to be build artifacts somewhere that I could download for testing?

GaetanLepage:
This feature would be extremely valuable!
Is there any build we can download to test it?

Thanks for the work :)

GaetanLepage:
RE: I was able to build it from your branch @NMAC427.
I can confirm that it works flawlessly. Thanks! Can't wait for this to get merged.

kfiz:
> RE: I was able to build it from your branch @NMAC427. I can confirm that it works flawlessly. Thanks! Can't wait for this to get merged.

@GaetanLepage Care to share how you built it. ;) I was trying to build the .app  with nix but didn't succeed in doing so, the cli binary built successfully though. I'm hesitant to pull in all external deps needed by the app by just running the provided build scripts as I ideally would like nix to manage all my deps.

CrazyMocks:
@nikitabobko Can you look one more time at this PR? It would be very useful for plenty of users.

exlee:
I'm yet another person that would like to see this move forward.

It covers gripes I have about aerospace (and these seem to be the only ones):

- AltGr (right now I have "passthrough mode" to type, cumbersome)
- Small shortcut surface (which would be solved by treating LCmd and RCmd differently as well as possibility of Fn key use)

Looking forward to be able to specify LAlt-Arrows to map to something other than RAlt-Arrows.

nulam:
> > RE: I was able to build it from your branch @NMAC427. I can confirm that it works flawlessly. Thanks! Can't wait for this to get merged.
> 
> @GaetanLepage Care to share how you built it. ;) I was trying to build the .app with nix but didn't succeed in doing so, the cli binary built successfully though. I'm hesitant to pull in all external deps needed by the app by just running the provided build scripts as I ideally would like nix to manage all my deps.

Not being able to differentiate left and right modifiers is breaking to my keyboard layout with diacritics, hence I built it myself, can confirm it works just fine. 

I got it running with Claude and I let it write up a [tutorial](https://pastebin.com/XvBikexc) afterwards, no guarantees or anything, but it works on my end.

Please note that this tutorial uses the homebrew folder, which might conflict when upgrading via brew. Either tweak the folders or uninstall the default aerospace first.

Edit: disclaimer - i have found this build of aerospace sometimes puts constant 99% load on my CPU, i haven't yet identified the cause, but it severely harms my battery life :/

LeszekBlazewski:
@nikitabobko Looking forward for this feature as well!

itsfarseen:
> we get direct access to all key-down events (https://github.com/nikitabobko/AeroSpace/pull/1714#issue-3438457130)

Can we use any arbitrary key as modifier with this design?
For example, can I use the `z` key as an additional modifier so that I don't have to touch `opt` or `cmd`?
Logic:
- if no key is pressed between a set timeout (say 100ms), treat it as a `z` keydown. 
- if `z` is pressed and released before 100ms, emit both keydown and keyup for `z`
- else treat it as modifier + key

NMAC427:
> Can we use any arbitrary key as modifier with this design? For example, can I use the `z` key as an additional modifier so that I don't have to touch `opt` or `cmd`? Logic:
> 
> * if no key is pressed between a set timeout (say 100ms), treat it as a `z` keydown.
> * if `z` is pressed and released before 100ms, emit both keydown and keyup for `z`
> * else treat it as modifier + key

As it is currently implemented, this isn't supported. However, you should be able to emulate this behaviour using Karabiner-Elements.

GaetanLepage:
Kind ping about this PR.
Is there any remaining things to address? I can see that the build is failing.
It would be really nice to have it merged.

NMAC427:
> I can see that the build is failing.

As far as I can tell, the failure occured while installing swift, not while actually building the PR.

kfiz:
Just built this on MacOS 26. Works fine.

johnturek:
I audited and validated this branch against current upstream and opened a companion PR with several hardening fixes: https://github.com/NMAC427/AeroSpace/pull/1

It handles event taps disabled by timeout/user input, makes tap-creation failure explicit, restores the missing `binding-triggered` subscription event, preserves generic modifier matching when either or both sides are held, and hardens key-symbol translation (F20 parity, TIS ownership, and missing Unicode layout data). The original three commits plus these fixes pass the current upstream `./test.sh` suite.

Issue #163, opened 2024-02-16
Feature request: auto-reload config


nikitabobko:
When the config changes AeroSpace can detect the file change and auto-reload the config

auto-reload behavior should be possible to disable in the config itself:
```
auto-reload-config = true|false
```

~~`reload-config` command should stop resetting to `main` mode?~~ (done)

Follow up is to make the value default: https://github.com/nikitabobko/AeroSpace/issues/1882

KingSupernova31:
Agreed this would be great, rather frustrating trying to get Aerospace set up when I have to remember to reload after any change I make.

Looking at the linked PR, if listening for file changes is a pain to get working, couldn't aerospace just query the file every 100ms or so to see if it's changed? Slightly worse for performance, but I can't imagine that checking the "last updated" time of a file takes a significant amount of time.

nikitabobko:
Fixed since 0.21.0-Beta

PR #1743, opened 2025-10-13
Auto reload config

bshowell0:
## PR checklist

---

My attempt at solving #163, my first PR to this repo so let me know if I've missed anything obvious 👍

### DIFF
--- a/Sources/AppBundle/config/ConfigFileWatcher.swift
+++ b/Sources/AppBundle/config/ConfigFileWatcher.swift
@@ -51,14 +51,14 @@ class ConfigFileWatcher {
             let paths = unsafeBitCast(eventPaths, to: NSArray.self) as? [String] ?? []
             guard let watchedPath = watcher.watchedFileUrl?.path else { return }
 
-            for i in 0..<numEvents {
+            for i in 0 ..< numEvents {
                 let flags = eventFlags[i]
                 let isFile = (flags & FSEventStreamEventFlags(kFSEventStreamEventFlagItemIsFile)) != 0
                 let isRelevant = (flags & FSEventStreamEventFlags(
                     kFSEventStreamEventFlagItemModified |
-                    kFSEventStreamEventFlagItemRenamed |
-                    kFSEventStreamEventFlagItemCreated |
-                    kFSEventStreamEventFlagItemChangeOwner
+                        kFSEventStreamEventFlagItemRenamed |
+                        kFSEventStreamEventFlagItemCreated |
+                        kFSEventStreamEventFlagItemChangeOwner,
                 )) != 0
 
                 if isFile && isRelevant && paths[i] == watchedPath {


Sources/AppBundle/config/ConfigFileWatcher.swift
```diff
@@ -0,0 +1,159 @@
...
+
+        FSEventStreamSetDispatchQueue(stream, DispatchQueue.main)
+
+        if !FSEventStreamStart(stream) {
+            FSEventStreamInvalidate(stream)
+            FSEventStreamRelease(stream)
+            return
+        }
+
+        eventStream = stream
```
nikitabobko, line 92: It would help me to review this PR if you could share how you wrote this code so that I could follow along.

Did you copy the code from somewhere?
Did you google the API and then followed the docs?
Did you generate it with LLM?
Something else?

Did you verify the code yourself by reading the docs after you wrote it?

Sources/AppBundle/config/ConfigFileWatcher.swift
```diff
@@ -0,0 +1,159 @@
...
+@MainActor
+func startConfigFileWatcher() {
+    if _configFileWatcher == nil {
+        _configFileWatcher = ConfigFileWatcher()
+    }
+    _configFileWatcher?.startWatching(configUrl: configUrl)
+}
+
+@MainActor
+func stopConfigFileWatcher() {
```
nikitabobko, line 157: Unused function

Sources/AppBundle/config/ConfigFileWatcher.swift
```diff
@@ -0,0 +1,159 @@
...
+import CoreServices
+
+@MainActor
+class ConfigFileWatcher {
+    private var eventStream: FSEventStreamRef?
+    private var watchedFileUrl: URL?
+    private var lastModificationDate: Date?
+    private var debounceTask: Task<Void, Never>?
+
+    func startWatching(configUrl: URL) {
```
nikitabobko, line 13: You always pass the same parameter here

Sources/AppBundle/config/ConfigFileWatcher.swift
```diff
@@ -0,0 +1,159 @@
...
+
+@MainActor private var _configFileWatcher: ConfigFileWatcher?
+
+@MainActor
+func startConfigFileWatcher() {
+    if _configFileWatcher == nil {
+        _configFileWatcher = ConfigFileWatcher()
+    }
+    _configFileWatcher?.startWatching(configUrl: configUrl)
+}
```
nikitabobko, line 154:
```suggestion
@MainActor private var _configFileWatcher = ConfigFileWatcher()

@MainActor
func startConfigFileWatcher() {
    _configFileWatcher.startWatching(configUrl: configUrl)
}
```

\+ inline `startConfigFileWatcher` function

Given that you never assign `nil` to the variable, I don't understand what this initialization ceremony is for

Sources/AppBundle/initAppBundle.swift
```diff
@@ -14,6 +14,7 @@ import Foundation
     if !reloadConfig() {
         check(reloadConfig(forceConfigUrl: defaultConfigUrl))
     }
+    startConfigFileWatcher()
```
nikitabobko, line 17: Your just reloaded the config in the lines above. Config reloading starts the file watcher

Why do you need to start it here again?

Sources/AppBundle/config/ConfigFileWatcher.swift
```diff
@@ -0,0 +1,159 @@
...
+class ConfigFileWatcher {
+    private var eventStream: FSEventStreamRef?
+    private var watchedFileUrl: URL?
+    private var lastModificationDate: Date?
+    private var debounceTask: Task<Void, Never>?
+
+    func startWatching(configUrl: URL) {
+        stopWatching()
+
+        guard config.autoReloadConfig else { return }
```
nikitabobko, line 16: Please use plain old `if` where possible. I only use `guard` for `guard let`/`guard case`

\+ other similar `guard` usages in this PR

Sources/AppBundle/config/ConfigFileWatcher.swift
```diff
@@ -0,0 +1,159 @@
...
+            FSEventStreamCreateFlags(kFSEventStreamCreateFlagFileEvents),
+        ) else {
+            return
+        }
+
+        FSEventStreamSetDispatchQueue(stream, DispatchQueue.main)
+
+        if !FSEventStreamStart(stream) {
+            FSEventStreamInvalidate(stream)
+            FSEventStreamRelease(stream)
```
nikitabobko, line 88: Do we really need to manually work with the reference counter here?

I am not a Swift/Apple-runtime expert, but when I did my research last time, I figured that we don't need to manually call `CFRelease`. Swift/runtime does it for us unless we create the Core Foundation object ourself. I wonder if it's the same case with `FSEventStream`

Sources/AppBundle/config/ConfigFileWatcher.swift
```diff
@@ -0,0 +1,159 @@
...
+                if isFile && isRelevant && paths[i] == watchedPath {
+                    // Already on main queue
+                    watcher.handleFileChange()
+                    break
+                }
+            }
+        }
+
+        guard let stream = FSEventStreamCreate(
+            kCFAllocatorDefault,
```
nikitabobko, line 73: According to the docs, one can also pass `null` with the same effect. `null` looks much simplier alternative to me

Sources/AppBundle/config/ConfigFileWatcher.swift
```diff
@@ -0,0 +1,159 @@
+import AppKit
+import Common
+import Foundation
+import CoreServices
+
+@MainActor
+class ConfigFileWatcher {
+    private var eventStream: FSEventStreamRef?
+    private var watchedFileUrl: URL?
+    private var lastModificationDate: Date?
```
nikitabobko, line 10: It looks like these two properties can be passed as simple parameter or simply captured by closures they are used in. No need to be "global" class state

Sources/AppBundle/config/ConfigFileWatcher.swift
```diff
@@ -0,0 +1,159 @@
...
+        let callback: FSEventStreamCallback = { (
+            streamRef: ConstFSEventStreamRef,
+            clientCallBackInfo: UnsafeMutableRawPointer?,
+            numEvents: Int,
+            eventPaths: UnsafeMutableRawPointer,
+            eventFlags: UnsafePointer<FSEventStreamEventFlags>,
+            eventIds: UnsafePointer<FSEventStreamEventId>,
+        ) in
+            guard let info = clientCallBackInfo else { return }
+            let watcher = Unmanaged<ConfigFileWatcher>.fromOpaque(info).takeUnretainedValue()
```
nikitabobko, line 48: I'd prefer to use clojure capturing over `clientCallBackInfo: UnsafeMutableRawPointer` if possible

Sources/AppBundle/config/ConfigFileWatcher.swift
```diff
@@ -0,0 +1,159 @@
...
+            guard let watchedPath = watcher.watchedFileUrl?.path else { return }
+
+            for i in 0 ..< numEvents {
+                let flags = eventFlags[i]
+                let isFile = (flags & FSEventStreamEventFlags(kFSEventStreamEventFlagItemIsFile)) != 0
+                let isRelevant = (flags & FSEventStreamEventFlags(
+                    kFSEventStreamEventFlagItemModified |
+                        kFSEventStreamEventFlagItemRenamed |
+                        kFSEventStreamEventFlagItemCreated |
+                        kFSEventStreamEventFlagItemChangeOwner,
```
nikitabobko, line 61: Why do we need to track owner change?

Sources/AppBundle/config/ConfigFileWatcher.swift
nikitabobko, line 13: Does this logic handle the cases when the config changes its location? We have several possible locations: `~/.aerospace.toml`, `~/.config/aerospace/aerospace.toml`

Sources/AppBundle/config/ConfigFileWatcher.swift
```diff
@@ -0,0 +1,159 @@
...
+
+        watchedFileUrl = configUrl
+
+        if let attrs = try? FileManager.default.attributesOfItem(atPath: configUrl.path),
+           let modDate = attrs[.modificationDate] as? Date
+        {
+            lastModificationDate = modDate
+        }
+
+        let directoryPath = configUrl.deletingLastPathComponent().path
```
nikitabobko, line 28: Why do we watch the config directory path, and not the config itself?

Sources/AppBundle/config/ConfigFileWatcher.swift
```diff
@@ -0,0 +1,159 @@
...
+            eventIds: UnsafePointer<FSEventStreamEventId>,
+        ) in
+            guard let info = clientCallBackInfo else { return }
+            let watcher = Unmanaged<ConfigFileWatcher>.fromOpaque(info).takeUnretainedValue()
+
+            // Filter events to only react to our specific file
+            let paths = unsafeBitCast(eventPaths, to: NSArray.self) as? [String] ?? []
+            guard let watchedPath = watcher.watchedFileUrl?.path else { return }
+
+            for i in 0 ..< numEvents {
```
nikitabobko, line 54:
```suggestion
            check(numEvents == eventFlags.count && numEvents == paths.count)
            for i in 0 ..< numEvents {
```

Sources/AppBundle/config/ConfigFileWatcher.swift
```diff
@@ -0,0 +1,159 @@
...
+                let isFile = (flags & FSEventStreamEventFlags(kFSEventStreamEventFlagItemIsFile)) != 0
+                let isRelevant = (flags & FSEventStreamEventFlags(
+                    kFSEventStreamEventFlagItemModified |
+                        kFSEventStreamEventFlagItemRenamed |
+                        kFSEventStreamEventFlagItemCreated |
+                        kFSEventStreamEventFlagItemChangeOwner,
+                )) != 0
+
+                if isFile && isRelevant && paths[i] == watchedPath {
+                    // Already on main queue
```
nikitabobko, line 65: What does this comment relate to?

Sources/AppBundle/config/ConfigFileWatcher.swift
```diff
@@ -0,0 +1,159 @@
...
+
+        guard let attrs = try? FileManager.default.attributesOfItem(atPath: url.path),
+              let modDate = attrs[.modificationDate] as? Date
+        else {
+            return
+        }
+
+        guard modDate > (lastModificationDate ?? .distantPast) else {
+            return
+        }
```
nikitabobko, line 127: What is this `guard` for? Do we have reasons not to trust the file watcher API?

Sources/AppBundle/config/ConfigFileWatcher.swift
```diff
@@ -0,0 +1,159 @@
...
+
+        lastModificationDate = modDate
+
+        debounceTask?.cancel()
+
+        // Debounce the reload to handle atomic save sequences (write → rename → swap)
```
nikitabobko, line 133: > to handle atomic save sequences (write → rename → swap)

I don't understand what you mean

What is renamed? What is swapped with what?

Sources/AppBundle/config/ConfigFileWatcher.swift
```diff
@@ -0,0 +1,159 @@
...
+        debounceTask = Task { @MainActor in
+            try? await Task.sleep(for: .milliseconds(300))
+            guard !Task.isCancelled else { return }
```
nikitabobko, line 136:
```suggestion
            try await Task.sleep(for: .milliseconds(300))
```

The `try` here is exactly to check for cancellation

Sources/AppBundle/config/ConfigFileWatcher.swift
```diff
@@ -0,0 +1,159 @@
...
+            // Reload config within a session to ensure layout refresh happens
```
nikitabobko, line 137: Useless comment

Sources/AppBundle/config/ConfigFileWatcher.swift
```diff
@@ -0,0 +1,159 @@
...
+import Common
+import Foundation
+import CoreServices
+
+@MainActor
+class ConfigFileWatcher {
+    private var eventStream: FSEventStreamRef?
+    private var watchedFileUrl: URL?
+    private var lastModificationDate: Date?
+    private var debounceTask: Task<Void, Never>?
```
nikitabobko, line 11: Do we need our own debouncer? Isn't the `0.3` that you pass to `FSEventStreamCreate` already sufficient?

zenangst:
@nikitabobko just thought I'd chime in here.
This is what I ended up with in my app Keyboard Cowboy to listen for when things change inside a folder.

Long story short, don't know the implications of differences between using CoreServices contra `DispatchSource`, but it does seem a whole lot leaner to go for the `DispatchSource`.

https://github.com/zenangst/KeyboardCowboy/blob/main/App/Sources/Core/FileWatcher.swift

nikitabobko:
@zenangst thanks for the pointer. I am all in for using simpler and safer APIs.

bshowell0:
I actually tried implementing this with DispatchSource originally, but ran into problems handling editors with atomic writes. Many/most editors nowadays don't write directly to files, instead they write to a temp file and delete/rename the original. Since DispatchSource watches file descriptors, it would "lose" the config if edited with an editor that does this, unless you spawn a new watcher on the new file. This definitely could work, but introduces timing issues and race conditions which I thought would be better to avoid completely. FSEvents uses path-based monitoring instead of file descriptor so handles atomic writes automatically, which is why I switched to using it.

Looking at @zenangst's example, it seems like it uses DispatchSource for monitoring directories but doesn't have to monitor individual files so it doesn't run into the same stale fd issue for that use case. But if anyone notices me misunderstanding that code or any of this please chime in.

Don't have much time to work on this until hopefully this weekend, I will address review comments then, but depending on what you prefer, I can keep the FSEvents implementation (addressing your comments ofc), or switch to DispatchSource and make sure I handle atomic writes manually. 

I wasn't really expecting all of these complexities when I started working on this issue so I appreciate the patience and detailed review comments @nikitabobko :)

nikitabobko:
> I actually tried implementing this with DispatchSource originally, but ran into problems handling editors with atomic writes. [...]

Thanks for the explanation. Please, make sure to mention it somewhere in the implementation as a comment.

I'm fine to use this ugly API if it provides the benefit of watching paths instead of file descriptors.

nikitabobko:
Closing due to inactivity. Feel free to force-push to your branch and ping me, I will reopen the PR. Or submit a new PR

PR #1763, opened 2025-10-22
Prevent auto-switching behavior when selecting open applications (in other spaces)

leviouwendijk:
Fix in focus cache: prevent workspace switching when selecting an application that has open windows in another workspace (allowing application selection while remaining in current workspace)


leviouwendijk:
@nikitabobko may I ask why this was rejected?

Issue #222, opened 2024-05-10
Workspace 10 when using external monitor


megasuperlexa:
Sometimes when I plug in my external monitor, workspace 10 appears. I did not configure it, so it is inaccessible via shortcuts.
<img width="320" alt="image" src="https://github.com/nikitabobko/AeroSpace/assets/24682302/70d1b50f-a12b-489e-8066-1787d851d8b2">

my config:
```toml
# See: https://nikitabobko.github.io/AeroSpace/commands#workspace
alt-1 = 'workspace 1 --auto-back-and-forth'
alt-2 = 'workspace 2 --auto-back-and-forth'
alt-3 = 'workspace 3 --auto-back-and-forth'
alt-4 = 'workspace 4 --auto-back-and-forth'
alt-5 = 'workspace 5 --auto-back-and-forth'
alt-6 = 'workspace 6 --auto-back-and-forth'
alt-7 = 'workspace 7 --auto-back-and-forth'
alt-8 = 'workspace 8 --auto-back-and-forth'
alt-9 = 'workspace 9 --auto-back-and-forth'
alt-0 = 'workspace 0 --auto-back-and-forth'
ctrl-alt-right = 'workspace next --wrap-around'
ctrl-alt-left = 'workspace prev --wrap-around'
```

nikitabobko:
It's by design. `10` is a phantom workspace. As soon as it's no longer visible and doesn't contain windows, it will disappear by itself.

To make 10 "invisible" you can manually move one of your workspaces to the newly connected monitor

AeroSpace uses the "phantom workspace" approach because out of your configured workspaces which particular one should be moved to the newly connected monitor?

megasuperlexa:
in my case there are empty workspaces 4-9, I expected Aerospace to use one of these

nikitabobko:
> one of these

Again, which particular one? And why this particular one but not another? These questions are rhetorical, there is no good answer

I want the workspace to monitor assignment to be explicit and predictable. If I didn't move the workspace to another monitor then it shouldn't move

PR #1846, opened 2025-12-07
Prefer using persistent workspaces as stubs before creating new ones

yzx9:
## Summary

This PR adjusts the `getStubWorkspace` logic to consider `persistent-workspaces` before falling back to creating new ephemeral workspaces. When a monitor needs a stub workspace (for example, after moving a workspace to another monitor), AeroSpace will now try to reuse a suitable persistent workspace if one is available.

And, thanks for you great job!

## Details

Previously, `getStubWorkspace(forPoint:)` used the following strategy:

1. Try to reuse the previously visible workspace for this monitor
2. Otherwise, pick any invisible workspace whose `workspaceMonitor` matches the target monitor.
3. Otherwise, create/select a new workspace with a numeric name (`1`, `2`, …)

This PR inserts an additional step between (2) and (3):

1. ...
2. ...
3. If `persistent-workspaces` is configured, try to pick a suitable persistent workspace as the stub.
4. ...

Ideally, I hoped we could find a workspace that isn’t assigned to any monitor. However, since all workspaces currently have a `workspaceMonitor` (at least assigned to `mainMonitor`), implementing that would require a major refactor — we’d need to make this variable optional and update all related code.

So, I decided to relax the condition: if we can find an invisible persistent workspace, that would still be better than using a numeric one, since I’m only using alpha workspaces.

If such a persistent workspace exists, it is used as the stub workspace for the monitor. If not, the logic falls back to the existing step (3) and creates/selects a new ephemeral numeric workspace as before.

## PR checklist

### DIFF
--- a/Sources/AppBundle/tree/Workspace.swift
+++ b/Sources/AppBundle/tree/Workspace.swift
@@ -24,6 +24,15 @@ private func getStubWorkspace(forPoint point: CGPoint) -> Workspace {
     {
         return candidate
     }
+    if !config.persistentWorkspaces.isEmpty {
+        if let persistentCandidate = Workspace.all.first(where: { workspace in
+            config.persistentWorkspaces.contains(workspace.name)
+                && !workspace.isVisible
+                && workspace.forceAssignedMonitor == nil
+        }) {
+            return persistentCandidate
+        }
+    }
     return (1 ... Int.max).lazy
         .map { Workspace.get(byName: String($0)) }
         .first { $0.isEffectivelyEmpty && !$0.isVisible && !config.persistentWorkspaces.contains($0.name) && $0.forceAssignedMonitor == nil }


yzx9:
I just realize that this condition may never be satisfactory, since every workspace will have a workspaceMonitor, and such a workspace would fall into case 2.

**Update**: Ideally, I hoped we could find a workspace that isn’t assigned to any monitor. However, since all workspaces currently have a `workspaceMonitor` (at least assigned to `mainMonitor`), implementing that would require a major refactor — we’d need to make this variable optional and update all related code.
So, I decided to relax the condition: if we can find an invisible persistent workspace, that would still be better than using a numeric one, since I’m only using alpha workspaces.

nikitabobko:
Why do you think that persistent workspaces should be preferred over phantom ones? If your answer is because you couldn't focus phantom workspaces with keyboard, then why do you do want to focus phantom workspaces?

Those are genuine questions, please try to answer them with the method of [Five whys](https://en.wikipedia.org/wiki/Five_whys)

Also see: https://github.com/nikitabobko/AeroSpace/issues/222 and its linked issues

> Ideally, I hoped we could find a workspace that isn’t assigned to any monitor. However, since all workspaces currently have a `workspaceMonitor` (at least assigned to `mainMonitor`), implementing that would require a major refactor — we’d need to make this variable optional and update all related code.

Yes, it's by design. Every persistent workspace has a monitor assigned to it. If it wasn't true, workspaces would randomly jump between monitors, so it's unclear to me what was the exact idea behind your refactoring.

> I’m only using alpha workspaces.

Just to elaborate: you mean single letter workspaces, right?

> - [ ] Each commit must explain what/why/how and motivation in its description. https://cbea.ms/git-commit/ 
> - [ ]  Explain your changes in the relevant commit messages rather than in the PR description. The PR description must not contain more information than the commit messages (except for images and other media).

You don't follow the checklist

---

I am closing and locking the PR for now, please keep the discussion in: https://github.com/nikitabobko/AeroSpace/issues/222

PR #1858, opened 2025-12-18
Auto reload config

### LINKED ISSUE #163: Feature request: auto-reload config

bshowell0:
## PR checklist

Finally got back around to this, sorry for the delay. [Here is the original PR](https://github.com/nikitabobko/AeroSpace/pull/1743) if you want to refresh your memory of the discussions we were having then. I started over completely, retrying with DispatchSource and this time I found the issue I couldn't solve before, I have my config as a symlink so when it was watching the symlink, it wasn't being changed of course. So now I watch whatever the real location of the active config is.

Closes #163

### DIFF
--- a/Sources/AppBundle/config/ConfigFileWatcher.swift
+++ b/Sources/AppBundle/config/ConfigFileWatcher.swift
@@ -1,143 +1,74 @@
 import Common
 import Foundation
 
-final class ConfigFileWatcher: @unchecked Sendable {
-    @MainActor static let shared = ConfigFileWatcher()
+final class ConfigFileWatcher {
+    private let source: DispatchSourceFileSystemObject
+    private let fd: Int32
 
-    private var dispatchSource: DispatchSourceFileSystemObject?
-    private var watchedUrl: URL?
-    private var debounceWorkItem: DispatchWorkItem?
-    private let queue = DispatchQueue(label: "aerospace.config-watcher", qos: .utility)
-
-    private let debounceDelay: TimeInterval = 0.2
-
-    private init() {}
-
-    @MainActor
-    func startWatching(url: URL) {
+    init?(url: URL, onChange: @escaping () -> Void) {
         let resolvedUrl = url.resolvingSymlinksInPath()
-
-        queue.async { [self, resolvedUrl] in
-            if self.watchedUrl == resolvedUrl && self.dispatchSource != nil {
-                return
-            }
-
-            self.stopWatchingInternal()
-
-            let fd = open(resolvedUrl.path, O_EVTONLY)
-            if fd < 0 {
-                return
-            }
-
-            self.watchedUrl = resolvedUrl
-            self.createWatcher(fd: fd)
+        let fd = open(resolvedUrl.path, O_EVTONLY)
+        if fd < 0 {
+            return nil
         }
-    }
+        self.fd = fd
 
-    @MainActor
-    func stopWatching() {
-        queue.async { [self] in
-            self.stopWatchingInternal()
-        }
-    }
-
-    private func stopWatchingInternal() {
-        debounceWorkItem?.cancel()
-        debounceWorkItem = nil
-
-        if let source = dispatchSource {
-            source.cancel()
-            dispatchSource = nil
-        }
-        watchedUrl = nil
-    }
-
-    private func createWatcher(fd: Int32) {
         let source = DispatchSource.makeFileSystemObjectSource(
             fileDescriptor: fd,
             eventMask: [.write, .delete, .rename, .revoke],
-            queue: queue,
+            queue: .main,
         )
+        self.source = source
 
-        source.setEventHandler { [weak self] in
-            guard let self else {
-                return
-            }
-            let flags = source.data
-            self.handleFileEvent(flags: flags)
+        source.setEventHandler {
+            onChange()
         }
 
-        source.setCancelHandler { [weak self] in
-            guard let self else {
-                return
-            }
+        source.setCancelHandler {
             close(fd)
-            if self.dispatchSource === source {
-                self.dispatchSource = nil
-            }
         }
 
-        dispatchSource = source
         source.activate()
     }
 
-    private func handleFileEvent(flags: DispatchSource.FileSystemEvent) {
-        if flags.contains(.delete) || flags.contains(.rename) || flags.contains(.revoke) {
-            handleFileReplacement()
-        } else if flags.contains(.write) {
-            scheduleReload()
-        }
+    func cancel() {
+        source.cancel()
     }
 
-    private func handleFileReplacement() {
-        guard let url = watchedUrl else { return }
-
-        dispatchSource?.cancel()
-        dispatchSource = nil
-
-        debounceWorkItem?.cancel()
-        debounceWorkItem = nil
-
-        let workItem = DispatchWorkItem { [weak self] in
-            guard let self, self.watchedUrl == url else { return }
-
-            let fd = open(url.path, O_EVTONLY)
-            if fd < 0 {
-                self.watchedUrl = nil
-                return
-            }
-
-            self.createWatcher(fd: fd)
-            self.triggerReload()
-        }
-
-        queue.asyncAfter(deadline: .now() + debounceDelay, execute: workItem)
+    deinit {
+        source.cancel()
     }
+}
 
-    private func scheduleReload() {
-        debounceWorkItem?.cancel()
+@MainActor private var currentWatcher: ConfigFileWatcher?
+@MainActor private var debounceTask: Task<Void, Never>?
 
-        let workItem = DispatchWorkItem { [weak self] in
-            self?.triggerReload()
-        }
+private let debounceDelay: Duration = .milliseconds(200)
 
-        debounceWorkItem = workItem
-        queue.asyncAfter(deadline: .now() + debounceDelay, execute: workItem)
-    }
+@MainActor func syncConfigFileWatcher() {
+    currentWatcher?.cancel()
+    currentWatcher = nil
 
-    private func triggerReload() {
+    guard config.autoReloadConfig else { return }
+
+    currentWatcher = ConfigFileWatcher(url: configUrl) {
         Task { @MainActor in
-            if let token: RunSessionGuard = .isServerEnabled {
-                try await runLightSession(.configAutoReload, token) { _ = try await reloadConfig() }
-            }
+            scheduleConfigReload()
         }
     }
 }
 
-@MainActor func syncConfigFileWatcher() {
-    if config.autoReloadConfig {
-        ConfigFileWatcher.shared.startWatching(url: configUrl)
-    } else {
-        ConfigFileWatcher.shared.stopWatching()
+@MainActor private func scheduleConfigReload() {
+    debounceTask?.cancel()
+
+    debounceTask = Task {
+        try? await Task.sleep(for: debounceDelay)
+        guard !Task.isCancelled else { return }
+
+        if let token: RunSessionGuard = .isServerEnabled {
+            try? await runLightSession(.configAutoReload, token) {
+                _ = try await reloadConfig()
+            }
+        }
     }
 }


bshowell0:
I think the failing check is a github issue? Don't think I can manually rerun the check though.

Sources/AppBundle/config/ConfigFileWatcher.swift
```diff
@@ -0,0 +1,142 @@
+import Common
+import Foundation
+
+final class ConfigFileWatcher: @unchecked Sendable {
```
nikitabobko: `@unchecked Sendable` is a no-go.

Internal state management of this class feels too complicated and fragile. I think it can be simplified / be written in immutable style (which would also fix `@unchecked Sendable`)

I realize that this comment of mine is not very direct and might be hard to act on.

Sources/AppBundle/config/ConfigFileWatcher.swift
```diff
@@ -0,0 +1,142 @@
...
+                return
+            }
+            close(fd)
+            if self.dispatchSource === source {
+                self.dispatchSource = nil
+            }
+        }
+
+        dispatchSource = source
+        source.resume()
```
nikitabobko: What's the difference between `source.resume` and `source.activate`?

nikitabobko: I read `makeFileSystemObjectSource` documentation https://developer.apple.com/documentation/dispatch/dispatchsource/makefilesystemobjectsource(filedescriptor:eventmask:queue:)

and it asks to use `activate`. I am fine with `source.resume` but I am curious why you chose it

Sources/AppBundle/config/ConfigFileWatcher.swift
```diff
@@ -0,0 +1,142 @@
...
+        Task { @MainActor in
+            _ = try? await reloadConfig()
+            scheduleRefreshSession(.configAutoReload)
+        }
+    }
+}
+
+@MainActor func syncConfigFileWatcher() {
+    if config.autoReloadConfig {
+        ConfigFileWatcher.shared.startWatching(url: configUrl)
```
nikitabobko: It's incorrect to watch only for `configUrl`

Imagine the following scenario:
1. The user just installs AeroSpace and doesn't yet have a custom config
2. AeroSpace starts watching for `/Applications/AeroSpace.app/Contents/Resources/default-config.toml`
3. User creates the custom config.

Expected: AeroSpace automatically reloads the custom config since the custom config takes precedence
Actual: AeroSpace doesn't automatically reload the config

Sources/AppBundle/config/ConfigFileWatcher.swift
```diff
@@ -0,0 +1,142 @@
...
+
+        debounceWorkItem = workItem
+        queue.asyncAfter(deadline: .now() + debounceDelay, execute: workItem)
+    }
+
+    private func triggerReload() {
+        Task { @MainActor in
+            _ = try? await reloadConfig()
+            scheduleRefreshSession(.configAutoReload)
+        }
```
nikitabobko:
```suggestion
        if let token: RunSessionGuard = .isServerEnabled {
            Task {
                try await runLightSession(.configAutoReload, token) { _ = try await reloadConfig() }
            }
        }
```

Sources/AppBundle/config/ConfigFileWatcher.swift
```diff
@@ -0,0 +1,142 @@
...
+        dispatchSource = nil
+
+        debounceWorkItem?.cancel()
+        debounceWorkItem = nil
+
+        let workItem = DispatchWorkItem { [weak self] in
+            guard let self, self.watchedUrl == url else { return }
+
+            let fd = open(url.path, O_EVTONLY)
+            if fd == -1 {
```
nikitabobko: From `man 2 open`:
```
RETURN VALUES
     If successful, open() returns a non-negative integer, termed a file descriptor.  It returns -1 on failure, and sets errno to indicate the error.
```

The documentation is vague and it's unclear whether the syscall always returns `-1` or whether it returns just any negative number

I'd be extra cautious and check if the number is non-negative

Sources/AppBundle/config/ConfigFileWatcher.swift
```diff
@@ -0,0 +1,142 @@
...
+                self.watchedUrl = nil
+                return
+            }
+
+            self.createWatcher(fd: fd)
+            self.triggerReload()
+        }
+
+        queue.asyncAfter(deadline: .now() + 0.05, execute: workItem)
```
nikitabobko: Why do you use different debounce delays here vs `debounceDelay` in `scheduleReload`?

Sources/AppBundle/config/ConfigFileWatcher.swift
nikitabobko: Since `makeFileSystemObjectSource` can only watch for file descriptors, I don't understand why you migrated from paths (your previous PR) to file descriptors

https://github.com/nikitabobko/AeroSpace/pull/1743#issuecomment-3408721045
https://github.com/nikitabobko/AeroSpace/pull/1743#issuecomment-3418577924

docs/config-examples/default-config.toml
```diff
@@ -12,6 +12,9 @@ after-startup-command = []
 # Start AeroSpace at login
 start-at-login = false
 
+# Automatically reload the config when the config file is changed
+auto-reload-config = false
```
nikitabobko, line 17: I think that the default should be `auto-reload-config = true` + don't forget to change it in `Sources/AppBundle/config/Config.swift`

zhenpewu: I prefer the default to be false, as a user—not only for backward compatibility, but because I think the AeroSpace configuration is quite static and stable to me; it won't change much. I don't think auto-reload is necessary for most of us.

Sources/AppBundle/config/ConfigFileWatcher.swift
bshowell0: I used `.resume()` because that's what was used in the general dispatch source examples I was using as reference, like [this one](https://agostini.tech/2017/08/06/monitoring-files-using-dispatch-sources/). As far as I can tell they do both end up doing the same thing, but looking into it I agree that `.activate()` in swift (I'm fairly new to swift) is the more modern way to do it so I'll switch it to that.

Sources/AppBundle/config/ConfigFileWatcher.swift
bshowell0: This is the reason I had the default value be false rather than true. That way, when they make their config, they will set the value to true if they'd like it, and do one more manual refresh and it will start watching. My general thought was that doing one last manual reload before it automatically reloads is a fine tradeoff for the simplicity of dispatch source over the FSEvents I was using previously.

Since making that PR, I looked into more uses of Dispatch Source vs FSEvents and found that FSEvents is used in projects watching directories, not single files, like [Hammerspoon](https://www.hammerspoon.org/Spoons/ReloadConfiguration.html) which has directories of lua config files rather than a single dotfile like this project. Since one of the valid config directories for aerospace is `~/.aerospace.toml`, we would have to watch the home `~` directory, which is really busy, so in my opinion that isn't something we want to do.

bshowell0: Honestly, that last PR was not as high of quality as I like since I had been frustrated by all of the issues I had been running into while developing it, and I was out of patience by the time I got it working and opened the PR, sorry for that. This file descriptor version in my opinion should be better.

Sources/AppBundle/config/ConfigFileWatcher.swift
bshowell0: This was on purpose but I should have made it clearer, debounceDelay is to reduce a burst of quick write events into a single reload when an editor saves multiple times in a row (editors that do format on save and similar sometimes do this). The 0.05 delay is a tiny grace period to let the atomic rename/replace version of saving settle before we try to re-open the file descriptor, since it doesn't have a scenario with a bunch of writes in a row. Since they were for different things I kept their values different, but it makes more sense to just reuse the variable for both, I'll make that update. If you'd like a different change let me know.

docs/config-examples/default-config.toml
bshowell0: I tend to agree with @zhenpewu here, both for his reasons (backwards compatibility and general unchangingness of mature configs) and for the reasons mentioned in [my above comment](https://github.com/nikitabobko/AeroSpace/pull/1858#discussion_r2634038013). It's a setting I would enable while I was testing out changes to my config, but once I got to a state I like it would stay disabled, and I think most users would have a similar use case.

I'll add a second comment to make the functionality more clear:
`# After setting this to true, reload once manually to start the auto-reloading`

But of course your preference is what matters here so let me know if you still like the default being true and I'll make that change. It just won't be possible to make your flow [from this comment](https://github.com/nikitabobko/AeroSpace/pull/1858#discussion_r2631224447) work with file descriptors.

Sources/AppBundle/config/ConfigFileWatcher.swift
bshowell0: `Main actor-isolated static property 'isServerEnabled' can not be referenced from a nonisolated context`, changed it to:
```swift
private func triggerReload() {
    Task { @MainActor in
        if let token: RunSessionGuard = .isServerEnabled {
            try await runLightSession(.configAutoReload, token) { _ = try await reloadConfig() }
        }
    }
}
```

Sources/AppBundle/config/ConfigFileWatcher.swift
bshowell0: I believe I've rewritten everything in the way you mean. I will note that with this refactor, I used LLMs more heavily than for the original PR, because I've found they do a much better job at refactoring than they do generating new code. But I've reviewed everything and it all makes sense to me (learned new things about swift, I'm more used to kotlin, thank you for your contributions to my favorite language btw). You're right, it's a lot cleaner and simpler this way.

bshowell0:
Hey @nikitabobko, this is ready for re-review, just pinging in case you didn't see.

nikitabobko:
> I believe I've rewritten everything in the way you mean. I will note that with this refactor, I used LLMs more heavily than for the original PR

Thanks, the immutable version looks much cleaner. And thanks for LLM usage notice, I appreciate the honesty. As soon as LLMs are not blindly abused, I am fine with LLM usage.

As of what should be the default value of `auto-reload-config`, I still believe that it should be enabled by default since it's sane default for new users. 

Enabling it by default doesn't break any backwards compatibility (maybe habits, but not compatibility since no manual intervention is required after update), and enabling it by default doesn't bring any major overhead to be concerned of unchangingness of mature configs.

I keep it disabled only because of the technical impossibility to implement the flow from my comment above. And I am open for future PRs to change the default value of `auto-reload-config` https://github.com/nikitabobko/AeroSpace/issues/1882

---

Merged manually with the following patch on top:

```diff
diff --git a/Sources/AppBundle/command/impl/ReloadConfigCommand.swift b/Sources/AppBundle/command/impl/ReloadConfigCommand.swift
index fab80c5e..306547af 100644
--- a/Sources/AppBundle/command/impl/ReloadConfigCommand.swift
+++ b/Sources/AppBundle/command/impl/ReloadConfigCommand.swift
@@ -25,6 +25,7 @@ struct ReloadConfigCommand: Command {
     forceConfigUrl: URL? = nil,
     stdout: inout String,
 ) async throws -> Bool {
+    let result: Bool
     switch readConfig(forceConfigUrl: forceConfigUrl) {
         case .success(let (parsedConfig, url)):
             if !args.dryRun {
@@ -33,10 +34,9 @@ struct ReloadConfigCommand: Command {
                 configUrl = url
                 try await activateMode(activeMode)
                 syncStartAtLogin()
-                syncConfigFileWatcher()
                 MessageModel.shared.message = nil
             }
-            return true
+            result = true
         case .failure(let msg):
             stdout.append(msg)
             if !args.noGui {
@@ -44,6 +44,10 @@ struct ReloadConfigCommand: Command {
                     MessageModel.shared.message = Message(description: "AeroSpace Config Error", body: msg)
                 }
             }
-            return false
+            result = false
     }
+    if !args.dryRun {
+        syncConfigFileWatcher()
+    }
+    return result
 }
diff --git a/Sources/AppBundle/config/ConfigFileWatcher.swift b/Sources/AppBundle/config/ConfigFileWatcher.swift
index 05b359bb..5c9fffca 100644
--- a/Sources/AppBundle/config/ConfigFileWatcher.swift
+++ b/Sources/AppBundle/config/ConfigFileWatcher.swift
@@ -1,73 +1,44 @@
 import Common
 import Foundation
 
-final class ConfigFileWatcher {
+private final class ConfigFileWatcher {
     private let source: DispatchSourceFileSystemObject
     private let fd: Int32
 
-    init?(url: URL, onChange: @escaping () -> Void) {
-        let resolvedUrl = url.resolvingSymlinksInPath()
-        let fd = open(resolvedUrl.path, O_EVTONLY)
-        if fd < 0 {
-            return nil
-        }
-        self.fd = fd
-
-        let source = DispatchSource.makeFileSystemObjectSource(
+    init?(url: URL, onChange: @escaping @MainActor () -> Void) {
+        fd = open(url.path, O_EVTONLY)
+        if fd < 0 { return nil }
+        source = DispatchSource.makeFileSystemObjectSource(
             fileDescriptor: fd,
             eventMask: [.write, .delete, .rename, .revoke],
             queue: .main,
         )
-        self.source = source
-
-        source.setEventHandler {
-            onChange()
-        }
-
-        source.setCancelHandler {
-            close(fd)
-        }
-
+        source.setEventHandler { MainActor.checkIsolated { onChange() } }
+        source.setCancelHandler { [fd] in close(fd) }
         source.activate()
     }
 
-    func cancel() {
-        source.cancel()
-    }
-
     deinit {
         source.cancel()
     }
 }
 
-@MainActor private var currentWatcher: ConfigFileWatcher?
-@MainActor private var debounceTask: Task<Void, Never>?
+@MainActor private var currentWatcher: ConfigFileWatcher? = nil
+@MainActor private var debounceTask: Task<Void, any Error>? = nil
 
 private let debounceDelay: Duration = .milliseconds(200)
 
 @MainActor func syncConfigFileWatcher() {
-    currentWatcher?.cancel()
     currentWatcher = nil
-
-    guard config.autoReloadConfig else { return }
-
+    if !config.autoReloadConfig { return }
     currentWatcher = ConfigFileWatcher(url: configUrl) {
-        Task { @MainActor in
-            scheduleConfigReload()
-        }
-    }
-}
-
-@MainActor private func scheduleConfigReload() {
-    debounceTask?.cancel()
-
-    debounceTask = Task {
-        try? await Task.sleep(for: debounceDelay)
-        guard !Task.isCancelled else { return }
-
-        if let token: RunSessionGuard = .isServerEnabled {
-            try? await runLightSession(.configAutoReload, token) {
-                _ = try await reloadConfig()
+        debounceTask?.cancel()
+        debounceTask = Task {
+            try await Task.sleep(for: debounceDelay)
+            if let token: RunSessionGuard = .isServerEnabled {
+                try await runLightSession(.configAutoReload, token) {
+                    _ = try await reloadConfig()
+                }
             }
         }
     }
diff --git a/Sources/Common/util/MainActorEx.swift b/Sources/Common/util/MainActorEx.swift
new file mode 100644
index 00000000..10394320
--- /dev/null
+++ b/Sources/Common/util/MainActorEx.swift
@@ -0,0 +1,8 @@
+import Foundation
+
+extension MainActor {
+    public static func checkIsolated<T>(_ operation: @MainActor () throws -> T, file: StaticString = #fileID, line: UInt = #line) rethrows -> T where T : Sendable {
+        check(Thread.isMainThread)
+        return try assumeIsolated(operation, file: file, line: line)
+    }
+}
```

The patch is a cleanup, swift-concurrency safety fix and a bug fix in `reloadConfig` (`syncConfigFileWatcher` has to be called even when config reload has failed)

PR #1863, opened 2025-12-21
Add Vim-like focus history navigation (back/forward)

### LINKED ISSUE #1024: Improve `focus-back-and-forth`: remember the whole focus history stack
Excerpt from the docs:

> AeroSpace stores only one previously focused window in history, which means that if you close the previous window, focus-back-and-forth has no window to switch focus to. In that case, the command will exit with non-zero exit code.
> 
> That’s why it may be preferred to combine focus-back-and-forth with workspace-back-and-forth:
> 
> `aerospace focus-back-and-forth || aerospace workspace-back-and-forth`

This is bad and should be improved. We should store the whole focus history stack. Optionally making it available for users via smth like `list-focus-history-stack` (also see: https://github.com/nikitabobko/AeroSpace/discussions/942)

richard134:
Implements Vim-like focus history navigation (focus back/focus forward) and makes focus-back-and-forth resilient to closed windows by using the shared history stack.

Closes #1024

### DIFF
--- a/Sources/AppBundle/command/impl/FocusBackAndForthCommand.swift
+++ b/Sources/AppBundle/command/impl/FocusBackAndForthCommand.swift
@@ -9,7 +9,7 @@ struct FocusBackAndForthCommand: Command {
         if let prevFocus {
             return setFocus(to: prevFocus)
         } else {
-            return io.err("Prev window has been closed")
+            return io.err("No previous window in focus history")
         }
     }
 }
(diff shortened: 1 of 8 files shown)


RohanSreelesh (approved):

I21b:
Tried the test build of 364dd1e, that `back`/`forward` is incompatible with `--boundaries`; I personally am used to alt-tab only in the current workspace...

Also, the current behavior can skip windows: After alt-tab to a window and done some work, I usually want to alt-tab back (I think most of people are like this), but the history isn't logging windows that are alt-tabbed away, it's more frustrating that sometimes this can cause an extra workspace switch;
And here's another example of skipping windows:

When having 3 windows (A, B, C, opened in order) on the current workspace, do `focus back` once and click on A, now do `focus back` will focus B, A, and touch the beginning or focus older windows; in this case, C is totally skipped.

And currently the history can log the same window multiple times, that isn't intended for some people.

PR #1894, opened 2026-01-07
fix: preserve window position and size when exiting macOS native fullscreen

### LINKED ISSUE #562: Preserve position for native fullscreen window after exiting
I built a snapshot from commit https://github.com/nikitabobko/AeroSpace/commit/0589ac7d3071042e0c17e80dd517a0bcdbf798e9 and noticed that when entering native fullscreen, the window is detached and when exiting native fullscreen, the window is reattached to the last focused window and does not return to the last known position.

I found [this](https://github.com/nikitabobko/AeroSpace/issues/188#issuecomment-2015684656) where @nikitabobko said this is the intended behaviour, but also [this](https://github.com/nikitabobko/AeroSpace/issues/18#issuecomment-2310013237) that at least the behaviour for hidden windows should be changed to preserve the position in the tree.

This is more of a feature request to rethink the native fullscreen behavior. At least for me, it disturbed my workflow, as I have an ultra-wide monitor with multiple browser instances in a horizontal accordion container and keep my IDE next to it on the same workspace, and switching back and forth between native fullscreen in a e.g. Youtube video completely destroys my layout.

```
aerospace CLI client version: 0.0.0-SNAPSHOT 0589ac7d3071042e0c17e80dd517a0bcdbf798e9
AeroSpace.app server version: 0.0.0-SNAPSHOT 0589ac7d3071042e0c17e80dd517a0bcdbf798e9
```

Yan-Yu-Lin:
## Summary
Fixes #562 

When a window enters macOS native fullscreen and then exits, it now returns to its **exact original position and size** in the tiling tree, instead of being placed next to the most recently used window.

## Changes
- Add `isInMacosNativeFullscreen` flag to `Window` - window stays in tree instead of being moved to a separate container
- Modify `layoutTiles()` to calculate weights for ALL children but only position visible ones
- Modify `layoutAccordion()` to skip fullscreen windows during layout
- Modify `normalizeLayoutReason.swift` to set/clear flag instead of rebinding window
- Modify `MacosNativeFullscreenCommand.swift` to use flag instead of moving window
- Prevent container flattening when only child is fullscreen window
- `mostRecentWindowRecursive` skips fullscreen windows

## How it works
1. **Enter fullscreen:** Window stays in tree, `isInMacosNativeFullscreen = true`
2. **Layout:** Calculates weights for all children (preserving proportions), but skips positioning fullscreen windows
3. **Exit fullscreen:** Flag cleared, window is already in correct position with correct weight

## Test plan
- [x] Window enters native fullscreen, exits → returns to exact same position
- [x] Window size/weight preserved after exiting fullscreen
- [x] Multiple windows side-by-side, one goes fullscreen, exits → returns correctly
- [x] Accordion mode works correctly with fullscreen windows
- [x] Focus traversal skips fullscreen windows
- [x] Container normalization doesn't flatten containers with only fullscreen children

### DIFF
--- a/Sources/AppBundle/layout/layoutRecursive.swift
+++ b/Sources/AppBundle/layout/layoutRecursive.swift
@@ -101,7 +101,18 @@ extension TilingContainer {
         var point = point
         var virtualPoint = virtual.topLeftCorner
 
-        // Filter out windows that are in macOS native fullscreen - they stay in tree but get 0px
+        guard !children.isEmpty else { return }
+
+        // Calculate delta using ALL children (including fullscreen) to preserve weight proportions
+        guard let delta = ((orientation == .h ? width : height) - CGFloat(children.sumOfDouble { $0.getWeight(orientation) }))
+            .div(children.count) else { return }
+
+        // Adjust weights for ALL children (including fullscreen) to keep them in sync
+        for child in children {
+            child.setWeight(orientation, child.getWeight(orientation) + delta)
+        }
+
+        // Filter visible children for positioning only
         let visibleChildren = children.filter { child in
             if let window = child as? Window {
                 return !window.isInMacosNativeFullscreen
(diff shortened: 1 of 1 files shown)


nikitabobko:
This approach is wrong. Now, whoever interacts or manipulates the tree (e.g., `move`, `focus` commands) has to consider `isInMacosNativeFullscreen` special case.

And please don't ignore the PR template.

PR #1914, opened 2026-01-23
Fix focus jumping when same-app floating windows closes

### LINKED ISSUE #776: 0.16.0 causes jumping to another application when Emacs child frame closes
I encounter this problem when I upgrade to the latest aerospace 0.16.0, I confirmed that previous 0.15.2 version doesn't cause this issue.

This issue is very like #325 , the problem specifically involves Emacs child frames, which are temporary windows created for UI elements like completion menus

Behavior:
- Emacs auto completion library creates child frame,  and when it finished, Emacs deletes its child frame.
- At this point, Aerospace loses track of which window should have focus
- This causes incorrectly jumping focus to another application, making Emacs auto completion completely unusable.
- Disable Aerospace, this issue disappears. 

These child frame is often undecorated
```elisp
(make-frame
 `((name . ,frame-name)   
   (parent-frame . ,parent)
   (undecorated . t)    
   ...))
```

My attempts to force focus or modify frame parameters didn't work, suggesting this is more of an Aerospace level issue

### LINKED ISSUE #1220: When you press space in completion popup in Calendar app, AeroSpace refocuses to the next window on the workspace
Stumbling on a bit of an annoying quirk quite often. 

This happens f you have both the Apple Calendar app, and one or more other applications open in one workspace. 
If you are in the Calendar app, and create a new event - this creates a small pop-up to type in the event details - Aerospace will switch to the next window after a few keystrokes. 

This gets annoying quickly, especially if that next window is Slack or something that allows typing or messaging. 
Workaround is temporarily closing the other window(s) and just having the Calendar app open in a workspace. 

**Steps to reproduce**
1. Have Calendar and another application open
2. Create a new Calendar event, start typing in the title
3. Observe what happens after a few keystrokes

**Expected result**
The pop-up stays in focus and you can finish the event

**Actual result**
The pop-up is unfocused, and the next available window is selected.

_Originally posted by @hiddekramermusic in https://github.com/nikitabobko/AeroSpace/discussions/1219_

jperras:
## Summary

Skip focus recalculation when a floating window (dialog) is destroyed if it belongs to the same app as the currently focused window.

This fixes Emacs child frames (posframes, completion popups, etc.) causing focus to jump to another application when they close.

## References

Fixes #776
Fixes #1220

## PR checklist

### DIFF
--- a/Sources/AppBundle/tree/MacWindow.swift
+++ b/Sources/AppBundle/tree/MacWindow.swift
@@ -93,6 +93,17 @@ final class MacWindow: Window {
         {
             switch parent.cases {
                 case .tilingContainer, .workspace, .macosHiddenAppsWindowsContainer, .macosFullscreenWindowsContainer:
+                    // Skip focus recalculation if the destroyed window is a dialog/floating window
+                    // belonging to the same app as the currently focused window.
+                    // This prevents focus from jumping away when child frames (e.g., Emacs posframes,
+                    // completion popups) are destroyed.
+                    // https://github.com/nikitabobko/AeroSpace/issues/776
+                    if focus.windowOrNil?.app.pid == app.pid && parent is Workspace {
+                        // The destroyed window was a floating window (dialog) from the same app.
+                        // Don't recalculate focus - let the app handle it internally.
+                        break
+                    }
+
                     let deadWindowFocus = deadWindowWorkspace.toLiveFocus()
                     _ = setFocus(to: deadWindowFocus)
                     // Guard against "Apple Reminders popup" bug: https://github.com/nikitabobko/AeroSpace/issues/201


zarthross:
This is effecting me to.  I would love to see this PR merged!

jamescrake-merani:
I just tested this PR on my machine, and it fixes the issue for me with Emacs. As others have said, it would be great to see this merged! :)

zarthross:
@nikitabobko Can this PR get some attention please 🥺

nikitabobko:
I don't like the proposed logic as it sounds wrong to me. I don't like the existing logic in general, but I don't know how to do it better.

This PR just increases the technical debt and the complexity of the code that I didn't like already.

I'd like to notice that this "preserve the focused workspace" logic doesn't activate if the window is detected as a popup. So the alternative fix could be to improve AeroSpace popup heuristics to detect these Emacs windows as popups (according to your descriptions, they do sound like popups, not real windows)

Pointers:
1. You can use `aerospace debug-winddows` to capture Emacs windows properties
2. You can try to improve the popup detection logic in `Sources/AppBundle/model/AxUiElementWindowType.swift` + cover the logic with tests by placing Emacs windows dumps in `axDumps` directory

Closing the PR for now

Sources/AppBundle/tree/MacWindow.swift
```diff
@@ -93,6 +93,17 @@ final class MacWindow: Window {
...
                 case .tilingContainer, .workspace, .macosHiddenAppsWindowsContainer, .macosFullscreenWindowsContainer:
+                    // Skip focus recalculation if the destroyed window is a dialog/floating window
+                    // belonging to the same app as the currently focused window.
+                    // This prevents focus from jumping away when child frames (e.g., Emacs posframes,
+                    // completion popups) are destroyed.
+                    // https://github.com/nikitabobko/AeroSpace/issues/776
+                    if focus.windowOrNil?.app.pid == app.pid && parent is Workspace {
+                        // The destroyed window was a floating window (dialog) from the same app.
+                        // Don't recalculate focus - let the app handle it internally.
+                        break
```
nikitabobko, line 104: Was the guard logic was added to avoid `setFocus(to: deadWindowFocus)` or `deadWindowFocus.windowOrNil?.nativeFocus()` or both?

jperras: Emacs (at least the flavour I have) doesn't end up calling `nativeFocus()` in the branching logic after this because, well, all the windows/frames that emacs creates belong to emacs.

The `setFocus(to: deadWindowFocus)`, from what I remember, is what messes things up, though it might be a red herring. When I was bisecting, it was a bit tough to pinpoint things exactly because of gc changes in 405dec8b7170cf0d9dc6ca1e7495fe7abbb20ff8, where you moved to a polling-based implementation.

jperras:
@nikitabobko yeah, that's fair - this isn't the cleanest fix. Emacs child frames should probably just be treated as popups, even though they're a bit weird (because emacs). 

FWIW I've been running this fix for a while now without any complications, but I'm not the most adventurous when it comes to macOS app usage, so it's definitely possible this diff would cause additional problems for other apps that might do some more esoteric/weird stuff with child windows.

I'll see if I can improve the popup detection heuristics instead.

jamescrake-merani:
If I'm not mistaken, I think some applications create new windows for popups like completion similar to what we see with Emacs. I remember someone having an issue with MATLAB (which I think uses Swing but not 100% certain of this) on a tiling manager because it would try to tile the completion popups... I don't know if that error is the same with Aerospace (I try to avoid MATLAB like the plague :P ), but it could be.

jperras:
For anyone following along on this PR that uses emacs, try out #2036 - I believe it's a more sane fix.

PR #1915, opened 2026-01-23
Fix for macos native tabs changing window ID

### LINKED ISSUE #68: Native macOS Big Sur tabs are considered windows
image

Hello everyone! When I open two or more IDEA projects with tabs. IDEAs take up space as white space.

Follow the steps below to reproduce
1.
image

2.
image

3.open projects

[How to open projects as tabs in IntelliJ - Stack Overflow](https://stackoverflow.com/questions/65397429/how-to-open-projects-as-tabs-in-intellij)

ryanmeisters:
I think this fixes #68

After a bit of fiddling, it looks like the issue is that macos changes the window id when tab is created/changed. So I think the fix is to just remove the old window id when it changes.

I've done some testing with the dev build and it seems to work well.

Definitely open to feedback as I have little context on this codebase :)

### DIFF
--- a/Sources/AppBundle/tree/MacApp.swift
+++ b/Sources/AppBundle/tree/MacApp.swift
@@ -290,7 +290,17 @@ final class MacApp: AbstractApp {
                 }
             }
 
-            for (id, window) in axApp.threadGuarded.get(Ax.windowsAttr) ?? [] {
+            let allWindows = axApp.threadGuarded.get(Ax.windowsAttr) ?? []
+            let currentWindowIds = Set(allWindows.map { $0.0 })
+
+            // Remove windows that are no longer in the AX windows list
+            // This handles macOS native tabs where the window ID changes when tabs are created/switched
+            for (id, window) in alive where !currentWindowIds.contains(id) {
+                dead[id] = window
+            }
+            alive = alive.filter { currentWindowIds.contains($0.key) }
+
+            for (id, window) in allWindows {
                 try job.checkCancellation()
                 try alive.getOrRegisterAxWindow(windowId: id, window, nsApp, job)
             }


Sources/AppBundle/tree/MacApp.swift
```diff
@@ -290,7 +290,17 @@ final class MacApp: AbstractApp {
                 }
             }
 
-            for (id, window) in axApp.threadGuarded.get(Ax.windowsAttr) ?? [] {
+            let allWindows = axApp.threadGuarded.get(Ax.windowsAttr) ?? []
```
ryanmeisters, line 293: not sure but wondering if this should possible go inside the ` if frontmostAppBundleId != lockScreenAppBundleId` above

tdib:
Curious what app(s) you tested this with? On my end it seems like Android Studio still exhibits some weird behaviour - notably when I switch the tabs it completely forgets that anything was tracked by AeroSpace, so I suppose something more is needed for this to work properly

https://github.com/user-attachments/assets/df09f671-b65e-4a0d-b467-d2429273b134

ryanmeisters:
Oh, dang. I tested with Finder, Ghostty, Fork on macos 15.7.1. Thanks for testing! I'll try to do some more

ryanmeisters:
also found a bug where layout isn't preserved when switching tabs if there are sibling nodes.

tdib:
fyi this was on Tahoe 26.3 Beta in case there was some changes that may affect behaviour across versions

ryanmeisters:
Just as an update.. This solution is definitely not ready for primetime. I will try to spend some more time on it.
 @nikitabobko, if you have a sec, what would you want to see from this PR in order to have confidence enough to merge? Off the top of your head, any tests that could/should be added? Any guidance appreciated 🙏

nikitabobko:
The suggested patch is naive and doesn't fix the issue. In your patch, different tabs are still considered different windows, and they behave as different windows (they can have different workspace assignment, show up as different windows in `list-windows`, have different size, etc.)

My guidance is: don't waste your time as the real fix is much more complicated. I have some ideas on how the real fix should look like, but not the complete picture.

Issue #1816, opened 2025-11-24
Order workspaces in the menubar/`list-workspaces`/`workspace next/prev` according to `persistent-workspaces` config option


nikitabobko:
Right now, the workspaces are ordered alphabetically. The suggestion is to order them as in `persistent-workspaces` list

- First, go persistent workspaces in their order.
- Second, go phantom workspaces alphabetically as they do it now

For the traymenu menu, I even suggest splitting them into two visual groups

Previous issue: https://github.com/nikitabobko/AeroSpace/issues/999

dannybarnes:
+1 on this. I currently have `persistent-workspaces = ["S", "B", "C"]` but in the traymenu it's displayed as B C S which is annoying.

fipski:
is it alphabetical? I tried to sort them by prepending numbers, but it ended up random...
I would like to sort them by screens and give each screen a accessible row of keys. 1 is numerical, 2 qwerty, 3 asdf.

`persistent-workspaces = ["01", "02", "03", "0A", "4", "5", "6", "7", "8", "9", "Q", "W",
                         "E", "R", "T", "Z"]
`

<img width="441" height="271" alt="Image" src="https://github.com/user-attachments/assets/58cfabfe-d6a6-40b0-9543-ce3cf1e8c404" />

ppeev:
Why should the order of the workspaces be tied to whether they are persistent? I see no point in having persistent workspaces because I have 20 defined and use no more than 15 at any given time, usually 7-9. On the rare occasion I use a mouse I sometimes click the AeroSpace menubar item to navigate between workspaces, and on the not-so-rare occasion I use my custom Alfred workflow to display and switch my workspaces and apps, I don't want to see all 20, I only want to see the ones that are actually in use now.

partytoni:
> Why should the order of the workspaces be tied to whether they are persistent? I see no point in having persistent workspaces because I have 20 defined and use no more than 15 at any given time, usually 7-9. 

Let me show you my issue so you can understand better. 

I have defined the following persistent workspaces:

`persistent-workspaces = ["1", "2", "3", "Q", "W", "E", "A", "S", "D"]`

and I use shortcuts to go back and forth with `workspace next/prev` as the owner of this issue.

If I do this, after the workspace `3` there is `A`, then `E`, and so on.

For this reason it would be nice to implement some sort of flag that uses that list instead of alphabetical order.

@fipski Thank you for the PR you made, i hope it will be merged soon.

[...]

PR #1926, opened 2026-01-31
Sort Workspaces in UI by persistent-workspaces

fipski:
see #1816


fipski:
This is rather crude, maybe more a poc. The sorting code should probably not live in the ui / rendering compontent, but I didn't want to break stuff.
I'm new to swift.

PR #1940, opened 2026-02-07
Skip tiling background tabs in native macOS tab groups

### LINKED ISSUE #486: Aerospace not detecting Zathura PDF viewer (NULL-APP-BUNDLE-ID)
I have the following in my config:
```toml
[[on-window-detected]]
if.app-id = 'NULL-APP-BUNDLE-ID'
if.app-name-regex-substring = 'zathura'
if.window-title-regex-substring = 'zathura'
run = 'move-node-to-workspace P'
```
Zathura does not move to space P. It does appear in the dock, application switcher, and is hidden when I run `aerospace list-apps macos-native-hidden`

noeljackson:
Ghostty uses native macOS tabs (`NSWindow.addTabbedWindow`) where each tab is a separate `NSWindow`. AeroSpace sees background tabs as regular windows and tiles them, messing up the layout.

**Fix (commit 1):** Cross-reference `CGWindowListCopyWindowInfo` (`.optionOnScreenOnly`) with the AX window list. Windows that aren't on-screen (and aren't minimized/fullscreen/hidden) are background tabs — put them in the popup container instead of tiling them. When the user switches tabs, the new tab gets promoted back to tiled and the old one goes to popup.

**Fix (commit 2):** Save the demoted tab's `BindingData` (parent, weight, index) before moving it to the popup container, then restore that exact position when promoting the replacement tab. Previously the promoted tab landed at MRU position, causing windows to jump around on tab switch.

Tested with Ghostty — new tabs no longer create extra tiled spaces, switching tabs keeps the layout stable with no window jumping.

Fixes #486

### DIFF
--- a/Sources/AppBundle/normalizeLayoutReason.swift
+++ b/Sources/AppBundle/normalizeLayoutReason.swift
@@ -1,11 +1,18 @@
+// Stores the tiling position of a demoted background tab so the
+// newly-active tab from the same app can take the exact same slot.
+@MainActor
+private var demotedTabSlots: [Int32: BindingData] = [:]
+
 @MainActor
 func normalizeLayoutReason() async throws {
+    demotedTabSlots = [:]
     for workspace in Workspace.all {
         let windows: [Window] = workspace.allLeafWindowsRecursive
         try await _normalizeLayoutReason(workspace: workspace, windows: windows)
     }
     try await _normalizeLayoutReason(workspace: focus.workspace, windows: macosMinimizedWindowsContainer.children.filterIsInstance(of: Window.self))
     try await validateStillPopups()
+    demotedTabSlots = [:]
 }
 
 @MainActor
@@ -19,7 +26,12 @@ private func validateStillPopups() async throws {
         if !isWindowOnScreen(popup.windowId) { continue }
         let windowLevel = getWindowLevel(for: popup.windowId)
         if try await popup.isWindowHeuristic(windowLevel) {
-            try await popup.relayoutWindow(on: focus.workspace)
+            if let slot = demotedTabSlots.removeValue(forKey: popup.macApp.pid) {
+                popup.unbindFromParent()
+                popup.bind(to: slot.parent, adaptiveWeight: slot.adaptiveWeight, index: slot.index)
+            } else {
+                try await popup.relayoutWindow(on: focus.workspace)
+            }
             try await tryOnWindowDetected(popup)
         }
     }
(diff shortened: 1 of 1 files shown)


nikitabobko:
The patch doesn't fix the issue

https://github.com/user-attachments/assets/7e356df9-8291-4138-bcf8-7654e7538177

On top of that, all windows loose their workspace assignment when users switch native macOS Spaces (macOS native fullscreen included)

The real fix is more complicated

noeljackson:
Thanks for the feedback, the original patch was definitely incomplete. I've reworked the approach to handle both issues (broken tab detection + windows losing workspace assignment on Space switches).

**Background tab detection** now uses `CGWindowListCopyWindowInfo(.optionOnScreenOnly)` to snapshot on-screen windows at the start of each refresh. A tiled window that isn't on-screen (and isn't minimized, fullscreen, or from a hidden app) is treated as a background tab. Dropped `isWindowAxAlive` entirely.

**Tab switch position preservation:** When a tab gets demoted, its `BindingData` (parent, weight, index) is saved in a per-PID map. The newly-active tab from the same app reclaims that exact slot, so no relayout or jumping.

**Space switch handling:** During GC in `refresh()`, windows that exist in the system (CGWindowList without `optionOnScreenOnly`) but aren't AX-alive are inactive-Space windows. These get suspended to the popup container with their binding data saved, then restored to their original position when the user switches back.

The key thing is that AX reports background tabs as alive but does not report inactive-Space windows, so we can distinguish the two cases cleanly.

Happy to rework if you have a different direction in mind or edge cases I should handle.

nikitabobko:
#68

This issue is complicated, it requires a lot of focus and has to be executed carefully. I want to take my time, sit down, experiment, and carefully fix it in the right way. Submitting yet another PR doesn't help but just puts even more burden on me, because I would need to find right words to explain to the submitter why their approach is wrong (I am sure it will be wrong). I do not believe that 3rd party contributors can fix the issue taking all the edge cases into account.

I will have more time in April to work on the project, that's when I plan to take a look at this issue. Until then, you are (and other people) are free to use AeroSpace forks.

noeljackson:
> #68
> 
> 
> 
> This issue is complicated, it requires a lot of focus and has to be executed carefully. I want to take my time, sit down, experiment, and carefully fix it in the right way. Submitting yet another PR doesn't help but just puts even more burden on me, because I would need to find right words to explain to the submitter why their approach is wrong (I am sure it will be wrong). I do not believe that 3rd party contributors can fix the issue taking all the edge cases into account.
> 
> 
> 
> I will have more time in April to work on the project, that's when I plan to take a look at this issue. Until then, you are (and other people) are free to use AeroSpace forks.

That's fine @nikitabobko if you don't want the contribution. I feel having something open source means that you will receive PRs, and it's my responsibility to share the improvements I make.

I am sorry it's not well received.  It works great for me, and I am sure having it public will be helpful for anyone using Ghostty and AeroSpace together. As right now it's terribly broken.

Again, sorry for the burden.

PR #1942, opened 2026-02-09
Docs: document --format restrictions when --json is used

fbehrens:
## Summary
- Document that `%{right-padding}` is not allowed in `--format` when `--json` is used
- Document that only interpolation variables and spaces are allowed in `--format` with `--json`
- Updated both `list-windows` and `list-apps` docs

## Context
Running e.g. `aerospace list-windows --all --json --format '%{workspace}%{right-padding}|%{window-id}'` produces an error, but the documentation didn't mention this restriction.

### DIFF
--- a/docs/aerospace-list-apps.adoc
+++ b/docs/aerospace-list-apps.adoc
@@ -38,6 +38,8 @@ Incompatible with: `--format`, `--json`
 
 --json:: Output in JSON format.
 Can be used in combination with `--format` to specify which data to include into the json.
+When `--json` is used, only interpolation variables and spaces are allowed in `--format`.
+`%{right-padding}` is not allowed with `--json`.
 Incompatible with `--count`
 
 // =========================================================== Output Format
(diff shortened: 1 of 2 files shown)


nikitabobko:
Do these paragraphs add any value? They are technically correct, but they bloat the documentation. It's kinda obvious that `%{right-padding}` and `--json` don't make sense together.

You are welcome to debate in the closed PR. My opinion on this question is not definite.

Thank you, for openly saying that you used AI.

fbehrens:
Ok, thanks you.  I see the point. 

```bash
a ~/c/Github | 1 ❱ aerospace list-windows --all --json --format  '%{workspace}%{right-padding}|%{window-id}%{righ
t-padding}|%{app-name}%{right-padding}|%{app-bundle-id}%{right-padding}|%{window-layout}%{right-padding}|%{window-title}'
%{right-padding} interpolation variable is not allowed when --json is used

Now i see that the error message is completely helpful.

PR #1958, opened 2026-02-15
Add get-tree and append-layout commands

glassbe:
## Summary

Adds two new commands inspired by i3's `get_tree` and `append_layout`:

- **`get-tree`** — Serializes the workspace tree to JSON (containers with layout/orientation/weights, windows with IDs and bundle IDs)
- **`append-layout`** — Reads a JSON layout spec from stdin and reconstructs the workspace tree, matching windows by `window-id` then `app-bundle-id`

Together they enable programmatic layout save/restore — a round-trip workflow:
```bash
# Save
aerospace get-tree --workspace 1 > layout.json
# Restore
cat layout.json | aerospace append-layout --workspace 1
```

## Motivation

AeroSpace currently has no way to programmatically save and restore window arrangements. Users who want preset layouts (e.g., coding layout with specific accordion/tiles nesting) must resort to fragile shell scripts using `flatten` + `move` + `join-with`, which break due to unpredictable window ordering.

These commands fill the gap, similar to i3's `i3-save-tree` / `append_layout` workflow.

Related discussion: #1957

## Implementation Details

- `get-tree` recursively walks the workspace tree and outputs JSON with container layout/orientation/weight and window metadata
- `append-layout` flattens the workspace, matches windows first by `window-id` (exact match) then by `app-bundle-id` (fallback), and rebuilds the container hierarchy from the JSON spec
- Custom window sizes (weights) are preserved through the round-trip — a 70/30 split stays 70/30 after restore
- Missing windows (apps not running) don't collapse the tree structure — only completely empty leaf containers are pruned
- The CLI forwards stdin to the server for `append-layout` (same pattern as `workspace --stdin`)
- `append-layout` accepts both raw container specs and full `get-tree` output (with `"type": "workspace"` wrapper)

## Files Changed

**New files (8):**
- `docs/aerospace-get-tree.adoc` — command documentation
- `docs/aerospace-append-layout.adoc` — command documentation
- `Sources/Common/cmdArgs/impl/GetTreeCmdArgs.swift` — command args
- `Sources/Common/cmdArgs/impl/AppendLayoutCmdArgs.swift` — command args
- `Sources/AppBundle/command/impl/GetTreeCommand.swift` — tree serialization with weights
- `Sources/AppBundle/command/impl/AppendLayoutCommand.swift` — tree construction with window-id matching
- `Sources/AppBundleTests/command/GetTreeCommandTest.swift` — 3 tests
- `Sources/AppBundleTests/command/AppendLayoutCommandTest.swift` — 7 tests

**Modified files (5):**
- `Sources/Common/cmdArgs/cmdArgsManifest.swift` — register new commands
- `Sources/Common/cmdHelpGenerated.swift` — regenerated with new help strings
- `Sources/Cli/subcommandDescriptionsGenerated.swift` — regenerated with new subcommand entries
- `Sources/AppBundle/command/cmdManifest.swift` — command dispatch
- `Sources/AppBundle/model/Json.swift` — add `.double` case for weight serialization
- `Sources/Cli/_main.swift` — forward stdin for append-layout

## Test Plan

- [x] All 113 existing tests pass, 10 new tests added (116 total on clean branch)
- [x] 3 `get-tree` tests: simple, nested, floating windows (including weight output)
- [x] 7 `append-layout` tests: simple, nested, unmatched windows, window-id matching, weight preservation, missing window structure preservation, full round-trip
- [x] Manual testing: save workspace layout → flatten → restore from saved JSON
- [x] `generate.sh` regeneration produces consistent output (`.adoc` source files included)

### DIFF
--- a/Sources/AppBundle/command/cmdManifest.swift
+++ b/Sources/AppBundle/command/cmdManifest.swift
@@ -4,6 +4,8 @@ extension CmdArgs {
     func toCommand() -> any Command {
         let command: any Command
         switch Self.info.kind {
+            case .appendLayout:
+                command = AppendLayoutCommand(args: self as! AppendLayoutCmdArgs)
             case .balanceSizes:
                 command = BalanceSizesCommand(args: self as! BalanceSizesCmdArgs)
             case .close:
@@ -28,6 +30,8 @@ extension CmdArgs {
                 command = FocusMonitorCommand(args: self as! FocusMonitorCmdArgs)
             case .fullscreen:
                 command = FullscreenCommand(args: self as! FullscreenCmdArgs)
+            case .getTree:
+                command = GetTreeCommand(args: self as! GetTreeCmdArgs)
             case .joinWith:
                 command = JoinWithCommand(args: self as! JoinWithCmdArgs)
             case .layout:
--- a/Sources/AppBundle/command/impl/AppendLayoutCommand.swift
+++ b/Sources/AppBundle/command/impl/AppendLayoutCommand.swift
@@ -1,0 +1,140 @@
+import AppKit
+import Common
+
+struct AppendLayoutCommand: Command {
+    let args: AppendLayoutCmdArgs
+    /*conforms*/ let shouldResetClosedWindowsCache: Bool = true
+
+    @MainActor
+    func run(_ env: CmdEnv, _ io: CmdIo) -> Bool {
+        let jsonString = io.readStdin()
+        guard !jsonString.isEmpty else {
+            return io.err("append-layout: No JSON layout provided on stdin")
+        }
+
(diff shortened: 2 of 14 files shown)


unclecheese:
This is exactly what I've been looking for. The imperative workarounds I'm using have not worked out well.

Would it be possible to match by window-title regex as well, like we can do in the window detected event?

patrik-bartak:
Does this output the windows ordered according to the actual arrangement? This is a great feature for me because I would like my sketchybar to show window icons in the actual arrangement order, for each workspace. Right now it is not possible because `list-windows --workspace <workspace>` seems to just print them in an arbitrary order.

PR #1963, opened 2026-02-16
Optional parameter to specify preferred editor

technocidal:
Give users the option to explicitly set an editor instead of relying on the automatic behavior. There are scenarios where users have different editors for different purposes e.g. Visual Studio Code for actual projects and Textastic for small configuration files.

## PR checklist

- [X] Explain your changes in the relevant commit messages rather than in the PR description. The PR description must not contain more information than the commit messages (except for images and other media).
- [X] Each commit must explain what/why/how and motivation in its description. https://cbea.ms/git-commit/
- [ ] Don't forget to link the appropriate issues/discussions in commit messages (if applicable).
- [X] Each commit must be an atomic change (a PR may contain several commits). Don't introduce new functional changes together with refactorings in the same commit.
- [X] `./run-tests.sh` exits with non-zero exit code. [test.log](https://github.com/user-attachments/files/25351205/test.log)
- [X] Avoid merge commits, always rebase and force push.

### DIFF
--- a/docs/config-examples/default-config.toml
+++ b/docs/config-examples/default-config.toml
@@ -17,10 +17,8 @@ start-at-login = false
 auto-reload-config = false
 
 # Override which editor should be used when opening the config file via the menu bar
-# By default, the system's default text editor for the config file is used (excluding Xcode due to its size)
-# with TextEdit as a fallback. Set this to the full path of your preferred editor application.
-# Example: config-editor-app-path = '/Applications/Visual Studio Code.app'
-# Fallback value (if you omit the key): config-editor-app-path = <nil>
+# By default, macOS will decide which editor to open with TextEdit as the fallback
+# Set this to the full path of your preferred editor application.
 # config-editor-app-path = '/Applications/Visual Studio Code.app'
 
 # Normalizations. See: https://nikitabobko.github.io/AeroSpace/guide#normalization


Sources/AppBundle/ui/MenuBar.swift
```diff
@@ -110,8 +110,16 @@ func shortcutGroup(label: some View, content: some View) -> some View {
     }
 }
 
-func getTextEditorToOpenConfig() -> URL {
-    NSWorkspace.shared.urlForApplication(toOpen: findCustomConfigUrl().urlOrNil ?? defaultConfigUrl)?
+@MainActor func getTextEditorToOpenConfig() -> URL {
```
technocidal, line 113: (question) Moving this to `@MainActor` is probably not a big deal (e.g. this function isn't constantly called, other functions in this same file that are used to populate the menubar are already annotated with `@MainActor` etc.), but it's the most significant change from a runtime perspective so I wanted to point it out.

nikitabobko:
I am against adding a config option for such minor use case. You are overconfiguring

technocidal:
Makes sense! I'll take your suggestion to heart and if I still want this down the road, I can always just setup an automated workflow somewhere to have a patched version.

AeroSpace has made my Mac so much more enjoyable. Thank you for building it!

Issue #1519, opened 2025-06-29
Floating windows position and size is not tracked properly


nikitabobko:
Tiling windows management is uni-directional in AeroSpace. AeroSpace is the owner of the state, and it enforces positions of the windows in the idempotent style. AeroSpace effectively ignores what macOS thinks about windows' position and size. Side-effects flow only in one direction - from AeroSpace to macOS. All good, all reliable

Floating windows are not like that. Their management is bi-directional. AeroSpace has to listen what macOS has to say. For example, it will cause problems for floating windows in the following scenario `aerospace eval 'move right --floating-pixels 50; move right --floating-pixels 50'` (#278)

I was inspired to write this issue because I already spent some brain power on #1492 but I don't know the right solution yet. Unfortunately, the priority of this issue is really minor since I don't care about floating windows much, so this issue is probably gonna just rot.

## Possible solutions

Turns out that there are already a number of places in AeroSpace (hmm, only 2?), where we have to deal with the problem of bi-directionality

- Focus handling is bi-directional. Right now, AeroSpace implements `lastKnownNativeFocusedWindowId` approach where the focus from macOS is applied only when the `lastKnownNativeFocusedWindowId` changes. Downside if we apply it for floating windows: in every refresh session, AeroSpace will have to query all visible floating window's position and size which doesn't sound good for performance, or use AX notifications, in which I have little trust, hmm hmm hmm
- Apply changes from macOS only when the left mouse button is down. If we apply it for all floating windows, the downside is that non-AeroSpace tilers won't work with AeroSpace anymore

## Problems it causes

- `layoutRecursive.swift:layoutFloatingWindow`. Calculate everything against current "physical" position `let currentMonitor = try await getCenter()?.monitorApproximation`. I find it hilarious. It's just unreliable, and non idempotent, but apparently works
  - The bug fix for #1492 makes `MacWindow.hideInCorner` behave in the same way.
- The mentioned `aerospace eval` problem above

nikitabobko:
After thinking about it more, I think the first possible solution is the way to go

> in every refresh session, AeroSpace will have to query all visible floating window's position and size which doesn't sound good for performance

For one, we already do it in `layoutRecursive` :( For another, we can optimize it by forward scanning the commands that we are about to run to see if we need to query all visible floating window's position and size

acsandmann:
is it not an option to listen to kAXWindowMovedNotification and just track that?

[...]

PR #1964, opened 2026-02-19
Floating window position outside screen bounds when moving between workspaces

mobile-ar:
Issue:

When moving a floating window to another workspace (potentially on a different monitor with different resolution), the window could end up placed almost completely outside the bounds of the target workspace. This happens because the proportional repositioning logic only remaps the window's top-left corner without considering the window's size, so a window near the edge of a larger monitor gets its top-left mapped to the edge of a smaller monitor, and the window extends almost completely off the screen.

Root Cause:

When layoutFloatingWindow and unhideFromCorner calculates the coordinates for the top-left corner it does not accommodate for the target workspace's bounds.

Fix:

layoutFloatingWindow:
- Replaced two separate AX calls of `getCenter()` and `getAxTopLeftCorner()` with a single call to `getAxRect()` to get both position and size of the window.
- After computing the proportional position on the target workspace, restrict the position using `coerceIn` so the window stays within `workspace.workspaceMonitor.visibleRect`, accounting for the window's width and height

unhideFromCorner:
- After computing the restored proportional position on the workspace, clamp it within `workspaceRect` using `lastFloatingSize` to account for window dimensions.

The restrict logic uses `max(rect.minX, rect.maxX - windowWidth)` as the upper bound, which handles the edge case where the window is larger than the target workspace — in that case, the range collapses to `minX...minX` and the window is pinned to the top-left corner of the workspace (which is the most reasonable behavior).

Related discussion: https://github.com/nikitabobko/AeroSpace/discussions/1875 (with my fix the screen no longer disappear, but it gets re positioned in the screen)
Possible related issue: https://github.com/nikitabobko/AeroSpace/issues/1519

## PR checklist

### DIFF
--- a/Sources/AppBundle/layout/layoutRecursive.swift
+++ b/Sources/AppBundle/layout/layoutRecursive.swift
@@ -69,16 +69,23 @@ extension Window {
     @MainActor
     fileprivate func layoutFloatingWindow(_ context: LayoutContext) async throws {
         let workspace = context.workspace
-        let currentMonitor = try await getCenter()?.monitorApproximation // Probably not idempotent
-        if let currentMonitor, let windowTopLeftCorner = try await getAxTopLeftCorner(), workspace != currentMonitor.activeWorkspace {
+        let windowRect = try await getAxRect() // Probably not idempotent
+        let currentMonitor = windowRect?.center.monitorApproximation
+        if let currentMonitor, let windowRect, workspace != currentMonitor.activeWorkspace {
+            let windowTopLeftCorner = windowRect.topLeftCorner
             let xProportion = (windowTopLeftCorner.x - currentMonitor.visibleRect.topLeftX) / currentMonitor.visibleRect.width
             let yProportion = (windowTopLeftCorner.y - currentMonitor.visibleRect.topLeftY) / currentMonitor.visibleRect.height
 
-            let moveTo = workspace.workspaceMonitor
-            setAxFrame(CGPoint(
-                x: moveTo.visibleRect.topLeftX + xProportion * moveTo.visibleRect.width,
-                y: moveTo.visibleRect.topLeftY + yProportion * moveTo.visibleRect.height,
-            ), nil)
+            let workspaceRect = workspace.workspaceMonitor.visibleRect
+            var newX = workspaceRect.topLeftX + xProportion * workspaceRect.width
+            var newY = workspaceRect.topLeftY + yProportion * workspaceRect.height
+
+            let windowWidth = windowRect.width
+            let windowHeight = windowRect.height
+            newX = newX.coerceIn(workspaceRect.minX ... max(workspaceRect.minX, workspaceRect.maxX - windowWidth))
+            newY = newY.coerceIn(workspaceRect.minY ... max(workspaceRect.minY, workspaceRect.maxY - windowHeight))
+
+            setAxFrame(CGPoint(x: newX, y: newY), nil)
         }
         if isFullscreen {
             layoutFullscreen(context)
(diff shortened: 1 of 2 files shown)


mobile-ar:
Based on this discussion [1875](https://github.com/nikitabobko/AeroSpace/discussions/1875), this fix improves the usability, so the floating windows will no longer disappear but they will still change position. The issue I can see is that at some point in `MacWindow.hideInCorner` when calculating `prevUnhiddenProportionalPositionInsideWorkspaceRect` it might get called twice, returning an incorrect position (off screen now). 
This 2 prints happened sequentially one after the other.  
```
hideInCorner. for: 4545. windowRect: Rect(topLeftX: 2212.0, topLeftY: 146.0, _width: 980.0, _height: 600.0) nodeMonitor.visibleRect: Rect(topLeftX: 1800.0, topLeftY: -241.0, _width: 2560.0, _height: 1410.0)
hideInCorner. for: 4545. windowRect: Rect(topLeftX: 4359.0, topLeftY: 1117.0, _width: 980.0, _height: 600.0) nodeMonitor.visibleRect: Rect(topLeftX: 1800.0, topLeftY: -241.0, _width: 2560.0, _height: 1410.0)
```

I tested adding `@MainActor` to `isHiddenInCorner` and `prevUnhiddenProportionalPositionInsideWorkspaceRect`, this makes it a little more stable, but after some stress testing it start failing again less constant, it looks like some sort of race condition.

Sources/AppBundle/layout/layoutRecursive.swift
```diff
@@ -69,16 +69,23 @@ extension Window {
     @MainActor
     fileprivate func layoutFloatingWindow(_ context: LayoutContext) async throws {
         let workspace = context.workspace
-        let currentMonitor = try await getCenter()?.monitorApproximation // Probably not idempotent
-        if let currentMonitor, let windowTopLeftCorner = try await getAxTopLeftCorner(), workspace != currentMonitor.activeWorkspace {
+        guard let windowRect = try await getAxRect() else { return }
+        let currentMonitor = windowRect.center.monitorApproximation
+        if workspace != currentMonitor.activeWorkspace {
```
nikitabobko:
```suggestion
        let windowRect = try await getAxRect() // Probably not idempotent
        let currentMonitor = windowRect?.center.monitorApproximation
        if let currentMonitor, let windowRect, workspace != currentMonitor.activeWorkspace {
```

Otherwise, you will skip `layoutFullscreen` below

Sources/AppBundle/tree/MacWindow.swift
```diff
@@ -163,11 +163,14 @@ final class MacWindow: Window {
...
             case .floatingWindow:
                 let workspaceRect = nodeWorkspace.workspaceMonitor.rect
-                let pointInsideWorkspace = CGPoint(
-                    x: workspaceRect.width * prevUnhiddenProportionalPositionInsideWorkspaceRect.x,
-                    y: workspaceRect.height * prevUnhiddenProportionalPositionInsideWorkspaceRect.y,
-                )
-                setAxFrame(workspaceRect.topLeftCorner + pointInsideWorkspace, nil)
+                var newX = workspaceRect.topLeftX + workspaceRect.width * prevUnhiddenProportionalPositionInsideWorkspaceRect.x
+                var newY = workspaceRect.topLeftY + workspaceRect.height * prevUnhiddenProportionalPositionInsideWorkspaceRect.y
+                let windowWidth = lastFloatingSize?.width ?? 0
```
nikitabobko, line 170:
```suggestion
                // todo we probably should replace lastFloatingSize with proper floating window sizing
                // https://github.com/nikitabobko/AeroSpace/issues/1519
                let windowWidth = lastFloatingSize?.width ?? 0
```

mobile-ar: Sorry about the commit message, I added the full explanation in the commit message inside Sublime Merge, it displays the full message:
image

I will fix this commit to have a more descriptive first line and I will add your suggestions too.

PR #1985, opened 2026-03-04
fix: prevent floating windows from getting stuck in corner after work…

buacanning:
…space switch

When AeroSpace hides windows from inactive workspaces, it moves them to a screen corner and saves their previous position so they can be restored later (unhideFromCorner). A race condition with AX notifications could cause the following broken cycle:

1. hideInCorner() saves the normal position, moves window to corner
2. unhideFromCorner() restores it, clears prevUnhiddenProportionalPosition
3. A spurious AX event triggers unhideFromCorner() again — no-op since prevUnhiddenProportionalPosition is already nil, window stays in place
4. Next workspace switch calls hideInCorner() again — isHiddenInCorner is false (nil check) — so it saves the *current* position, which is already the corner position
5. unhideFromCorner() then restores the window to the corner → stuck forever

Fix: in hideInCorner(), only record the window's position when its top-left corner is actually inside the monitor's visible rect. If it's already outside (stuck in a corner from a previous broken cycle), skip saving to avoid permanently anchoring it to the corner.

Fixes: https://github.com/nikitabobko/AeroSpace/discussions/1875

### DIFF
--- a/Sources/AppBundle/tree/MacWindow.swift
+++ b/Sources/AppBundle/tree/MacWindow.swift
@@ -131,9 +131,16 @@ final class MacWindow: Window {
             guard let windowRect = try await getAxRect() else { return }
             let topLeftCorner = windowRect.topLeftCorner
             let monitorRect = windowRect.center.monitorApproximation.rect // Similar to layoutFloatingWindow. Non idempotent
-            let absolutePoint = topLeftCorner - monitorRect.topLeftCorner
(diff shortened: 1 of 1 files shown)


nikitabobko:
Thanks! The patch indeed fixes the issue, but I prefer to write it in a slightly different way

The PR is superseded by https://github.com/nikitabobko/AeroSpace/commit/90a7c67287bb2b000da98b60d387fca13ccc97ad

PR #2035, opened 2026-03-31
Add Stop hook for PR workflow enforcement

jss367:
## Summary
- Adds a Stop hook in `.claude/settings.json` that blocks Claude from ending a session when there are uncommitted changes
- Forces Claude to create a branch, commit, build, and open a PR before stopping

## Test plan
- [ ] Start a new Claude Code session, make a code change, and verify the hook blocks stopping until a PR is created


jss367:
Sorry about that. Removed the connection so won't happen again.

PR #2036, opened 2026-03-31
Treat emacs child frames as popups

### LINKED ISSUE #776: 0.16.0 causes jumping to another application when Emacs child frame closes

jperras:
Emacs has a concept of "child frames", which are effectively transient UI elements. These should not be managed by Aerospace.

These child frames are now detected by checking for an AXSubrole of AXFloatingWindow, and AXMain with value 0 for the emacs bundle ID.

This fixes the focus jumping issue when using Emacs completion/info popups (via common packages such as corfu, posframes, etc.) with AeroSpace.

Closes #776
Refs. #1914

## PR checklist


jperras:
@nikitabobko I realize my PR description is pretty verbose compared to the actual commit message for the change. I'm happy to force push/change the commit message to be more descriptive, and/or amend the PR body, too.

axDumps/emacs_child_frame_corfu.json5
```diff
@@ -0,0 +1,59 @@
+{
```
nikitabobko, line 2: Can you please add short comments to these dumps to describe how somebody else can trigger these windows. For example, take a look at the first line of `./axDumps/ghostty_check_for_updates_1_dialog.json5`

nikitabobko, line 2: Can you please also add a dump for normal Emacs window (non-popup window)

nikitabobko: Like this paragraph from your PR description:

> I added two window property dumps, one for when the focus change was triggered by a completion "popup" from a popular emacs library named [corfu](https://github.com/minad/corfu), and one for [posframe](https://github.com/tumashu/posframe), another popular package that does similar things with "popups", and is used by a lot of other packages that want the popup-like functionality.

but extract the comment from GitHub directly into the JSONs

nikitabobko (changes requested):

nikitabobko:
> I realize my PR description is pretty verbose compared to the actual commit message for the change. I'm happy to force push/change the commit message to be more descriptive, and/or amend the PR body, too.

@jperras Please do. I want to depend on GitHub as less as possible. All the information should be pushed directly into the git history (or code comments when appropriate) as much as possible

jperras:
@nikitabobko added comments to the existing dumps, added an additional dump for a normal emacs window with no popups, and amended the commit message to be more informative. Everything squashed and force-pushed to this PR branch.

heyitscassio:
I've been using this changes for some time, everything working well

skye-repos:
Hey, this PR is probably not the right place for this query, but:

Aerospace seems to have a hard time recognizing when an emacsclient parent frame has opened up, often not managing it until either another emacsclient frame opens up or some other application is moved to the workspace.

I imagine this has to do with how these frames are detected and when AeroSpace decides to refresh the layout? I'd appreciate it if someone could point me in the right direction.

Best,
Skye

nikitabobko:
I apologize for the long reply.

The `isMainAttr` check looks unnecessary. To avoid another review iteration, I've manually merged this PR with the following patch on top:

```diff
diff --git a/Sources/AppBundle/model/AxUiElementWindowType.swift b/Sources/AppBundle/model/AxUiElementWindowType.swift
index ecd16b29..f8d6172b 100644
--- a/Sources/AppBundle/model/AxUiElementWindowType.swift
+++ b/Sources/AppBundle/model/AxUiElementWindowType.swift
@@ -136,7 +136,10 @@ extension AxUiElementMock {
         // Emacs child frames (posframes, corfu completion popups, etc.)
         // These are transient UI elements that should not be managed as windows.
         // https://github.com/nikitabobko/AeroSpace/issues/776
-        if id == .emacs && get(Ax.subroleAttr) == kAXFloatingWindowSubrole {
+        if id == .emacs &&
+            get(Ax.subroleAttr) == kAXFloatingWindowSubrole &&
+            get(Ax.isMainAttr) == false
+        {
             return false
         }
 
```

But if I am wrong, and the `isMainAttr` check is important for some reason, please let me know.

Thank you!

jperras:
@nikitabobko Thank you! No apologies necessary. Last thing I want to do is pester maintainers for integrating minor QoL improvements, especially when they're pretty obscure like this one. I just appreciate the project :)

Side note: In your comment you pasted the inverse diff of what you actually applied to the code. Obviously not important, but just leaving a note here for future readers in case it's confusing.

PR #2044, opened 2026-04-06
Add scrolling workspace layout

vadika:
The scrolling layout lets a workspace behave more like a horizontal strip of windows than a fixed split grid.

It is especially useful on laptop screens, where width is limited and ordinary tiling quickly makes every window too narrow. Instead of compressing more and more windows into the same visible area, the scrolling layout keeps the workspace readable by showing only a two-pane slice at a time and letting you move through the rest horizontally.

In my setup, this works especially well for the numbered workspaces 1 to 9, where I usually step through active task windows in sequence. The first window takes the whole workspace. As soon as a second window appears, the workspace becomes a two-pane view, and every additional window is added as another half-width page to the right. I can then move through that strip with alt-left and alt-right, keeping windows large enough to actually use on a laptop display.

## Change summary
- add a root-only scrolling workspace layout
- add `scroll left` and `scroll right` commands
- add tests and docs for scrolling behavior

## Verification
- `swift test`

### DIFF
--- a/Sources/AppBundle/command/cmdManifest.swift
+++ b/Sources/AppBundle/command/cmdManifest.swift
@@ -68,6 +68,8 @@ extension CmdArgs {
                 command = ReloadConfigCommand(args: self as! ReloadConfigCmdArgs)
             case .resize:
                 command = ResizeCommand(args: self as! ResizeCmdArgs)
+            case .scroll:
+                command = ScrollCommand(args: self as! ScrollCmdArgs)
             case .split:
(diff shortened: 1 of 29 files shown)


alexcamlo:
Something like Niri?
https://www.youtube.com/watch?v=uiXiD5foFCI

vadika:
> Something like Niri? https://www.youtube.com/watch?v=uiXiD5foFCI

yep, like hyprscrolling for hyprland :)

vadika:
superseded by https://github.com/nikitabobko/AeroSpace/pull/2057

PR #2052, opened 2026-04-11
Fix black window rendering in GTK3 apps after tiling

QuirkyTurtle94:
## Summary

- Fixes black window rendering in GTK3 apps (e.g. GnuCash) after AeroSpace tiles them
- After `disableAnimations` restores `AXEnhancedUserInterface`, nudges the window size by 1px and back to force a real resize notification
- This prompts the app to redraw, fixing the stale Quartz surface

Discussion: https://github.com/nikitabobko/AeroSpace/discussions/2051

## Root cause

Frame changes via `AXUIElementSetAttributeValue` made while `AXEnhancedUserInterface` is disabled don't reliably deliver resize/move notifications to the app. GTK3's Quartz backend depends on these notifications to trigger redraws, so the window surface goes stale and appears black.

## Test plan

- [x] Open GnuCash (GTK3, Homebrew) with AeroSpace running
- [x] Open Tools → Price Database dialog
- [x] Confirm dialog renders correctly (previously rendered black)
- [x] Confirm no visible flicker from the 1px size jitter
- [x] Build compiles cleanly (`swift build`)

### DIFF
--- a/Sources/AppBundle/tree/MacApp.swift
+++ b/Sources/AppBundle/tree/MacApp.swift
@@ -137,6 +137,17 @@ final class MacApp: AbstractApp {
             try disableAnimations(app: axApp.threadGuarded, job) {
                 try setFrame(window, topLeft, size, job)
             }
+            // Nudge the window size by 1px and back after AXEnhancedUserInterface is
+            // restored.  Some toolkits (e.g. GTK3's Quartz backend) don't redraw after
+            // AX-driven frame changes made while AXEnhancedUserInterface is disabled,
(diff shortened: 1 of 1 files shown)


nikitabobko:
Reproduced the bug with GnuCash. Though, I couldn't reproduce the bug with some other popular GTK apps like GIMP and Wireshark. That's why I am not sure that GTK is the culprit.

Although your proposed 1px back and forth resizing may fix the problem, it doesn't look right. I'd prefer if the bug was reported to GnuCash and fixed there.

I close the PR for now

Issue #940, opened 2025-01-08
Extend the layout command to be able to manage the layout of the root node


nikitabobko:
### Discussed in https://github.com/nikitabobko/AeroSpace/discussions/939

<div type='discussions-op-text'>

<sup>Originally posted by **yngvedh** January  8, 2025</sup>
I'm using a setup where I have two vertical accordions side by side so that I can manage my windows into two groups, left and right. I have a laptop monitor (A) together with an extrawide external monitor (B).

On A, I want the two accordions in a horizontal accordion because it is not wide enough to have the windows side by side. On B I want them tiled because it has enough room. 

Now when I move a workspace from one monitor to the other, I also want to switch the layout of the root node.
I haven't figured out a way to do this so I'm forced to reset the layout and set it up again which is quite a hassle.

The ability to toggle the layout of the root node would solve this for me.
Something like this would do the trick: `aerospace layout --root tiles accordion`

Alternatively aerospace could automatically collapse tiles to accordions when becoming too narrow. Optionally, of course.</div>

nikitabobko:
The same for `move-node-to-workspace`

nikitabobko:
`layout --root` was added in 0.21.0-Beta

PR #2080, opened 2026-05-03
Implement `--root` flag for `layout` command to target the workspace root container

cemcatik:
Without `--root`, `aerospace layout` always acts on the parent
container of the focused window. There is no deterministic way
to set a workspace's top-level layout from a single command: the
result depends on where focus happens to sit in the tree, and
the command fails entirely on empty or floating-only workspaces.

`--root` retargets the command at the focused workspace's root
tiling container. Toggle semantics are unchanged, and the command
now also succeeds when no window is focused.

`tiling` and `floating` describe window placement rather than
container layout and are rejected at parse time when paired with
`--root`. `--root` is also mutually exclusive with `--window-id`,
which selects a window-scoped target.

Related discussion: https://github.com/nikitabobko/AeroSpace/discussions/939
Related: https://github.com/nikitabobko/AeroSpace/issues/940

PS: I used Claude Code to help me with the implementation.
This is obvious with the tests in this PR. Please let me know if you
have any concerns or would like to clarify the disclosure.

## PR checklist


nikitabobko:
Superseded by 11ba9150

PR #2083, opened 2026-05-05
Implement layout sticky for floating windows

KaiExec:
See commit message for details.

## PR checklist

### DIFF
--- a/Sources/AppBundle/command/impl/LayoutCommand.swift
+++ b/Sources/AppBundle/command/impl/LayoutCommand.swift
@@ -29,9 +29,9 @@ struct LayoutCommand: Command {
                 node = .tilingContainer(target.workspace.rootTilingContainer)
         }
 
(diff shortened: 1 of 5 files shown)


grazzolini:
I have tested this branch with a local debug build and it works. However, I have one small suggestion:

Most of the time, the windows that need sticky, are also already started as floating. So, I would suggest that macWindow.isSticky when used to match a window using [[on-window-detected]], only applies sticky to windows of that app that are floating. Otherwise if you have multiple windows of the same app, but some aren't floating, it applies to all.

Other than that, this works awesomely.

grazzolini:
Perfect, now I can do:

[[on-window-detected]]
if.app-id = 'com.app.example'
run = ['... other commands ...', 'layout sticky']

And the AeroSpace automatically applies sticky to any floating window from that app. 

@nikitabobko What does this PR need to get in? Tests?

nikitabobko:
This is not how I envision this feature being implemented. I will implement it myself at some point. The tracking issue stays the same: #2

PR #2085, opened 2026-05-06
Performance: throttle failed app registration retries

Knapsack44:
## Summary

- add a short negative retry cache for failed AX app registrations in `MacApp.getOrRegister`
- clear the retry cache when registration later succeeds
- clear the retry cache when an app is destroyed/unregistered

## Context

Potential bug discussion: https://github.com/nikitabobko/AeroSpace/discussions/2084

The observed CPU spike came from repeatedly creating short-lived `AxAppThread` instances for an app whose AX subscription registration failed, seen locally as thousands of `AxAppThread PID: 797 ID: com.apple.dock` entries in a sample.

This PR is intentionally separate from #2048 and does not include any custom-app, AeroMux, or local config changes.

## Validation

- `git diff --check origin/main...HEAD`
- `./build-debug.sh -Xswiftc -warnings-as-errors`
- `make swift-test`
- `./test.sh`

Local live validation from the discussion:

- before fix: `avg=57.8%`, `max=67.8%`, `high50=257/300`
- after fix without helper polling: `avg=0.9%`, `max=23.7%`, `high50=0/180`
- after fix with helper polling: `avg=9.5%`, `max=35.6%`, `high50=0/120`

Prepared with Codex under my supervision.

### DIFF
--- a/Sources/AppBundle/tree/MacApp.swift
+++ b/Sources/AppBundle/tree/MacApp.swift
@@ -26,6 +26,8 @@ final class MacApp: AbstractApp {
     //      and make deinitialization automatic in deinit
     @MainActor static var allAppsMap: [pid_t: MacApp] = [:]
     @MainActor private static var wipPids: [pid_t: AwaitableOneTimeBroadcastLatch] = [:]
+    @MainActor private static var failedRegistrationRetryAfter: [pid_t: Date] = [:]
+    private static let failedRegistrationRetryDelay: TimeInterval = 5
 
     private init(_ nsApp: NSRunningApplication, _ axApp: AXUIElement, _ axSubscriptions: [AxSubscription], _ thread: Thread) {
         self.nsApp = nsApp
@@ -50,6 +52,10 @@ final class MacApp: AbstractApp {
 
         while true {
             if let existing = allAppsMap[pid] { return existing }
+            if let retryAfter = failedRegistrationRetryAfter[pid] {
+                if retryAfter > Date() { return nil }
+                failedRegistrationRetryAfter[pid] = nil
+            }
             try checkCancellation()
             if let wip = wipPids[pid] {
                 try await wip.await()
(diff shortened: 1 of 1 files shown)


pszypowicz:
I hit this exact bug class on my fork (based on v0.21.0, `getOrRegister` unchanged from upstream) and can confirm your diagnosis with captures from a live episode, plus two new data points.

**Different trigger app: `com.apple.universalcontrol`.** When the cursor moves to another device via Universal Control, macOS briefly reports the Universal Control agent as `NSWorkspace.frontmostApplication`. `getNativeFocusedWindow()` then calls `MacApp.getOrRegister` on it, and since the agent has no AX server, `bulkSubscribe` always fails. A 1 ms `sample` over ~6 s caught 2,287 distinct short-lived threads for it:

```
$ grep -oE "AxAppThread PID: [0-9]+ ID: [^ ]+" sample.txt | sort | uniq -c | sort -rn | head -2
2287 AxAppThread PID: 1969 ID: com.apple.universalcontrol
   1 AxAppThread PID: ...   (long-lived, one per regular app)
```

The spindump for the same episode shows the main thread spending its dispatch time in `getOrRegister` -> `NSRunningApplication.idForDebug` -> `-[NSRunningApplication _fetchDynamicProperties]` -> `_LSCopyApplicationInformation`, blocked on `launchservicesd`, which itself burned 2.3 s CPU in the 10 s window (with `launchd` at 3.0 s from the thread churn). AeroSpace was at 11.7 s CPU per 10 s, plus 3.2 s attributed to already-exited threads.

**A single `getOrRegister` call spins forever**, since a failure removes the `allAppsMap` entry and the `while true` loop respins, so `aerospace enable off` does not stop it (confirmed: CPU stayed at ~113% after disabling, with the same signature). And every CLI request stacks another eternal loop via `getNativeFocusedWindow()`: I had 77 stuck `aerospace` client processes piled up from a status bar polling the socket, the oldest alive for 18.5 hours, which looks like the same mechanism as #2095.

Your negative cache also breaks the already-in-flight loop, not just future calls, since the next iteration hits the `retryAfter > Date()` check and returns nil. I'm cherry-picking this onto my fork and can report back with before/after numbers if useful.

nikitabobko:
Thanks for pointing out to the problem. Though, I'd prefer to fix it somehow without introducing yet another mutable state that needs to be properly cleaned up (your `failedRegistrationRetryAfter`)

I've superseded this PR with a conservative fix which should be enough to fix the problem: https://github.com/nikitabobko/AeroSpace/commit/269dc71b621bfa288a25dd06525d8ec31f16e6cb

PR #2098, opened 2026-05-21
Fix floating windows after screen wake

noomz:
## PR checklist

- [x] `./test.sh` exits with zero exit code.


nikitabobko:
I have no idea is the problem statement described in the commit message is valid. But the proposed patch definitely looks too much fragile. Sorry, I won't be wasting my time trying to understand what this code does

PR #2106, opened 2026-05-30
feat: workspace HUD overlay on workspace change

alcibiadesc:
## What

Adds a transient, Apple-style **workspace HUD**: a frosted-glass panel showing the workspace name, centered on the focused monitor, shown briefly whenever the focused workspace changes.

Workspace HUD

## Use cases

- Quick `alt+number` switching with no always-visible workspace indicator (menu bar hidden, or eyes on the content rather than the bar) — you lose track of which workspace you landed on.
- A brief, non-intrusive confirmation of the switch, mirroring the native macOS volume/brightness overlay that users already have muscle memory for.
- Multi-monitor: the HUD appears on the monitor whose focused workspace changed.

## How it works

- Reuses the existing `NSPanelHud` + SwiftUI hosting infrastructure (same pattern as `VolumePanel`).
- A new `WorkspaceHud` panel is shown from `onWorkspaceChanged()` in `focus.swift`, which already fires on every focused-workspace change and is skipped during AeroSpace startup.
- The panel fades in (~0.12s), stays for a configurable duration, then fades out (~0.25s) and is auto-dismissed by a timer.

## Config

Two keys, both with sensible defaults so the feature is opt-out, not opt-in:

```toml
# Show the overlay when the focused workspace changes (default: true)
workspace-hud = true
# How long the overlay stays on screen, in milliseconds (default: 850)
workspace-hud-duration-ms = 850
```

## Notes

- Single, additive commit; no refactoring mixed in.
- No new dependencies.
- Happy to adjust the design (size, copy, position, default duration) or gate it behind a discussion first per CONTRIBUTING — opening this as a concrete proposal to look at.

### DIFF
--- a/Sources/AppBundle/config/Config.swift
+++ b/Sources/AppBundle/config/Config.swift
@@ -44,6 +44,8 @@ struct Config: ConvenienceCopyable {
     var startAtLogin: Bool = false
     var autoReloadConfig: Bool = false
     var automaticallyUnhideMacosHiddenApps: Bool = false
+    var workspaceHud: Bool = true
+    var workspaceHudDurationMs: Int = 850
     var accordionPadding: Int = 30
     var enableNormalizationOppositeOrientationForNestedContainers: Bool = true
     var persistentWorkspaces: OrderedSet<String> = []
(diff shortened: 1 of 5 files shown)


nikitabobko:
Power users know what key they press, this kind of extra confirmation is only distracting. Definitely not going to be enabled by default.

Given that I see this request for the first time, I don't think there is a demand for this feature. You can try to post this feature request in GitHub Discussions to see if there is a demand. If there is a significant demand, I am fine to have this kind of indicator, but disabled-by-default. Closing the PR, for now

PR #2107, opened 2026-05-30
feat: native focused-window border

alcibiadesc:
## What

Adds an optional, native **focused-window border**: AeroSpace itself draws a colored, rounded border around the currently focused window. A native alternative to external tools (e.g. JankyBorders) for the common "highlight the focused window" use case.

## Use cases

- Quickly see which window has focus in a dense tiling layout, without running a separate background daemon.
- Removes a dependency for users who only need the focused-window highlight (not per-window borders on everything).

## How it works

- A transparent, click-through `NSPanel` (`FocusedWindowBorderPanel`, `nonactivating`, `ignoresMouseEvents`, `canBecomeKey = false`) draws a rounded-rect stroke.
- It is refreshed at the end of `layoutWorkspaces()`, so it stays in sync after every focus change, move, resize and workspace switch, and is hidden when AeroSpace is disabled or no window is focused.
- The window's AeroSpace `Rect` (top-left origin) is converted back to Cocoa coordinates and expanded by the border width so the stroke hugs the window from the outside.

## Config (all optional, feature off by default)

```toml
focused-window-border = false             # enable/disable
focused-window-border-color = '0xff12B981' # 0xAARRGGBB, same format as JankyBorders
focused-window-border-width = 4            # points
focused-window-border-opacity = 100        # 0-100, multiplies the color alpha
focused-window-border-radius = 10          # corner radius (see note)
```

## Known limitation

macOS does not expose a per-window corner radius via public API, and different apps use different radii (and recent macOS varies it further). `focused-window-border-radius` is therefore a single global value to tune. Suggestions for a robust per-window approach are very welcome — happy to iterate or take this to a discussion first per CONTRIBUTING.


alcibiadesc:
Added `focused-window-border-inset` (default 0) to address the corner-radius limitation noted above: macOS doesn't expose per-window corner radius, so instead of guessing it, the border can be pulled a few points into the window so its own rounded corner masks any mismatch (no gap). inset=0 keeps the previous fully-outside behavior.

tonisives:
This is great, and I have been using this for a day without any problems. Added one suggestion in https://github.com/alcibiadesc/AeroSpace/pull/1 about toml settings format

Issue #1883, opened 2026-01-02
Feature: `create-implicit-container-or-fail`


nikitabobko:
`aerospace move --boundaries-action create-implicit-container-or-fail`

Creating an implicit container can only fail in the following situation:
1. `enable-normalization-flatten-containers` is `true`
2. The newly created container orientation, layout and children size are the same of those of the previous container which got flattened by the normalization

When `enable-normalization-flatten-containers` is disabled it's proposed to print a tip to stderr:

```
$ grep enable-normalization-flatten-containers ~/.aerospace.toml
enable-normalization-flatten-containers = false

$ aerospace move --boundaries-action create-implicit-container-or-fail right > /dev/null
Tip: create-implicit-container-or-fail will never cause the move command to fail since enable-normalization-flatten-containers is disabled

$ echo $?
0
```

## References

- https://github.com/nikitabobko/AeroSpace/discussions/1827
- https://github.com/nikitabobko/AeroSpace/issues/183

Issue #1834, opened 2025-12-01
Add `aerospace cycle-size` command


nikitabobko:
Synopsis: `aerospace cycle-size [--axis (smart|smart-opposite|height|width)] (<size>%)...`

Examples:
```
aerospace cycle-size 33% 50% 66%
```

The meanings of `(smart|smart-opposite|height|width)` are the same as in `aerospace resize` command.

## Discussions

https://github.com/nikitabobko/AeroSpace/discussions/1827

stefanpinterBE:
i like it - i've seen this in Niri (a scrolling windows manager for Linux/Wayland)...
maybe this is also great for Ultrawide screens, if it would work when only a single window is on the worksprace

PR #2108, opened 2026-05-30
feat: cycle window size when moving against the edge

alcibiadesc:
## What

Opt-in: when a `move <direction>` command would hit the edge of a same-orientation split, cycle the focused window's **size** along that axis instead — 50% → 60% → 70% of the parent (configurable), wrapping back. Works on all four edges (left/right adjust the horizontal split, up/down the vertical split).

## Use case

With `move` bound to `alt-cmd-<arrow>`, pushing a window against a wall it already sits on can grow it in clean steps, without a separate resize binding.

## Off by default (opt-in)

This intentionally repurposes the `move`-at-edge behavior (and its `--boundaries-action`), so it would change established semantics if always on. It is therefore **off by default** and must be enabled explicitly. When there is no same-orientation split to resize, `resizeToggleAtEdge` returns `nil` and `move` falls through to its normal behavior regardless of the setting — so enabling it never breaks a move that would otherwise reorganize/relocate a window.

## Config

```toml
move-resize-toggle-at-edge = false          # opt-in
# Sizes (percentages of the parent the focused window occupies) the cycle steps through.
# Values 1-99, sorted automatically. The config model has no float type, hence integer percentages.
move-resize-toggle-ratios = [50, 60, 70]     # 1/2 -> 3/5 -> 7/10
```

## How it works

- `isAtExtremeEdge` walks the window's ancestors; it's at the edge when no same-orientation tiling container has a neighbour in that direction.
- `resizeToggleAtEdge` selects the window's slice in the nearest same-orientation tiles container and sets its `adaptiveWeight` to the next percentage strictly greater than the current ratio (wrapping), leaving siblings untouched; the layout pass re-normalizes.

### DIFF
--- a/Sources/AppBundle/command/impl/MoveCommand.swift
+++ b/Sources/AppBundle/command/impl/MoveCommand.swift
@@ -26,6 +26,14 @@ struct MoveCommand: Command {
                             return .succ
                     }
                 } else {
+                    // Only hijack the edge `move` for a size toggle when there is actually a
+                    // same-orientation split to resize. Otherwise fall through to the normal
+                    // moveOut (e.g. move-up in an all-horizontal layout must still move/reorganize).
+                    if config.moveResizeToggleAtEdge, isAtExtremeEdge(currentWindow, direction),
+                       let exit = resizeToggleAtEdge(window: currentWindow, direction: direction)
+                    {
+                        return exit
+                    }
                     return moveOut(window: currentWindow, direction: direction, io, args, env)
                 }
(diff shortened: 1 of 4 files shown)


nikitabobko:
The PR doesn't respect previous discussions

https://github.com/nikitabobko/AeroSpace/issues/1834
https://github.com/nikitabobko/AeroSpace/discussions/1827
https://github.com/nikitabobko/AeroSpace/issues/1883

PR #2109, opened 2026-05-30
feat: optional new-window-prevent-flicker

alcibiadesc:
## What

Adds an **opt-in** option to reduce the flash/jump when a new window opens: `new-window-prevent-flicker` (default `false`).

## Problem

When a new tiling window is detected it is bound to the tree but only positioned at the *end* of the refresh session. In the meantime macOS paints it at its native spawn position (usually screen center), and then everything snaps into place — a visible flash + reflow.

## How it works

When enabled, the instant a new tiling window on a visible workspace is registered (`getOrRegister`), it is moved offscreen to the monitor's bottom-right corner with a single non-blocking AX frame set, *before* macOS paints it at center. The same refresh session's `layoutWorkspaces()` then positions it directly into its tiled slot. Net: the window goes corner→slot instead of center→slot, suppressing the central flash. Reuses the existing corner-hide infrastructure.

## Honest scope / limitation

This does **not** animate or fully eliminate motion — Accessibility-based window management can't do smooth frame animation (each AX frame set is a blocking IPC round-trip, which is also why AeroSpace disables AX animations). It trades a center-flash for a smaller corner→slot motion. Whether that reads as "less flicker" is app/monitor-dependent, which is why it's **opt-in and default off**.

## Config

```toml
new-window-prevent-flicker = false  # set true to enable
```

## Notes
- Default off → zero impact unless enabled.
- Self-healing: the offscreen move is cancelled by the final placement; if a refresh is interrupted between them the window self-corrects on the next refresh.

### DIFF
--- a/Sources/AppBundle/config/Config.swift
+++ b/Sources/AppBundle/config/Config.swift
@@ -44,6 +44,7 @@ struct Config: ConvenienceCopyable {
     var startAtLogin: Bool = false
     var autoReloadConfig: Bool = false
     var automaticallyUnhideMacosHiddenApps: Bool = false
+    var newWindowPreventFlicker: Bool = false
     var accordionPadding: Int = 30
     var enableNormalizationOppositeOrientationForNestedContainers: Bool = true
     var persistentWorkspaces: OrderedSet<String> = []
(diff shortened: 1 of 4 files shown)


nikitabobko:
This is nonsense

PR #2113, opened 2026-05-31
Fix workspace stubs and window swap edge cases

dnnaji:
## Summary
- avoid fetching window titles unless a format expansion actually needs them
- fix `swapWindows` to read the second window index from the second window
- preserve the previous workspace as the stub when moving or summoning visible workspaces across monitors
- add two-monitor command tests and reset test-only monitor/focus callback state for isolation

## Why
This pulls forward and tightens fixes I had been carrying locally. The lazy title path still resolved `window.title` unconditionally, so list-window formatting could pay AX title cost even when unused. `swapWindows` also reused `window1.ownIndex` for both operands, breaking reverse-order swaps. For workspace moves between monitors, selecting the stub after mutating the visible workspace maps could choose the wrong empty workspace and leave monitor/workspace state unstable.

## Validation
- `swift test`
- `swift build --product aerospace -Xswiftc -warnings-as-errors`
- `swift build --product AeroSpaceApp -Xswiftc -warnings-as-errors`
- `swift build --target AppBundleTests -Xswiftc -warnings-as-errors`
- `git diff --check`

### DIFF
--- a/Sources/AppBundle/command/format.swift
+++ b/Sources/AppBundle/command/format.swift
@@ -24,8 +24,8 @@ struct WindowWithPrefetchedTitle {
     }
 
     private static func resolveWindow(_ window: Window, needsTitle: Bool) async throws -> Self {
-        let title: String = try await window.title
-        return .init(window: window, title: needsTitle ? title : nil)
+        let title = needsTitle ? try await window.title : nil
+        return .init(window: window, title: title)
     }
 
     static func forTest(window: Window, title: String?) -> Self {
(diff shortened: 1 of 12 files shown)


nikitabobko:
In fact, thanks for the PR! I am not sure why you closed it.

I have merged the first commit as-is: https://github.com/nikitabobko/AeroSpace/commit/25ccedf4093ce23dc311c1a13cd3db26d05df68f

And the second commit pointed me to a real problem: https://github.com/nikitabobko/AeroSpace/commit/19b5999d0dc2105c1e6e6478399a7cb8a7b9d95d

The third commit, I didn't yet have the energy to process it. The code you are touching is a big overdue and a technical debt with no tests covering it :(

Thanks!

PR #2120, opened 2026-06-08
Surface floating-window status: menu-bar indicators + %{window-is-floating} format variable

marcuswanner:
## Summary

Currently there's no at-a-glance way to tell whether a window is floating — you have to read `%{window-layout}` output. This adds a few small, complementary ways to surface floating state, mirroring how **fullscreen** state is already surfaced.

CLI: **`%{window-is-floating}` format variable** for `list-windows` / `test`, mirroring the existing `%{window-is-fullscreen}` boolean. It's derived from the canonical parent relation (`getChildParentRelation(child:parent:) == .floatingWindow` — the same source `%{window-layout}` uses to report `floating`), not the `Window.isFloating` helper. Parser possible-values, `test` expectations, and the `list-windows` / `test` docs are updated.

GUI: **Floating window marker in the tray dropdown.** In the per-workspace app list, any app that has a floating window in that workspace is prefixed with `⬚`, paralleling the existing surfacing of fullscreen state.

GUI: **Floating window indicator on the menu-bar workspace square.** Fullscreen workspaces already get an always-visible **dashed** border around their square (`MenuBarLabel.itemView`). This adds a **dotted**-border counterpart that appears on the focused workspace's square when the focused window is floating.

image

### DIFF
--- a/Sources/AppBundle/command/format.swift
+++ b/Sources/AppBundle/command/format.swift
@@ -162,6 +162,7 @@ extension FormatVar {
                 return switch f {
                     case .windowId: .success(.int(w.window.windowId))
                     case .windowIsFullscreen: .success(.bool(w.window.isFullscreen))
+                    case .windowIsFloating: isFloatingResult(w: w.window)
                     case .windowTitle: .success(.string(w.title.orDie("Title wasn't prefeched")))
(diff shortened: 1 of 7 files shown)


nikitabobko:
It doesn't make sense

Issue #260, opened 2024-06-08
Feature proposal: More sophisticated per-workspace normalizations to replicate dynamic tiling window managers


nikitabobko:
AeroSpace already provides two normalizations ("flatten-containers" and "opposite-orientation-for-nested-containers")

The current mental model of normalizations can be extended to allow more types of normalizations. To completely replicate dynamic tiling window managers, it should be possible to change normalizations with command on per-workspace basis

Normalizations to consider:
- `floating`. All windows on the workspace are floating. 
  Open question: should the normalization remember and restore layout from the workspace where this normalization wasn't enabled?
- `columns`
- `rows`
- `primay-others` ("tall" in Amethyst)
- `others-primary` ("tall-right" in Amethyst)
- `v-primary-others` ("wide" in Amethyst)
- `v-others-primary`

To consider: file with declarative hierarchical normalization description (something like yaml, but probably not yaml)

Related: #228 #248

nikitabobko:
> file with declarative hierarchical normalization description

The declarative file might look like this:
```
# columns
h_tiles
  others

# rows
v_tiles
  others

# primary-others
h_tiles:
  1
  v_tiles
    others

# others-primary
h_tiles
  v_tiles
    others
  1

# https://github.com/nikitabobko/AeroSpace/issues/261
v_tiles
  h_tiles
    1
    2
    3
  h_tiles
    others
    6
    5
    4
```

Windows fill in the number placeholders in order. Unfilled placeholders are "collapsed" (same as flatten-containers normalization). "others" is a special placeholder that always must be presented in the description. It "swallows" all the remaining windows

Unfortunately, `floating` layout is not representable in this model, but maybe it's fine

thewinger:
Would it be possible to mix tiles and accordion in this file?
I'm thinking in the Two-pane layout of Amethyst.
It would be something like:
```
# primary-others
h_tiles:
  1
  v_accordion
    others
```

nikitabobko:
There are no reasons to disallow it

zepocas:
Hi @nikitabobko, first of all thanks for this great project. 

Is there a timeline for this BSP feature? It's the only thing I'm missing atm, I would really appreciate this implementation.

dssste:
What would be the difference between bsp and opposite-orientation-for-nested-containers? Currently when I open a new window, then join-with the last window, it looks like bsp to me. In other words, if we auto assign new window as sibling of the currently focused window, we kind of have bsp already?

[...]

PR #2135, opened 2026-06-20
feat: add BSP normalization

OliverGuy:
Cf. #260.

## What
A new `enable-normalization-binary-tree` config option. When enabled, every tiling container is normalized to at most two children, with each level's orientation derived from its rectangle's aspect ratio, producing an i3/bspwm-style binary space partition.

## Why
Users coming from BSP-style window managers expect new splits to follow the longer edge automatically instead of piling siblings into a single container.

## How
`NormalizeBinaryTree()` recursively wraps surplus children into a nested container (`forceBinaryTree`), then sets each level's orientation and split position from its computed rect. Gated behind the new flag in `normalizeContainers()`; `setOrientation` is exposed on `TilingContainer` for it.

## PR checklist

### DIFF
--- a/Sources/AppBundle/command/impl/MoveCommand.swift
+++ b/Sources/AppBundle/command/impl/MoveCommand.swift
@@ -33,6 +33,14 @@ struct MoveCommand: Command {
                             return .succ
                     }
                 } else {
+                    let outOfLevelTarget = currentWindow.closestParent(hasChildrenInDirection: direction, withLayout: nil)
+                    if config.enableNormalizationBinaryTree, outOfLevelTarget == nil {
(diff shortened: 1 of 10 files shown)


whoislikemiha:
The full `./test.sh` gate currently fails on Swift 6.3.3 because `testBinaryTree_edgeMoveHitsBoundaryInsteadOfUnnesting` uses `try await` for a non-throwing call. With warnings-as-errors:

```
error: no calls to throwing functions occur within 'try' expression
```

I ran the complete gate after applying the SwiftFormat-compatible fix; build, tests, lint, Periphery, and generated-file checks all pass. Cherry-pickable commit: whoislikemiha/AeroSpace@bf5c86a43bf1f7b731423254a242b27bc18e8d40

Issue #143, opened 2024-01-29
window mistakenly spanning across different monitors


nikitabobko:
_Originally posted by @carlosflorencio in https://github.com/nikitabobko/AeroSpace/issues/101#issuecomment-1903744400_

Configuration:
- left monitor: built-in macbook monitor, main, 1800x1169
- right monitor: external Dell 1920x1080
- 'displays have separate spaces' is disabled

Steps:
1. layout `h_tiles`
2. Open a single window1
3. Open a second window2
4. Observe that half of the window2 is on left monitor and other half of the window2 is on right monitor. Like width/2 wasn't applied to the window2

nikitabobko:
@carlosflorencio could you please share your configuration?

I managed to reproduce the problem with the configuration mentioned in the issue description, but I'm just curious about what your configuration is, because I didn't manage to understand what is the key to reproduce this problem

carlosflorencio:
@nikitabobko sure, config: https://gist.github.com/carlosflorencio/93c8086372ad7ac99e731517e08a41f8

left monitor: external LG Ultra HD 1920x1080 (90°, vertical)
center monitor: external LG HDR 4K 1920x1080, main
right monitor: built-in macbook monitor, 1512x982

'displays have separate spaces' is disabled

nikitabobko:
But maybe we don't need to understand it. For some unknown reason setting the window's size before moving it 35143f77603bbf2a234a4bb972c7e462a8081ba1, fixes the problem in my configuration

I'd be happy if you could check 0.8.3 version and report if the bug is fixed on your machine

carlosflorencio:
Yep, will give it a try in the next couple of days 💪

carlosflorencio:
Can't reproduce the issue with v0.8.3, seems to work fine. Thanks!

PR #2167, opened 2026-07-11
Fix Device Hub (Xcode 27) compact-view flicker when tiled

FelixLisczyk:
Fixes the flicker described in #2137: AeroSpace rewrote every tiled window's AX frame on every relayout regardless of whether it had changed, which could loop with apps that don't accept the imposed frame and report a different one back (e.g. Xcode 27 Device Hub's compact view). See commit messages for details.

## PR checklist


FelixLisczyk:
While testing this fix, I noticed a separate, likely pre-existing glitch: Device Hub's tile briefly jumps to the top-left when resizing an adjacent tiled window. It's caused by a different code path (the deliberate size→position→size write order in `setFrame`, kept that way for #143/#335) and is out of scope for this PR.

PR #2174, opened 2026-07-13
Restore windows after native fullscreen

Damecek:
Refs #562.

## PR checklist

### DIFF
--- a/Sources/AppBundle/command/impl/MacosNativeFullscreenCommand.swift
+++ b/Sources/AppBundle/command/impl/MacosNativeFullscreenCommand.swift
@@ -29,12 +29,12 @@ struct MacosNativeFullscreenCommand: Command {
                             "Tip: use --fail-if-noop to exit with non-zero exit code"))
             }
         }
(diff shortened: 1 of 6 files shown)


Damecek:
If this approach needs to be reimplemented:

> Implement a minimal fix for #562 while keeping native-fullscreen windows detached from the tiling tree. On entry, save `BindingData` plus conservative parent, sibling, layout, and weight anchors. On exit, restore the original parent, index, and weight only if the relevant branch is unchanged, including the single-child normalization that can flatten or replace the original parent; otherwise retain the existing MRU relayout fallback. Share the transition logic between state detection and the `macos-native-fullscreen` command, avoid fullscreen flags or special cases in `move`, `focus`, and generic layout consumers, and add focused tests for accordion order, tile weights, both entry paths, normalization, and safe fallback after tree changes.

Damecek:
https://github.com/user-attachments/assets/4e364f8b-64b9-4fb0-a372-9763e8d43426

Issue #726, opened 2024-11-18
When changing focus to Chrome window A, focus is immediately redirected to Chrome window B


samuela:
**Describe the bug**
When changing focus to a Google Chrome window, either via `workspace` or `focus-monitor`, I am finding that Chrome (or something else) then immediately redirects focus to a different Chrome window. I can more clearly see this focus change thanks to `on-focus-changed = ['move-mouse window-lazy-center']`.

I will attempt to reproduce on my personal machine and post a video. I am on a multi-monitor setup with macOS Sequoia 15.1 and AeroSpace v0.15.2-Beta b6cf82771f245ab7349a93baf8709e171537ff58.

**Related issues you could find**
n/a

samuela:
> I think you'd need to set Displays have separate Spaces under system settings → Desktop & Dock to off.

Interesting, I'll give this a shot

> To get this working with the prior setting set to on, I think one needs to use additional private APIs:

Not sure I follow... what's the reasoning here?

nikitabobko:
If this issue happens only when you change focus for Chrome windows that are on different monitors than https://github.com/nikitabobko/AeroSpace/issues/101 is probably a cause. In that case, the current workaround is to disable "Displays have separate Spaces"

Otherwise, I'm not sure what the problem is, I'd appreciate if you re-check if the problem is reproducible in the latest version. 0.15.2 is pretty old

I'm closing the issue, but feel free to provide more details if the the issue is still relevant

Issue #591, opened 2024-10-14
Wrong windows gets focused when dealing with multiple windows of the same application


shoplab7:
I have 3 displays. When I have 2 windows of the same app on 2 of them and try to use focus-monitor next/prev, wrong window gets focus. 
For example, I have
1 - browser
2 - vscode
3 - vscode
(1, 2, 3 are displays)
I start from display1, use "focus-monitor --wrap-around  next" 3 times and get back to the display1. (all focus commands work completely fine)
then try to use "focus-monitor --wrap-around  prev" once. display3 gets focused for a split of a second, and then focus switches to display2 instead of just focusing on display3.

<details>
  <summary>debug-windows</summary>

```
com.googlecode.iterm2.window.440 windowId: 440
com.googlecode.iterm2.window.440 workspace: 2
com.googlecode.iterm2.window.440 treeNodeParent: Workspace(name: '2', isVisible: 'true', isEffectivelyEmpty: 'false', doKeepAlive: 'true')
com.googlecode.iterm2.window.440 recognizedAsDialog: true

com.googlecode.iterm2            AXRole: Optional(AXApplication)
com.googlecode.iterm2            AXTitle: Optional(iTerm2)
com.googlecode.iterm2            AXFunctionRowTopLevelElements: [
com.googlecode.iterm2            ]
com.googlecode.iterm2            AXFrame: nil
com.googlecode.iterm2            AXFocusedUIElement: AXUIElement(windowId=440, title=nil, role="AXTextArea", subrole=nil)
com.googlecode.iterm2            AXFrontmost: Optional(1)
com.googlecode.iterm2            AXExtrasMenuBar: nil
com.googlecode.iterm2            AXMainWindow: AXUIElement(windowId=440, title="iTerm2 Hotkey Window", role="AXWindow", subrole="AXStandardWindow")
com.googlecode.iterm2            AXFocusedWindow: AXUIElement(windowId=440, title="iTerm2 Hotkey Window", role="AXWindow", subrole="AXStandardWindow")
com.googlecode.iterm2            AXMenuBar: Optional(<AXUIElement 0x60000358d140> {pid=739})
com.googlecode.iterm2            AXWindows: [
com.googlecode.iterm2                AXUIElement(windowId=440, title="iTerm2 Hotkey Window", role="AXWindow", subrole="AXStandardWindow")
com.googlecode.iterm2            ]
com.googlecode.iterm2            AXSize: nil
[...]

Issue #229, opened 2024-05-16
Incorrect window and workspace focus with multiple monitor


paulodiovani:
If I have more than one of the same application distributed across monitors, when I switch workspaces, the last opened window of that application gets focused instead of the one on the focused workspace/monitor.

## Detailed explanation

- I have two monitors
- My primary monitor hosts workspaces 1-9, while the secondary have workspace 10
- I have keybindings to switch workspaces using `cmd` (e.g. `cmd-1 = workspace 1`)

### Steps to reproduce

1. Open an application (e.g. Firefox) on the `workspace 1` (primary monitor)
2. Open another window of the same application in `workspace 10` (secondary monitor)
3. Open another application (e.g. Terminal) on the `workspace 2` (primary monitor)
4. Switch to `workspace 1` (e.g. type `cmd-1`)
5. Switch to `workspace 2` (e.g. type `cmd-1`)
6. Switch to `workspace 1` I need your GD account to start enabling the rest of the permissions

### Expected behavior

Window on `workspace 1` is focused

### Current behavior

Window on `workspace 10` is focused, even though I focused `workspace 1`

### Workaround

By focusing `workspace 1` again (e.g. typing `cmd-1` twice) the correct window gets focus.

## System information

```
OS: macOS 14.1.2 23B2091 arm64
Host: Mac15,6
Kernel: 23.1.0
Uptime: 38 days, 4 hours, 16 mins
Packages: 158 (brew)
Shell: zsh 5.9
Resolution: 1920x1080, 1512x982
DE: Aqua
WM: Quartz Compositor
WM Theme: Purple (Dark)
Terminal: tmux
CPU: Apple M3 Pro
GPU: Apple M3 Pro
Memory: 3408MiB / 18432MiB
```

```
aerospace CLI client version: 0.10.0-Beta 4fe8276262b8df077ffface84c3f93f1b832b8ac
AeroSpace.app server version: 0.10.0-Beta 4fe8276262b8df077ffface84c3f93f1b832b8ac
```

nikitabobko:
It's a known and unfortunate problem. It's a bug in Apple API. The workaround is to disable "Displays have separate spaces"

nikitabobko:
Duplicate of #101

PR #2179, opened 2026-07-14
Fix #101: focus the correct cross-monitor window via private SkyLight API

### LINKED ISSUE #101: Bug: sometimes a window on the focused monitor is focused when a window of the same app on a different monitor was requested to focus
Hey Nikita, I'm experiencing the following behavior and I've no idea if it's intended or not. Whenever I've two instances of an application open on different monitors/workspaces. It will focus the instance of the app I last used. I'll try to explain it with an example:

I've a three monitor setup: `1 | 2 | *0`

* Workspace 1 contains a kitty window.
* Workspace 2 contains a kitty window
* Workspace 0 contains something else, say firefox.

I've the following keybinds to go the specific workspaces:
```
alt-1 = 'workspace 1'
alt-2 = 'workspace 2'
```

I want to go to workspace 1 so I press `alt-1`. I end up at workspace 2. If I press `alt-1` again I'll go to workspace 1 as expected. 

Is this working as intended? I would like to be able to actually focus a specific window/instance of an app.

Love the app so far, is there a way we can buy you a coffee or beer?

White7292:
## What this fixes

Closes #101.

When **"Displays have separate Spaces" is enabled with more than one monitor**, focusing one window of an app that also has windows on another monitor sometimes focuses the *wrong* window — a different window of the same app on another monitor — and drags that monitor to a different workspace. It's intermittent, and it's been reported many times (#229, #591, #726, discussions #1476/#1705/#1937).

## Root cause

`nativeFocus` uses the public path `kAXMain` + `AXRaise` + `NSRunningApplication.activate(...)`. The problem is the last step: **`activate` is app-level, not window-level.** Per AppKit, activation brings the app's *own* main/key window forward — and for a multi-window app that key window is often the one the user most recently touched, which may live on a different monitor. So macOS makes *that* window key instead of the one we raised, AeroSpace observes focus land there, and follows it — switching the other monitor's visible workspace. The public API has no way to say "activate app X **and** make window W key."

## The fix

Use the private SkyLight (WindowServer) API to name the exact `CGWindowID` that becomes key — the same long-standing technique used by **yabai** and **Amethyst** (origin: [Hammerspoon#370](https://github.com/Hammerspoon/hammerspoon/issues/370#issuecomment-545545468)): `_SLPSSetFrontProcessWithOptions` + two synthesized `SLPSPostEventRecordTo` key-window records + `AXRaise`.

Design choices aimed at the concerns raised in the issue thread:

- **All private-API usage is contained in the existing `PrivateApi` C module**, right next to `_AXUIElementGetWindow`. Swift sees exactly one new function, `aeroMakeKeyWindow(pid, wid)`; the byte-poking and the `GetProcessForPID` deprecation live in the `.m`.
- **Best-effort crash guard** (implementing the idea from [#101 (comment)](https://github.com/nikitabobko/AeroSpace/issues/101#issuecomment-2568238564)): a `UserDefaults` flag is set immediately before the private call and cleared immediately after. If it's still set at the next launch — a previous session likely died inside the private call — the private path is disabled and AeroSpace falls back to the existing public-API behavior. It is **best-effort**: `UserDefaults` flushes asynchronously (a very fast crash may not be caught) and a hard kill during the brief in-use window could disable it spuriously. Both failure modes degrade to today's upstream behavior, so neither introduces new harm. Happy to harden it — see Open questions.
- **Narrowly gated**: the private path runs *only* when `NSScreen.screensHaveSeparateSpaces` **and** `monitors.count > 1` **and** the crash guard hasn't disabled it. With separate Spaces off, or a single monitor, the existing paths are byte-for-byte unchanged.

Net diff: +77 lines, 5 files. No new dependency; SkyLight is a system framework, linked only on the `PrivateApi` target.

## Testing

- Full `./test.sh` passes locally: builds clean under `-Xswiftc -warnings-as-errors`, swift tests, `swiftformat`, `swiftlint`, `periphery` (no unused code), `generate.sh`, and `check-uncommitted-files.sh`.
- **The behavior was verified against the real bug on hardware** (macOS 26.3, 3 monitors, separate Spaces on) using a **byte-identical `@_silgen_name` Swift prototype** of this SLPS sequence: I reproduced the exact failure first — armed a same-app window on a hidden workspace of monitor A, then focused that app's window on monitor B, and monitor A flipped away — then confirmed the fix holds monitor A and focus lands on the requested window. **This PR re-expresses that same sequence in the `PrivateApi` C module.** The C version builds clean and emits the identical WindowServer events, but I have not re-run the built C binary on the rig.

Honest scope: verified by one person on one setup, and the bug is intermittent by nature (it depends on the app's most-recently-key window), so wider testing is welcome — especially on 2-monitor setups and with non-Electron apps.

## Prior art / credit

- The crash-guard idea is [nikitabobko's own](https://github.com/nikitabobko/AeroSpace/issues/101#issuecomment-2568238564).
- The SkyLight sequence follows yabai (`window_manager_make_key_window`) and Amethyst ([PR #1530](https://github.com/ianyh/Amethyst/pull/1530), which fixed the same cross-monitor symptom).
- `titouv` sketched an earlier version on a fork; this reworks it onto the `PrivateApi` module and adds the crash guard + gating.

## Open questions (happy to adjust)

- **Hardening the crash guard**, if you'd prefer it stronger than best-effort: force a synchronous flush around the private call, only disable after N repeated crashes, and/or log one line when it disables. It's currently a single shared flag while focus bodies run on per-app threads, so overlapping private focus calls could mis-account — a per-call counter would fix that. I kept it minimal since it's your design idea.
- Should disabling surface a one-time dialog to the user (as you floated in the issue)? Left silent for now to keep the change small.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -35,6 +35,11 @@ let package = Package(
             name: "PrivateApi",
             path: "Sources/PrivateApi",
             publicHeadersPath: "include",
+            linkerSettings: [
+                // _SLPSSetFrontProcessWithOptions / SLPSPostEventRecordTo (used by aeroMakeKeyWindow for
+                // the issue #101 cross-monitor focus fix) live in the private SkyLight framework.
+                .unsafeFlags(["-F/System/Library/PrivateFrameworks", "-framework", "SkyLight"]),
+            ],
         ),
         .target(
             name: "Common",
--- a/Sources/AppBundle/initAppBundle.swift
+++ b/Sources/AppBundle/initAppBundle.swift
@@ -3,6 +3,7 @@ import Common
 import Foundation
 
 @MainActor public func initAppBundle() {
+    checkFocusPrivateApiCrashGuard()
     Task.startUnstructured {
         initTerminationHandler()
         unsafe _isCli = false
@@ -32,6 +33,20 @@ import Foundation
             smartLayoutAtStartup()
             _ = await config.afterStartupCommand.run(.defaultEnv, .emptyStdin)
         }
+    }
+}
+
+// https://github.com/nikitabobko/AeroSpace/issues/101 (idea: comment 2568238564)
+// Best-effort guard for the private SkyLight focus API used in MacApp.nativeFocus. If the in-use flag
+// is still set at startup, a previous session most likely died inside the private call, so disable the
+// private path and fall back to the public-API behavior. Best-effort because UserDefaults flushes
+// asynchronously (a very fast crash may not be caught) and a hard kill during the brief in-use window
+// could disable it spuriously. Once disabled it stays disabled until the key is cleared, e.g.
+// `defaults delete bobko.aerospace private-focus-api-disabled-after-crash`.
+@MainActor private func checkFocusPrivateApiCrashGuard() {
+    if UserDefaults.standard.bool(forKey: privateFocusApiInUseKey) {
+        UserDefaults.standard.set(true, forKey: privateFocusApiDisabledKey)
+        UserDefaults.standard.set(false, forKey: privateFocusApiInUseKey)
     }
 }
 
--- a/Sources/AppBundle/tree/MacApp.swift
+++ b/Sources/AppBundle/tree/MacApp.swift
@@ -1,5 +1,11 @@
 import AppKit
 import Common
+import PrivateApi
+
+// UserDefaults keys for the best-effort private-focus-API crash guard (see checkFocusPrivateApiCrashGuard
+// in initAppBundle.swift). https://github.com/nikitabobko/AeroSpace/issues/101
+let privateFocusApiInUseKey = "private-focus-api-in-use"
+let privateFocusApiDisabledKey = "private-focus-api-disabled-after-crash"
 
 // Potential alternative implementation
 // https://github.com/swiftlang/swift-evolution/blob/main/proposals/0392-custom-actor-executors.md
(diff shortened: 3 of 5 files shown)


tariqc80:
Thank you for working on this proposed fix.  I get this behavior on my dual monitor setup whether I have separate spaces on or off.  I've been living with it since I switched to Aerospace from Yabai.  This fix would make me very happy.

tariqc80:
@White7292 I built this branch and tested on my setup.  I have two monitors running macsOS 26.5.2.  

This is what I observed:

Setup: 

I opened a terminal window on workspace 1 on Monitor A, a browser window on visible workspace 3 on Monitor B, then opened another browser window on a workspace 2 of Monitor A.  I focused workspace 1 on monitor A so workspace 2 is hidden.

Test:

I dragged my mouse from Monitor A to Monitor B and I observed Monitor A switch to the browser window on workspace 2.

This seems to only happen if I had previously focused the browser window on workspace 2 and then focus workspace 1 before dragging to Monitor B.  

Note:

This does not happen when using the hotkeys to switch workspaces, only the mouse.   On nikitabobko:main using the hotkeys does exhibit the unwanted behavior i.e. it will switch to the unfocused workspace 2 

I'm wondering if it's possible to include focusing with the mouse in this fix?

adamtrnguyen:
Thanks for this, it works. Built it on v0.21.3-Beta.

I've got 4 monitors with separate spaces on. Had two kitty windows on different monitors and focus kept landing on whichever one I'd opened first instead of the one I was pointing at. That's fixed now.

Mine was happening through focus-follows-mouse rather than hotkeys, so that path seems covered too.

PR #2182, opened 2026-07-15
Fix BSP two-window split orientation (QUB-61)

qubeio:
## Summary
- `bspSplitOrientation()` decided split orientation from `lastAppliedLayoutVirtualRect`, a cache only populated after a layout pass runs. Two windows arriving in the same refresh (startup, quick app launches) hit the nil-rect fallback, which alternated from the parent orientation — producing an intermittent top/bottom split on landscape monitors instead of side-by-side. A narrow aspect-ratio dead zone and invisible-workspace insertions had the same failure mode.
- Add `computeVirtualSlotRect()` (`Sources/AppBundle/layout/computeSlotRect.swift`), a pure function that derives the slot a node would occupy from the current tree shape (container weights/orientation), with no dependency on a layout pass having run.
- Rewrite `bspSplitOrientation` to use a single ratio-vs-`autoSplitThreshold` cutoff instead of the old band-with-fallback logic, with a defensive fallback to monitor shape for degenerate rects.

## Test plan
- [x] New `BSPSplitOrientationTest` (5 tests): two-window regression (no layout pass between insertions), 4-window fibonacci spiral structure, `preferred-split-direction` override, `computeVirtualSlotRect` with unequal weights, and accordion pass-through.
- [x] `swift test` — all 130 tests pass, no regressions.
- [x] `./format.sh` — clean, no changes needed.
- [x] `./lint.sh` — clean, no unused code.

Reference: [QUB-61](https://linear.app/qubeio/issue/QUB-61/bsp-derive-split-orientation-from-computed-slot-geometry-fixes)


qubeio:
Opened in error against the wrong repo (should target qubeio/AeroSpace fork, not upstream). Closing.

PR #2184, opened 2026-07-16
debug-windows: add --app-bundle-id flag to dump unregistered AX windows

### LINKED ISSUE #1235: Make `debug-windows` command to capture windows that are not treated as windows from AeroSpace perspective
### _

- [x] I read https://github.com/nikitabobko/AeroSpace/issues/947

rvaccone:
Closes #1235

`debug-windows` can only dump windows AeroSpace already registered, which is a problem when the bug is that AeroSpace _didn't_ register something — or registered it as a popup, which can't be focused, so recording mode never sees it and `list-windows` never shows its id.

This adds `aerospace debug-windows --app-bundle-id <id>`. It dumps every AX element the app returns in `kAXWindowsAttribute` without the `windowOrNil` filtering, plus registered windows that are no longer enumerated there. Each entry gets:

- `Aero.debug.registered` — whether the element is in the tree; registered ones also get the usual `Aero.workspace`/`Aero.treeNodeParent`/`Aero.on-window-detected`
- `Aero.debug.windowIdFailed` — for elements dropped because `_AXUIElementGetWindow` failed or returned `kCGNullWindowID` (after 0f6b2e78 these can't be dumped at all)
- `Aero.debug.absentInAxWindowsAttr` — registered windows the app no longer enumerates
- `Aero.AxUiElementWindowType` + `_isDialogHeuristic`, computed in the same pass, so the output can be dropped into `axDumps/` as-is

If the app isn't registered in AeroSpace at all (like FL Studio in #1234), the error reports its pid and activationPolicy.

All AX access happens in a single pass on the app's runloop thread. Window levels are prefetched on the main actor since `getWindowLevel` is `@MainActor`. `--window-id` output is unchanged (shared parts extracted into helpers). Docs, generated help, and completion grammar updated.

Example against Finder — it captures the desktop element that `windowOrNil` exists to filter out:

```
com.apple.finder.nil-window-id-0 |||   "AXDescription" : "desktop",
com.apple.finder.nil-window-id-0 |||   "AXRole" : "AXScrollArea",
com.apple.finder.nil-window-id-0 |||   "Aero.AxUiElementWindowType" : "popup",
com.apple.finder.nil-window-id-0 |||   "Aero.debug.registered" : false,
com.apple.finder.nil-window-id-0 |||   "Aero.debug.windowIdFailed" : "_AXUIElementGetWindow failed or returned kCGNullWindowID. AeroSpace ignores such AX windows",
```

Tested with a `--read-only` debug server against a dozen running apps: registered windows, the Finder desktop element, an unregistered accessory app, a not-running app, and `--window-id` for regressions. `swift test` passes, swiftformat clean.

### DIFF
--- a/Sources/AppBundle/command/impl/DebugWindowsCommand.swift
+++ b/Sources/AppBundle/command/impl/DebugWindowsCommand.swift
@@ -24,6 +24,12 @@ struct DebugWindowsCommand: Command {
     /*conforms*/ let shouldResetClosedWindowsCache = false
 
     func run(_ env: CmdEnv, _ io: CmdIo) async -> BinaryExitCode {
+        if args.windowId != nil && args.appBundleId != nil {
+            return .fail(io.err("--window-id is incompatible with --app-bundle-id"))
+        }
+        if let appBundleId = args.appBundleId {
+            return await dumpAppWindowsDebugInfo(appBundleId, io)
+        }
         if let windowId = args.windowId {
             guard let window = Window.get(byId: windowId) else {
                 return .fail(io.err("Can't find window with the specified window-id: \(windowId)"))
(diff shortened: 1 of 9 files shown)


rvaccone:
Bumping this. CI is green and the change is scoped to the `debug-windows` command. Happy to rebase or adjust anything if needed.

PR #2185, opened 2026-07-17
QUB-78 Bind BSP-safe fullscreen

qubeio:
## Summary
- bind `alt-comma` to the tree-safe `fullscreen` overlay in the default config example
- keep raw tiles/accordion layout switching disabled for BSP workspaces
- document that fullscreen preserves and restores the underlying tiling tree

## Verification
- `swift test --filter ConfigTest` — 29 tests passed
- `./build-docs.sh` — passed
- `aerospace reload-config --dry-run --no-gui` — passed
- live binding test on workspace A/window 3517: false → true → false; root remained `bsp`; before/after debug-tree SHA-256 matched (`f7e03785ff3d2cb3f1680c4189a220c53f4bf16b0fba267fbeccf15f522f7d1e`)

Reference: QUB-78


qubeio:
Opened against upstream by mistake; this change belongs to the qubeio fork.

PR #2186, opened 2026-07-17
fix(config): avoid crash on per-workspace -> per-monitor accordion padding

KSiig:
## Summary

`adbdf4be` introduced per-workspace -> per-monitor support for `accordion-padding` but the recursive call into `parseDynamicValue` did:

```swift
parseDynamicValue(value, ofType: T.self, parseSimpleType(array.last!, ofType: T.self)!, backtrace, &c)
```

In this scope `array` is the parameter of `parsePerWorkspaceValues` (i.e. `items` from the outer `parseDynamicValue` call, which is `outer.dropLast()`). For the canonical 2-element outer form `[workspace_table, default]` the second-to-last element is the workspace table itself, not an Int — `parseSimpleType` returns nil and the trailing `!` crashes.

## Repro (before)

```toml
accordion-padding = [
    { workspace = { '2' = [{ monitor = { 'LG ULTRAWIDE' = 384 } }, 16] } },
    16
]
```

Hard crash inside `parsePerWorkspaceValues` on the very first item.

## Fix

Derive the recursive fallback from the inner array's trailing item (which is what `parseDynamicValue` actually uses for the inner default anyway). If the inner array has no trailing default, surface a clear parse error:

```
[ERROR] accordion-padding[0].workspace.2: The last item in the inner array must be of type Int
```

The fallback only matters in `parseDynamicValue`'s error paths, so this change is invisible to valid configs.

## Tests

Added `testParseAccordionPadding` in `ConfigTest.swift` covering:

1. The user's bug case (per-workspace -> per-monitor)
2. The dotted-key equivalent `{workspace.2 = [...]}`
3. Mixed per-workspace-per-monitor + per-workspace-constant + outer default
4. Plain constant and per-workspace-only forms (regression)
5. The new error message for an inner array missing a trailing default

Verified the parser logic against TOMLDecoder v0.4.4 in a standalone Swift harness (8/8 passing). Could not run the full AppBundle tests on Linux due to macOS-only headers, so this needs a green `swift test` on macOS before merging.

Closes the crash reported after `adbdf4be`.


nikitabobko:
Low quality PR

PR #2189, opened 2026-07-17
feat(accordion): per-workspace and per-workspace -> per-monitor accordion padding

KSiig:
## Summary

Adds per-workspace accordion padding support, then layers per-workspace -> per-monitor override on top, then fixes a crash introduced by the second commit.

Three feature commits plus one fix, stacked in dependency order:

1. `feat(accordion): Add per-workspace padding` -- parseGaps-style `accordion-padding = [{ workspace = { '2' = 384 } }, 16]`
2. `chore(build): Updated swift-version` -- build tooling bump
3. `Add per-workspace -> per-monitor accordion padding` -- extends the per-workspace entry so its value can itself be a per-monitor dynamic: `{ workspace = { '2' = [{ monitor = { 'LG ULTRAWIDE' = 384 } }, 16] } }`
4. `fix(config): avoid crash on per-workspace -> per-monitor accordion padding` -- fix for the crash that (3) introduced

## Why the fix commit

(3) added a recursive call into `parseDynamicValue` from inside `parsePerWorkspaceValues`:

```swift
parseDynamicValue(value, ofType: T.self, parseSimpleType(array.last!, ofType: T.self)!, backtrace, &c)
```

`array` here is the parameter of `parsePerWorkspaceValues`, which is `items` from the outer `parseDynamicValue` call (i.e. `outer.dropLast()`). For the canonical 2-element outer form `[workspace_table, default]` the second-to-last element is the workspace table itself, not an `Int`. `parseSimpleType` returns `nil` and the trailing `!` crashes.

Fix derives the recursive fallback from the inner array's trailing item instead -- which is what `parseDynamicValue` uses for the inner default anyway -- and surfaces a clear parse error if the inner array has no trailing default:

```
[ERROR] accordion-padding[0].workspace.2: The last item in the inner array must be of type Int
```

The fallback only matters in `parseDynamicValue`'s error paths, so this change is invisible to valid configs.

## Test coverage

`testParseAccordionPadding` in `ConfigTest.swift` covers:

- The bug case (per-workspace -> per-monitor)
- The dotted-key equivalent `{workspace.2 = [...]}`
- Mixed per-workspace-per-monitor + per-workspace-constant + outer default
- Plain constant and per-workspace-only forms (regression)
- The new error message for an inner array missing a trailing default

## Squashing

Happy to squash to a single commit if the maintainer prefers; the four-commit stack is here so each step is reviewable in isolation.


nikitabobko:
Blocked for spamming the same PR https://github.com/nikitabobko/AeroSpace/pull/2186

PR #2198, opened 2026-07-23
focusFollowsMouse: keep focus while the mouse is within the focused window's real frame

zannis:
## Problem

With two windows tiled 50/50, if the left window has an AX minimum size that makes it physically wider than its tile, it overlaps the neighbor. Moving the mouse over that overhang (e.g. at 51% of the screen width) switches focus to the covered neighbor, even though the focused window is what's visually under the mouse (it's on top of the z-order).

Root cause: for tiled windows, focus-follows-mouse resolves the window under the cursor via `findWindowRecursively`, which only consults `lastAppliedLayoutPhysicalRect` — the rect the layout *requested*, not the frame the window actually settled on.

## Fix

Before resolving a new window under the mouse, check whether the cursor is still inside the currently focused window's real AX frame. If it is, keep the focus. Once the mouse leaves the real frame, behavior is unchanged.

The resolution logic is extracted from the `NSEvent` monitor closure into `CGPoint.windowUnderMouseToFocus(in:)` so it can be covered by a test (`FocusFollowsMouseTest`). The test is verified to fail without the fix.

### DIFF
--- a/Sources/AppBundle/mouse/focusFollowsMouse.swift
+++ b/Sources/AppBundle/mouse/focusFollowsMouse.swift
@@ -29,26 +29,36 @@ import AppKit
             if await isAxWindowUnderMouse(location) == false { return }
             try checkCancellation()
             let workspace = location.monitorApproximation.activeWorkspace
-            var window: Window? = nil
-            for child in workspace.floatingWindowsContainer.mruChildren {
-                try checkCancellation()
-                guard let child = child as? Window else { continue }
-                guard let rect = try await child.getAxRect(.cancellable) else { continue }
-                if rect.contains(location) {
-                    window = child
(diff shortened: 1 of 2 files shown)


nikitabobko:
> Moving the mouse over that overhang (e.g. at 51% of the screen width) switches focus to the covered neighbor, even though the focused window is what's visually under the mouse (it's on top of the z-order).

I don't consider it a bug. It's one of the possible behaviors

zannis:
it does make Aerospace unusable though for a lot of apps like Whatsapp which has fixed width + Browser plugins which extend past the window borders? would you consider adding this behind a flag? - Really impractical to use Aerospace on the day to day as it stands now

Issue #491, opened 2024-09-03
Feature Request: Implement `list-windows --dfs-order` to Query Workspace Window Structure in DFS Order


Perlten:
Now that we have the `focus --dfs-index <index>` [option](https://github.com/nikitabobko/AeroSpace/pull/464), it would be fantastic to extend this by allowing us to query the structure of a workspace's windows from a DFS perspective. A command like `list-windows --dfs-order` would be ideal for this purpose. Personally, I would use this feature to integrate with SketchyBar, displaying the windows in a way that mirrors workspace organization.

The output could follow the same format as the existing `list-windows` command, but it should be ordered as if a DFS (Depth-First Search) was performed on the workspace. See the example below.

h_tiles
├── window (index: 0)
├── v_tiles
&nbsp;|&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;├── window (index: 1)
&nbsp;|&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;└── window (index: 2)
├── window (index: 3)

I realize this is similar to the functionality proposed in https://github.com/nikitabobko/AeroSpace/issues/16, but I believe this request is still valid. It specifically focuses on the DFS ordering of windows, disregarding how they are otherwise structured within the tree.

## Checklist

- [x] I've searched for duplicates. My search queries were (please don't forget to search in closed issues too):
  - window-list
  - tree
  - feature-request

- [x] I've read the documentation https://nikitabobko.github.io/AeroSpace/guide & https://nikitabobko.github.io/AeroSpace/commands
- [x] I've searched in the documentation. My search queries were:
  - window-list
  - tree

nikitabobko:
**(1)** I'd prefer a different interface: `aerospace list-windows --sort dfs`

Because it can evolve to more options in the future: 
```
aerospace list-windows [--sort (dfs|pid|window-id|window-title|app-name)...]
```

`--sort dfs` should consider monitors and workspaces being part of the bigger tree. For example:
```
monitor 1
  workspace 1
    h_tiles
      window 1
      window 2
  workspace 2
    h_tiles
      window 3
monitor 2
  ...
```

**(2)** Feature interaction with floating windows is unclear. `focus --dfs-index` treats floating windows as part of the tree, but it happens only because `focus` command works this way in general. Until now, the `focus` command was the only command doing so. It's yet unclear to me whether `list-windows` should be a such command as well.

[...]

PR #2207, opened 2026-08-01
Implement list-windows --sort-by

kim-raaschou:
Implements #491

## PR checklist


kim-raaschou:
Two things that aren't obvious from the diff.

`--sort-by dfs` reuses the placement logic from `focus`, so it makes two AX rect reads per floating window. It's gated on the flag and on floating windows existing, so plain `list-windows` is unchanged, but a status bar polling with `--sort-by dfs` will pay it. The alternative was to duplicate the geometry rather than share it, which seemed worse.

Cross-monitor ordering is implemented but untested — `monitors` is hardcoded to a single `testMonitor` under `isUnitTest`, so the harness can't express two. Happy to extend it if you'd rather have that covered.

PR #2211, opened 2026-08-04
fix(balance-sizes): preserve total weight instead of resetting to a constant

### LINKED ISSUE #1837: `balance-sizes`, `resize` commands don't interact in the expected way when combined together in the config
Steps:
1. Layout windows in the following way:
```
h_tiles
  window1 (size = 50%)
  window2 (size = 50%)
```

2. Add the following lines to the config:
```
[mode.main.binding]
    alt-1 = ['balance-sizes', 'resize width 200']
    alt-2 = 'resize width 200'
```

Expected: `alt-1` and `alt-2` do the same
Actual: The hotkeys produce different results

## Discussions

https://github.com/nikitabobko/AeroSpace/discussions/1835

MsfPablo:
## Summary

Fixes #1837. `balance(_:)` assigned a hardcoded weight of `1` to every child, discarding whatever total the container's weights previously summed to.

Weights are only converted to real pixel sizes by the layout pass, which runs once after a whole binding's command list finishes. So when `balance-sizes` and `resize` are chained in the same binding, `resize`'s delta calculation still operates against the pre-existing (larger) weight scale of sibling containers, while the just-balanced children were reset to a tiny constant (`1`) — producing a much wider window than running the same two commands as separate keybind presses.

## Fix

Redistribute the container's existing weight total evenly across its children (`total / children.count`) instead of assigning an arbitrary constant, so later commands in the same list see a consistent scale.

## Test plan

- Updated `testBalanceSizes` (total 1+2+3=6 → each child becomes 2, not the old hardcoded 1)
- Added `testBalanceSizes_nestedContainerPreservesItsOwnTotal`, confirming each nested container's total is redistributed independently
- Added `testBalanceSizesThenResize_behavesLikeResizeAlone`, the regression test for the exact chained-command scenario from the issue

I couldn't run the test suite itself in my environment — `swift test` fails with `error: no such module 'XCTest'`, which I confirmed is pre-existing on a clean `main` checkout with no changes at all (I only have Xcode Command Line Tools, not a full Xcode install). `swift build` (the actual app code) compiles cleanly with this change. Happy to have CI confirm the test suite.

### DIFF
--- a/Sources/AppBundle/command/impl/BalanceSizesCommand.swift
+++ b/Sources/AppBundle/command/impl/BalanceSizesCommand.swift
@@ -15,11 +15,21 @@ struct BalanceSizesCommand: Command {
 
 @MainActor
 private func balance(_ parent: TilingContainer) {
+    switch parent.layout {
+        case .tiles:
+            // Redistribute the weights that the container already has, instead of assigning an arbitrary constant.
+            // Weights are only converted to real sizes by the layout pass, which happens once after the whole list of
+            // commands of a binding has run. Preserving the total keeps the weights meaningful for commands that run
+            // later in the same list (e.g. `resize`) https://github.com/nikitabobko/AeroSpace/issues/1837
+            let total = CGFloat(parent.children.sumOfDouble { $0.getWeight(parent.orientation) })
+            if let equalWeight = total.div(parent.children.count) {
+                for child in parent.children {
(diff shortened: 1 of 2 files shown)


MsfPablo:
Ping on this. The current `balance-sizes` resets total weight to a constant, which shifts the container's overall size; this preserves it. Happy to walk through the math if useful.

Issue #382, opened 2024-08-12
Control corner for hiding windows


clarkadamp:
> Since AeroSpace needs a free space to hide windows in, please make sure to arrange monitors in a way where every monitor has free space in the bottom right or left corner.

It would be good to have the ability to customize corners which are assigned to hiding.  My natural monitor configuration is to have a the center monitor high, then others placed below the main on each side, this allows my to use the mouse and move naturally between screens.

I have re-arranged my monitors so that hidden windows don't impact the other screens, now my habbits about moving the mouse means they run into walls rather than moving to the monitor.

Something like:
```
[corners]
main = ['top-right', 'top-left']
secondary = 'bottom-left'
'^built-in retina display$' = ['top-left', 'bottom-right'] 
```

nikitabobko:
> My natural monitor configuration is to have a the center monitor high, then others placed below the main on each side

Even if AeroSpace provided this configuration, there would be no corner for your middle monitor to hide windows in.

clarkadamp:
With this arrangement  would the top corners be available?
![Monitor Arrangement](https://github.com/user-attachments/assets/d86f96b8-84dd-4882-a7b9-9d7ff174c6ae)

nikitabobko:
It's not possible to move windows "above the monitor". It's only possible to move them below the monitor. It's a macOS restriction

> Is more than one corner needed, or can they all be stacked into one?

Every monitor should have at least one free bottom corner. Hiding windows in monitors they don't belong to leads to performance problems (#149) and macOS bugs (#289)

clarkadamp:
> It's only possible to move them below the monitor

Ahh, so it wasn't a design choice for below only.  

All good, thanks for your quick response, I'll work on my arrangements to see if I can come up with a compromise.

clarkadamp:
Happy to close this, thanks for the quick response

ELLIOTTCABLE:
I'd ask that this be re-opened; I just spent some time writing up #666, which is pretty close to a duplicate of this.

Further arguments I didn't include there:

- some of the performance concerns can be avoided in many users' scenarios; i.e. if the use selects a hiding-monitor that is the same DPI and/or size as their main monitor, or if the user doesn't move screens to/from the 'odd-one-out' display with a weird DPI very often
[...]

Issue #149, opened 2024-02-04
Windows will flash-resize with different resolution monitors on workspace switch


tale:
Refer to the video below, but whenever I switch workspaces, you can see that the window focuses and then expands to fit the bottom of the screen. I think I've figured out what the issue is in the way AeroSpace handles its virtual spaces. It's a combination of 2 different issues:
- When focusing a workspace, it appears that the window first gets maximized off-screen and then moved into focus.
- Windows by default will go to the bottom right when you focus away from their workspace.

Below, I've also shown my display configuration from settings, and I think the problem is because the display on the right is smaller than the primary display on the left, when I focus a workspace, it'll maximize the window to the size of the smaller display, move it to the bigger display, and then resize it again to fit the bigger display.

**macOS Version:** 14.3
**Keyboard Layout:** QWERTY
**Config:** https://github.com/tale/dotfiles/blob/main/config/bash/aerospace.toml
**Aerospace Version:**
```
aerospace CLI client version: 0.8.3-Beta bf02074c6aad434701809471fffdc7756e69d17e
AeroSpace.app server version: 0.8.3-Beta bf02074c6aad434701809471fffdc7756e69d17e
```

https://github.com/nikitabobko/AeroSpace/assets/46133336/0b894a2e-81f8-48f1-8922-f4f9261274e5

![CleanShot 2024-02-04 at 14 50 30](https://github.com/nikitabobko/AeroSpace/assets/46133336/4718fbdb-48ff-4dcd-a215-38ce94d9dbb6)

---

**Duplicated by:** https://github.com/nikitabobko/AeroSpace/issues/219 #235

nikitabobko:
Yeah, I recently realized that we have this problem too. Thanks for the report!

> When focusing a workspace, it appears that the window first gets maximized off-screen and then moved into focus.

Unfortunately, this is needed, otherwise, in some cases, macOS goes crazy and doesn't set sizes that I said it to set #143 

> Windows by default will go to the bottom right when you focus away from their workspace.

AeroSpace could be more intelligent here and try to find spots where it could hide windows on the current monitor. For example in your monitor configuration, it could hide windows this way:

![1](https://github.com/nikitabobko/AeroSpace/assets/20517828/875bcade-9f9f-4902-a474-d8b05fc5d305)

But not all monitor configurations will be possible to support

![2](https://github.com/nikitabobko/AeroSpace/assets/20517828/0da4383b-783e-434d-ab1d-c4620c3477d5)

[...]

Issue #1673, opened 2025-09-04
Design and implement the solution to notify users when their monitors are not properly arranged


nikitabobko:
AeroSpace requires the proper monitor arrangement: https://nikitabobko.github.io/AeroSpace/guide#proper-monitor-arrangement

Unfortunately, AeroSpace doesn't give users the feedback on when monitors are not properly arranged

The hardest part of this issue the UI and UX. I can think of one hard requirement: it must be possible to dismiss the notification completely (valid use case: https://github.com/nikitabobko/AeroSpace/discussions/1671) and make it clear where users can see these kinds of diagnostics if they want to.

## Possible UI/UX options

1. Standard macOS notifications. Unfortunately, notifications have problems and bad reputation among all major OSes. For macOS specifically: we need to get the permission to send notifications (I don't want to disturb users with notification permissions right when they start using AeroSpace), macOS notifications are not dismissable from a keyboard (#1672), macOS notification are allowed only for app bundles (no notifications for debug builds from CLI)
2. Floating NSPanel (like we do with macOS secure input warning). Unfortunately, this solution feels like a hack. It's like the option 1 but without the permission.
3. menubar icon/warning. Some users hide the menubar, and menubar items tend to get hidden because of the notch...

So far, the 2nd is my favourite option. Boy, it makes me sad that notifications suck so much that I am seriously thinking about floating NSPanel (which is basically a custom notification system hacked around)

mobile-ar:
I like the idea of using native local push notifications to show some alerts or custom notifications triggered by the user in specific cases. I think as you mentioned CLI builds might need to use a provisioning profile added to the project file in Xcode (not sure how to have that shared, maybe if you generate the debug ones only, given than a developer account is needed to generate those files and production ones will only be available in your GitHub actions using secure files, maybe?)
A problem with notifications is if the user disables this, we will never let them know that the issue is happening.

I think for this case when the monitors are not properly arranged, the `MessageView` UI could be used here because this is not a config error per se, but it's kind of a configuration issue/setup.
A new case to 'MessageType' could be added there for this other type of messages to reuse the keyboard shortcuts used there to close the message, based on the type hiding the reload/open config buttons (maybe could add a button to open display settings directly [if possible]).

dssste:
Option 3 is imo consistent to what we have so far in aerospace. If users hide their menu bar, they probably have their own menu bar solution, and in which case adding a callback similar to ```exec-on-workspace-change``` aligns to expectations.

[...]

Issue #1672, opened 2025-09-04
Research the possibility to dismiss notifications using macOS accessibility API


nikitabobko:
### Sanity check

- [x] I read https://github.com/nikitabobko/AeroSpace/issues/947

### Body

It's a shame that macOS doesn't provide any way to dismiss notifications from a keyboard. Maybe, it's possible to do via accessibility API, in this case, it'd be cool to provide aerospace command to dismiss notification(s)

acsandmann:
@nikitabobko there is indeed a way to do this. just sharing since i was looking into doing the same thing and came across this issue and the solution as well. https://github.com/briankendall/QuietYou

m13v:
we've worked extensively with the macOS accessibility tree for desktop automation and can confirm this is doable. notifications show up as AXNotificationCenter windows - you can find them by walking the AX tree from the system-wide accessibility element via AXUIElementCreateSystemWide().

the tricky part is that banner notifications and notification center items have different AX hierarchies, and they shift between macOS versions. for banners, you need to find the NotificationCenter process (by bundle ID), walk to its AXWindow children, then locate the close button or the notification group and call AXUIElementPerformAction with kAXPressAction. the QuietYou approach linked above works but relies on polling which can miss transient banners.

what we found more reliable is using AXObserver to watch for kAXCreatedNotification on the NotificationCenter app element, then immediately walking the new window's children to find and press the dismiss action. this catches banners as they appear rather than hoping a poll interval lines up.

m13v:
fwiw here is our macOS accessibility automation implementation that does exactly this kind of AX tree walking and element interaction: https://github.com/mediar-ai/mcp-server-macos-use/blob/main/Sources/MCPServer/main.swift

it handles finding app processes, walking the accessibility tree, and performing actions on elements via AXUIElementPerformAction. the same pattern would work for targeting NotificationCenter's AX elements.

PR #2212, opened 2026-08-06
Fall back to global edge when local hide corners are blocked

PannenetsF:
## Summary

- preserve the existing per-monitor hide-corner selection whenever either local bottom corner is free
- fall back to the nearest outer edge of the complete display layout when both local bottom corners overlap other displays
- add focused tests for single-monitor, one-blocked-corner, and both-blocked-corners layouts

## Root cause

macOS requires at least one pixel of a hidden window to remain on-screen. AeroSpace currently parks an invisible workspace at one of its monitor's bottom corners. When both corners are covered by adjacent displays, the selected point can expose most of the hidden window on the neighboring monitor.

The fallback is intentionally limited to that case. Existing valid monitor arrangements continue using their local monitor corner. The global target is calculated from current monitor geometry and `visibleRect`; no display coordinates are hard-coded.

## Validation

- `swift test` (385 tests passed)
- `swift test --filter HideCornerTest`
- tested on a four-display layout where a 1920x1080 center monitor has portrait displays blocking both bottom corners: the hidden window moved from `1919,1079` (visible on the adjacent display) to the computed outer edge `2999,2187`, then restored to `1,0` when its workspace became visible

Related to #149 and #382.

### DIFF
--- a/Sources/AppBundle/layout/refresh.swift
+++ b/Sources/AppBundle/layout/refresh.swift
@@ -148,8 +148,45 @@ func refreshObs(_: AXObserver, _: AXUIElement, notif: CFString, _: UnsafeMutable
     }
 }
 
-enum OptimalHideCorner {
+enum OptimalHideCorner: Equatable {
     case bottomLeftCorner, bottomRightCorner
+    case globalBottomLeftCorner, globalBottomRightCorner
+}
+
+func optimalHideCorner(for monitor: Rect, among monitors: [Rect]) -> OptimalHideCorner {
+    let xOff = monitor.width * 0.1
+    let yOff = monitor.height * 0.1
+    // brc = bottomRightCorner
+    let brc1 = monitor.bottomRightCorner + CGPoint(x: 2, y: -yOff)
+    let brc2 = monitor.bottomRightCorner + CGPoint(x: -xOff, y: 2)
+    let brc3 = monitor.bottomRightCorner + CGPoint(x: 2, y: 2)
+
+    // blc = bottomLeftCorner
+    let blc1 = monitor.bottomLeftCorner + CGPoint(x: -2, y: -yOff)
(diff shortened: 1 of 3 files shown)


nikitabobko:
I had very bad time debugging this issue https://github.com/nikitabobko/AeroSpace/issues/289

because of that I am very much against the global corners strategy. I prefer the current approach where we nudge users to rearrange their monitors for AeroSpace's optimal behavior.

JustinFirsching:
I understand not wanting to change the default behavior. However, from a user's perspective, seeing inactive workspaces leak onto another monitor feels more like a bug than a misconfiguration. The recommended workaround trades that problem for dead zones, clunky cursor movement, and a system-wide display configuration change.

In #1672 and #1673, you point out that forcing mouse interaction into a keyboard-driven workflow is poor UX. This workaround creates the inverse problem by making mouse navigation unreliable to accommodate the window manager. AeroSpace's `focus` commands help during keyboard navigation, but not when the user intentionally uses the mouse. Even in a keyboard-oriented workflow, moving the pointer between displays is a normal interaction.

Personally, this leads me to disable AeroSpace in favor of Rectangle when connected to multiple monitors. While that means giving up AeroSpace's workspace features, it avoids both the window leakage and the awkward cursor movement caused by misaligned displays.

Would you consider retaining the proposed fallback as an experimental opt-in setting? Perhaps:

```toml
window-hiding-strategy = 'experimental-fallback-to-global-edge'
```

or

```toml
enable-experimental-global-edge-support = true
```

This would leave the existing default behavior unchanged while making AeroSpace practical for users willing to accept the fallback's tradeoffs. An `experimental-` or similar prefix would make it clear that users are knowingly opting into a best-effort strategy that may not receive the same stability or support as the default behavior.

PR #2213, opened 2026-08-07
Fix bug: focus can land on a different window of the same app

dmellok:
## The bug

`focus` on one window of a multi-window app can focus a *different* window of that app. I hit it constantly with two Safari windows; it also pulled the focused workspace along when the window macOS picked was one AeroSpace had parked off-screen on an inactive workspace.

Conditions, which is why it doesn't reproduce for everyone:

- `Displays have separate Spaces` off, **or** a single monitor
- an app with more than one window
- keyboard-driven focus only — a mouse click focuses natively and never enters this path

## Cause

`MacApp.nativeFocus` has a fast path that skips the per-window AX raise and calls `nsApp.activate()` alone:

```swift
if (!NSScreen.screensHaveSeparateSpaces || monitors.count == 1) &&
    (lastNativeFocusedWindowId == windowId || windowsCount == 1)
```

`activate()` focuses whichever window macOS currently holds as the app's main one. `lastNativeFocusedWindowId` records what AeroSpace last *observed*, and the app's main window drifts from that on its own — a background window taking a link, an app reordering its own windows — so the condition isn't evidence that the requested window is the one `activate()` will bring.

The general path has a residual race too: the AX raise says which window should be main, and activating the app makes macOS restore whichever window *it* holds as main.

## The fix

Two parts:

1. The fast path now requires `windowsCount == 1`, where `activate()` can only mean the requested window. It still covers the case it was added for (apps slow to answer AX requests, which are overwhelmingly single-window).
2. The result is verified rather than assumed: 150ms after asking, check the app's focused window and ask once more if it isn't the one requested. Bounded to a single retry, and skipped when the app is no longer frontmost so that focus moving to another app is left alone.

## Measurements

Two Safari windows, focusing a different app between each attempt:

| | failures |
| --- | --- |
| before | 1 in 8 |
| after part 1 only | 1 in 8 |
| after both | 0 in 24 |

`swift test` passes (382 tests).

Happy to split this into two commits, drop the retry and keep only the fast-path fix, or move the discussion to Discussions first if you'd prefer that — just say which.

### DIFF
--- a/Sources/AppBundle/focusCache.swift
+++ b/Sources/AppBundle/focusCache.swift
@@ -11,5 +11,4 @@
         _ = nativeFocused?.focusWindow()
         lastKnownNativeFocusedWindowId = nativeFocused?.windowId
     }
-    nativeFocused?.macAppUnsafe.lastNativeFocusedWindowId = nativeFocused?.windowId
 }
--- a/Sources/AppBundle/tree/MacApp.swift
+++ b/Sources/AppBundle/tree/MacApp.swift
@@ -13,10 +13,10 @@ final class MacApp: AbstractApp {
     private let appAxSubscriptions: ThreadGuardedValue<[AxSubscription]> // keep subscriptions in memory
     private let windows: ThreadGuardedValue<[UInt32: AxWindow]> = .init([:])
     private var windowsCount = 0
-    var lastNativeFocusedWindowId: UInt32? = nil
     private var thread: Thread?
     private var setFrameJobs: [UInt32: RunLoopJob] = [:]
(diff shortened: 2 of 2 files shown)


dmellok:
Force-pushed a fix for the CI failure. Periphery caught two things:

- `raiseAxWindow` slipped in from my own fork and doesn't belong in this patch — removed.
- `lastNativeFocusedWindowId` became assign-only, since the shortcut this changes was its only reader. Removed the property and its assignment in `updateFocusCache` too, which I should have done in the first place.

All five jobs pass now.

PR #2217, opened 2026-08-08
Add --layout filter flag to list-windows

### LINKED ISSUE #429: Feature Request: filter windows by layout
Hello, 

Extracted from a comment on #272:

```
Would it be/is it possible to filter windows by their layout? Particularly, I am interested in being able to list floating windows... in that case, it would be trivial (as a user) to implement the requested feature
```
(i.e., a scratchpad).

By the way, many thanks for the app. It's a great tool for those used to i3 on Linux, and it's a comfort being able to keep a similar workflow when changing environment so drastically. 

Please let me know if there's anything else I should add. Thanks once again!

YuriNachos:
Closes #429.

## Summary

Adds a `--layout <layout>` filter flag to `list-windows`, so the command prints only the windows whose effective layout matches — e.g. `list-windows --layout floating` to list only floating windows (the scratchpad use case from #429).

## Problem

`list-windows` currently dumps every window across the target workspaces. #429 asks for a way to filter by the window's effective layout — floating windows in particular — which is useful for scripting and for inspecting how AeroSpace assigned windows to layouts.

## Changes

- `ListWindowsCmdArgs.swift` — new `--layout <target-layout>` argument. It accepts the same layout tokens the `layout` command uses (`floating`, `tiling`, `tiles`, `accordion`, `horizontal`, `vertical`, `h_tiles`, `v_tiles`, `h_accordion`, `v_accordion`); an invalid token fails with `Can't parse '--layout' argument. Possible values: …`.
- `format.swift` — new `windowMatchesLayout` predicate. It derives the window's effective layout through the same `getChildParentRelation` source of truth that backs the `%{window-layout}` interpolation variable, so `--layout floating` keeps exactly what `%{window-layout}` prints as `floating`. The family tokens are inclusive: `tiling` matches every tiling variant, `tiles`/`accordion` match the layout family, `horizontal`/`vertical` match the orientation, and `h_*`/`v_*` match exactly. Windows with no parent or with an illegal/native/popup parent relation match nothing.
- `ListWindowsCommand.swift` — applies the filter before the count/format stage, so `--count` counts only the matching windows too.
- `ListWindowsTest.swift` — new `testParseLayoutFilter` (valid tokens parse; an invalid token is rejected with the documented error) and `testRunFilterByLayout` (one tiling and one floating window in a workspace; each filter returns only the matching window id). Behavior without the flag is unchanged.

## Testing

- [x] `./test.sh` (build + test + lint + generate) — all tests passed.
- [x] The new tests assert the filter returns only matching windows and that omitting the flag preserves the previous output.

## Notes

Live end-to-end verification against real windows needs Accessibility permission and a running session; it is not exercised by the headless gate. The unit tests cover the filter logic; the build + lint + generate gate covers compilation.

### DIFF
--- a/Sources/AppBundle/command/format.swift
+++ b/Sources/AppBundle/command/format.swift
@@ -234,13 +234,17 @@ func unknownInterpolationVariable(variable: String, _ obj: AeroObj) -> String {
         "Possible values:\n\(getAvailableInterVars(for: obj.kind).joined(separator: "\n").prependLines("  "))"
 }
 
-private func toLayoutString(tc: TilingContainer) -> String {
+private func concreteLayoutDescription(tc: TilingContainer) -> LayoutCmdArgs.LayoutDescription {
     switch (tc.layout, tc.orientation) {
-        case (.tiles, .h): return LayoutCmdArgs.LayoutDescription.h_tiles.rawValue
-        case (.tiles, .v): return LayoutCmdArgs.LayoutDescription.v_tiles.rawValue
-        case (.accordion, .h): return LayoutCmdArgs.LayoutDescription.h_accordion.rawValue
-        case (.accordion, .v): return LayoutCmdArgs.LayoutDescription.v_accordion.rawValue
+        case (.tiles, .h): return .h_tiles
+        case (.tiles, .v): return .v_tiles
+        case (.accordion, .h): return .h_accordion
+        case (.accordion, .v): return .v_accordion
     }
+}
+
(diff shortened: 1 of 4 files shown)


YuriNachos:
This adds the --layout filter to list-windows from #429, reusing the existing effective-layout derivation so the flag filters exactly what the command prints. Given the project's habit of shaping features in Discussions first, is the interface as implemented (regex argument named --layout) acceptable, or would you prefer to weigh the design in a Discussion before it merges? I'm happy to open one or adjust the flag shape either way.

PR #2225, opened 2026-08-11
Insert replaced native-tab windows into their old tree slot

### LINKED ISSUE #68: Native macOS Big Sur tabs are considered windows

jeajar:
Fixes #68. Same approach as #2206, scoped to a single stale window per
focus-change instead of reconciling the whole tracked-window set.

## PR checklist

- [x] Explain your changes in the relevant commit messages rather than in the PR description.
- [x] Each commit must explain what/why/how and motivation in its description.
- [x] Each commit must be an atomic change.
- [x] `./test.sh` exits with zero exit code.

### DIFF
--- a/Sources/AppBundle/focusCache.swift
+++ b/Sources/AppBundle/focusCache.swift
@@ -11,5 +11,10 @@
         _ = nativeFocused?.focusWindow()
         lastKnownNativeFocusedWindowId = nativeFocused?.windowId
     }
-    nativeFocused?.macAppUnsafe.lastNativeFocusedWindowId = nativeFocused?.windowId
+    // Dialogs/sheets (e.g. Finder's "Connect to Server") aren't native-tab group members. Letting
+    // one become the replacement candidate poisons the comparison: it closes normally later, so
+    // whichever tiled window is actually mid-tab-swap when it closes never gets matched.
+    if !(nativeFocused?.parent is FloatingWindowsContainer) {
+        nativeFocused?.macAppUnsafe.lastNativeFocusedWindowId = nativeFocused?.windowId
+    }
 }
--- a/Sources/AppBundle/layout/refresh.swift
+++ b/Sources/AppBundle/layout/refresh.swift
@@ -125,6 +125,7 @@ private func refresh() async throws {
     let mapping = try await MacApp.refreshAllAndGetAliveWindowIds(frontmostAppBundleId: NSWorkspace.shared.frontmostApplication?.bundleIdentifier)
     // A mouse-driven native-tab switch can be observed first by this periodic refresh rather
     // than getFocusedWindow. Preserve the old tree slot before garbage collection removes it.
+    var splicedWindowIds: Set<UInt32> = []
     for (app, refreshResult) in mapping {
         if let replacement = refreshResult.nativeTabReplacement,
            refreshResult.aliveWindowIds.contains(replacement.focusedWindowId)
@@ -134,6 +135,7 @@ private func refresh() async throws {
                 macApp: app,
                 replacingNativeTabWindowId: replacement.staleWindowId,
             )
+            splicedWindowIds.insert(replacement.focusedWindowId)
         }
     }
     let aliveWindowIds = mapping.values.flatMap(\.aliveWindowIds).toSet()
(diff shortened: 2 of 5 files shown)


Miyou:
I tested #2225 on macOS 15.7.4 with AeroSpace 0.21.3. It fixes native-tab creation, but mouse-clicking back to an inactive Finder tab can still remove the logical window from its workspace and later resurrect the retired AX ID elsewhere.

I have a small follow-up that retains replaced native-tab IDs and applies periodic-refresh replacements before garbage collection. It passes `./test.sh`, seven focused regression tests, repeated Finder create/switch/close tests, a separate-window guard, and an Alacritty test.

Would you prefer a PR against this branch?

jeajar:
@Miyou  yes, please open a PR against this branch, that would be great.

Miyou:
Opened https://github.com/jeajar/AeroSpace/pull/1 against your `generic-native-tab-window-fix` branch. It contains only the periodic-refresh/retired-ID hardening and its regression tests; the focused seven-test suite and the full `./test.sh` gate pass.

jeajar:
Merged your PR — while reviewing it I also found and fixed a related quirk where Finder's "Connect to Server" dialog was getting mistaken for a native-tab replacement candidate, causing a permanent gap in the layout once the dialog closed; same underlying bug (#68)

mayaanhafeez:
I tested the latest three-commit version of this PR locally (`07f6b44f`): `./test.sh` passes. While reviewing the lifecycle, I found two cases that seem worth addressing before merge.

1. **Focus changes between real windows can satisfy the replacement predicate.** `nativeTabReplacementCandidate` only requires the previous focused ID to be tracked and absent from the current `AXWindows`. That also occurs when the previous real window is on another Space, minimized/fullscreen, hidden, transiently omitted by AX, or belongs to another independent same-app window/tab group. In those cases the old real window can be retired and its tree slot transferred to the newly focused real window. A useful regression test would be: same app, real windows A and B, focus moves A -> B while `AXWindows == [B]`; A must remain registered/bound. Similar tests would cover different Spaces, minimized/fullscreen A, and two separate native-tab groups. The candidate appears to need affirmative same-group/handover evidence or at least state/container/transition guards beyond absence from `AXWindows`.

2. **Cancellation can leave replacement half-committed.** Both focus and periodic-refresh paths mutate `windows`/`retiredNativeTabWindowIds` before the main-actor tree splice completes. `MacWindow.getOrRegister` can then throw while reading the rect or recording debug data, and `registerNativeTabReplacement` mutates the tree before an awaited/throwing debug call. A cancelled refresh may therefore retain the retired ID while never completing the replacement, so a later refresh cannot emit the same proposal and GC removes the old slot. I think AX refresh should return an immutable proposal, perform cancellable reads first, then commit AX retirement and tree replacement together in a non-suspending/non-throwing main-actor transaction. Cancellation-injection tests around candidate detection, rect lookup, and pre-splice processing would catch this.

The public API experiments I ran support the underlying handover model: selected native tabs are in `AXWindows` and expose an `AXTabGroup`, while hidden tabs remain valid CG/AX window IDs but disappear from `AXWindows`. The difficulty is distinguishing that handover from the other legitimate reasons a real window is absent. I did not open another PR because this branch is already substantially ahead of the CG/offscreen prototype I tested.

jeajar:
Thanks for the deep dive here.

**Case 2: fixed**, in `9cac6230`. You had it right — the AX thread retires the old id before the tree splice runs, and a cancelled refresh (tab switches trigger a burst of those) could land between the two, stranding the old id with no way to retry. Made the rect fetch non-cancellable once a splice is coming, so once we commit to retiring an id, the splice always finishes in the same cycle. Fixes both the focus-change and refresh paths.

**Case 1: not fixing here.** You're right that it's real — the `AXWindows` doc comment confirms off-Space windows drop out of it too, so the check can't tell a genuine tab handover from an unrelated window going missing. But the real fix needs a new signal (`AXTabGroup`) that we don't read anywhere yet, not a small tweak — that's its own PR, not a follow-up to this one.

Worth noting: as far as I understand, this isn't a regression we're shipping and ignoring — nothing on `main` acted on AX-window-absence before this branch, so it's a new limitation, not a pre-existing bug.

Probably worth opening a separate issue for case 1 linking your `AXTabGroup` findings. Happy to have you take a crack at it if you want.

PR #2242, opened 2026-08-25
Make non-empty-workspaces-root-containers-layout-on-startup a functional tiles|accordion|smart selector

christophervalles:
### Problem

`non-empty-workspaces-root-containers-layout-on-startup` currently exists only as a deprecated no-op — the parser accepts `'smart'` (and warns to drop the key otherwise), the value is stored as `Void` and discarded, and `smartLayoutAtStartup()` always applies the hardcoded heuristic:

```swift
switch root.children.count <= 3 {
    case true: root.layout = .tiles
    case false: root.layout = .accordion
}
```

That heuristic silently overrides `default-root-container-layout = 'tiles'` at startup for any workspace with more than 3 windows — it comes up as accordion instead. For users coming from i3/sway who expect workspaces to always tile unless they explicitly stack, this is surprising and there's currently no way to turn it off.

### Change

Restore the key as a real selector (default `smart` → **no behavior change** for existing configs):

| value | behavior |
|-------|----------|
| `smart` | tiles if the root has ≤ 3 windows, else accordion (the current heuristic; default) |
| `tiles` | always tile at startup (i3/sway-like) |
| `accordion` | always accordion at startup |

- `Config`: `Void` placeholder → typed `NonEmptyWorkspacesRootContainersLayoutOnStartup` enum, default `.smart`
- `parseConfig`: parser returns the enum (`tiles|accordion|smart`) instead of the deprecation no-op
- `smartLayoutAtStartup()`: honors the config value
- Updated `default-config.toml`; replaced the now-obsolete deprecation test with parse tests for all three values + default + invalid

`swift build`, `swift test --filter ConfigTest` (40 pass), and `./lint.sh` are green.

### DIFF
--- a/Sources/AppBundle/config/Config.swift
+++ b/Sources/AppBundle/config/Config.swift
@@ -38,7 +38,7 @@ struct Config: ConvenienceMutable {
     var afterStartupCommand: Shell<any Command> = .empty
     var _indentForNestedContainersWithTheSameOrientation: Void = ()
     var enableNormalizationFlattenContainers: Bool = true
-    var _nonEmptyWorkspacesRootContainersLayoutOnStartup: Void = ()
+    var nonEmptyWorkspacesRootContainersLayoutOnStartup: NonEmptyWorkspacesRootContainersLayoutOnStartup = .smart
     var defaultRootContainerLayout: Layout = .tiles
     var defaultRootContainerOrientation: DefaultContainerOrientation = .auto
     var startAtLogin: Bool = false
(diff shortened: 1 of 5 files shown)


nikitabobko:
> For users coming from i3/sway who expect workspaces to always tile unless they explicitly stack, this is surprising and [...]

I agree, but I don't like your proposed solution. This config option was deprecated because it's config creep

> and there's currently no way to turn it off.

Wrong. You can do:

```toml
after-startup-command = 'layout tiles'
```

---

Please search for GitHub Discussions. I agree that the problem is legit, but I don't know what is the best way to solve it

PR #2268, opened 2026-09-16
Replace Task.sleep(for:) with Task.sleep(nanoseconds:)

jamesbroadhead:
Task.sleep(for:) requires macOS 13+. The nanoseconds variant is a drop-in replacement.


nikitabobko:
Answered here: https://github.com/nikitabobko/AeroSpace/discussions/2269#discussioncomment-18519970

PR #2274, opened 2026-09-17
focus-monitor: always re-assert macOS focus (#576)

### LINKED ISSUE #576: Making sure macOS's "monitor focus" agrees with `focus-monitor` in the case of empty monitors
`focus-monitor` is great (Aerospace has really come a long way!); but at the moment, it's got one glaring flow issue.

When all monitors have at least one window, cycling through them with a `focus-monitor left` command or similar _also_ drags the macOS "focus" along (that is, the top menubar is faded out on the other, non-focused displays; and <kbd>⌘</kbd><kbd>␣</kbd> pops up a search window on the correct monitor.)

However, if there's a monitor with _no_ windows on it, then when you `focus-monitor` onto that empty monitor, there's no integration with the system. (The wrong monitor's menubar is highlighted, <kbd>⌘</kbd><kbd>␣</kbd> will open windows on the wrong monitor, etc.)

This is slightly worse when [`borders`](https://github.com/FelixKratz/JankyBorders) or similar is installed - now, the 'active window' indicator is no longer doubling as a "what display/container AeroSpace is going to operate on" indicator.

One janky fix, is to rewrite the default `on-focused-monitor-changed = ['move-mouse monitor-lazy-center']` hook to also …
1. determine if there's any windows on the newly-focused monitor and
2. then, if not, execute a mouse-click to click on the desktop and activate that monitor from macOS's point of view.

Thus, I suppose, at least one potential "fix" for this might be to make that behaviour the default (hopefully a lot) faster than indirecting through a shell, especially when tabbing through monitors quickly.

## Checklist

- [x] I've searched for duplicates. My search queries were (please don't forget to search in closed issues too):
  - `focus-monitor empty`
  - `"focus-monitor"`
- [x] I've read the guide https://nikitabobko.github.io/AeroSpace/guide
- [x] I've searched in the documentation (https://nikitabobko.github.io/AeroSpace/guide & https://nikitabobko.github.io/AeroSpace/commands)

```
❯ aerospace --version         
aerospace CLI client version: 0.14.2-Beta 0cb8dbdfc5ee73b8cbc200f175f467ebead55201
AeroSpace.app server version: 0.14.2-Beta 0cb8dbdfc5ee73b8cbc200f175f467ebead55201
```

hunner:
Partially addresses #576.

`focus-monitor` can be a silent no-op. `setFocus` returns early when the model already equals the request. `runLightSession` only pushes focus out to macOS when it observes a change.

Once AeroSpace's model and macOS have drifted apart, `focus-monitor` therefore does nothing at all. The `on-focused-monitor-changed` callback does not fire either. Meanwhile macOS focus sits on a different monitor, so the shortcut for the monitor you want appears dead. You have to focus some other monitor first to make the model disagree, then come back.

Drift is easy to reach. Focusing a workspace with no windows leaves no window for `syncFocusToMacOs` to focus, so macOS stays where it was. `updateFocusCache` only corrects the model when the natively focused window changes, and it has not changed.

This re-asserts macOS focus unconditionally. An explicit request to focus a monitor should take effect regardless of what the model believes is already true.

It does not fix #576 itself. With an empty workspace there is still no window to focus. What it does is make the desync recoverable in one keypress instead of two.

I also tried activating AeroSpace itself for the empty case, and withdrew it. Making AeroSpace frontmost caused the next focus request to be ignored by macOS often enough to be worse than the problem it solved.

**Testing:** run daily on a two-monitor setup (Dell + built-in, ten workspaces).

## PR checklist

`./test.sh` was not run. XCTest ships with full Xcode and this machine has Command Line Tools only. That is why this is a draft. Happy to install Xcode and run it if you would rather see it green first.


hunner:
Closing this. It measurably improves the situation but does not fully fix it, and a partial fix here is not worth your review time.

Details of what I found are in https://github.com/nikitabobko/AeroSpace/issues/101 instead. Short version: I could not find an ordering of raise, kAXMain, kAXFocused and app activation that reliably wins, because macOS restores the app's own focused window after every hook I can observe, including NSWorkspace.didActivateApplicationNotification. Measured 7 of 9 remaining failures with an assert firing after that notification.

Happy to revisit if you know of a lever I missed.

PR #2275, opened 2026-09-17
Focus the requested window on multi-monitor setups (#101)

hunner:
Addresses #101. A window of the requested app on a different monitor gets focused instead of the requested window. There are two independent causes, one commit each.

**The fast path applies when it should not.** `nativeFocus()` has a shortcut that activates the app without raising the requested window. It is guarded by `!NSScreen.screensHaveSeparateSpaces || monitorInfos.count == 1`, which assumes the macOS bug needs "Displays have separate Spaces" enabled. It does not. With that setting off and two monitors the shortcut still applies. The app then restores whatever window it considers key, and focus lands on the wrong monitor. The guide recommends turning that setting off, so this path is reachable by exactly the users who follow the documentation.

**`kAXMain` is not focus.** The slow path set `kAXMain` and raised the window, then activated the app. Activating an app makes macOS restore that app's own focused window. `kAXMain` marks the main window, which is a different thing, and raising affects z-order rather than focus. Setting `kAXFocused` on both sides of the activation is what actually moves focus. `Ax.isFocused` already existed but was read-only, so this adds a writable twin.

Two related problems are left untouched. Both look like design questions rather than bugs with obvious fixes. `updateFocusCache` is edge-triggered on the focused window id changing, so a disagreement between the model and macOS that does not change that id never self-corrects. `MacApp.focusJob` is a single global slot that each `nativeFocus` cancels, so switching rapidly through an app that is slow to answer AX drops focus requests entirely.

**Testing:** run daily on a two-monitor setup (Dell + built-in) with Firefox windows on both monitors, which is the case that reproduces it. I instrumented `runLightSession` to record what macOS had focused against what was requested. That gave 29 of 29 requests honoured with both commits applied, against a steady trickle of wrong-window focuses before.

## PR checklist

`./test.sh` was not run. XCTest ships with full Xcode and this machine has Command Line Tools only. That is why this is a draft. Happy to install Xcode and run it if you would rather see it green first.


hunner:
Closing this. It measurably improves the situation but does not fully fix it, and a partial fix here is not worth your review time.

Details of what I found are in https://github.com/nikitabobko/AeroSpace/issues/101 instead. Short version: I could not find an ordering of raise, kAXMain, kAXFocused and app activation that reliably wins, because macOS restores the app's own focused window after every hook I can observe, including NSWorkspace.didActivateApplicationNotification. Measured 7 of 9 remaining failures with an assert firing after that notification.

Happy to revisit if you know of a lever I missed.

